PR #7, opened 2025-05-28
DEMOS-166 - SideNav Setup

nunesmlucas:
####  Summary

This PR implements a full visual and structural revamp of the `SideNav` component, fully aligned with the Figma design system. Enhancements include spacing, typography, icon sizing, collapsed behavior, and route wiring across new placeholder pages for demonstration.

---

#### Updated Routing in `DemosRouter.tsx`

- Routes are now nested under a shared `AppLayout` wrapped with `PrivateRoute`
- Includes newly added pages:  
  - `/demonstrations`  
  - `/actions`  
  - `/tasks`  
  - `/dashboard`  
  - `/reports`  
  - `/budget`
- Dev-only route `/components` preserved conditionally via `NODE_ENV`

---

#### 📁 File Changes

- `SideNav.tsx` — Major layout overhaul and toggle logic improvements
- `DemosRouter.tsx` — Routing updated with shared layout and new pages
- `components/icons/` — Includes new icons (`ActionsIcon`, `BudgetIcon`, `DashboardsIcon`, etc.)
- `pages/` — Added placeholder display files:
  - `Demonstrations.tsx`
  - `Actions.tsx`
  - `Tasks.tsx`
  - `Reports.tsx`
  - `Budget.tsx`


client/src/components/icons/DashboardsIcon.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from "react";
+
+export const DashboardsIcon = (props: React.SVGProps<SVGSVGElement>) => (
```
dustbuster, line 3: Should this be singural?
They rest of the components are singural. 
It is one icon, right? Not a set of icons. 

Screenshot 2025-05-29 at 09 43 08

client/src/components/layout/NavLinkItem.tsx
```diff
@@ -0,0 +1,27 @@
+// Reusable individual nav item
+import React from "react";
+
+import {
+  Link,
+  useLocation
+} from "react-router-dom";
+
+interface NavLinkItemProps {
+  link: { label: string; href: string; icon?: React.ReactNode };
```
tresdonicf, line 10: I think we might want to pass these props as separate props rather than nest, we can keep it defined as it is in `SideNav` but at this level it might be best to have them separated out. Otherwise maybe we create a type / interface for `LinkProps` so we can pass in `link: LinkParameters`type stuff

client/src/public/css/index.css
```diff
@@ -57,3 +57,14 @@
...
+.font-semibold {
+  font-weight: var(--font-semibold) !important;
+}
+.text-nav {
+  font-weight: 400;
+  font-size: 16px;
+  line-height: 100%;
+  letter-spacing: 0;
+  text-align: center;
+}
```
tresdonicf, line 70: Since we're using tailwind we should translate this stuff to tailwind instead of adding classes

client/src/components/icons/DashboardsIcon.tsx
nunesmlucas: Nice catch

client/src/public/css/index.css
```diff
@@ -18,7 +18,7 @@
   --font-h4: 2rem / auto;
   --font-h5: 1.6rem / auto;
   --font-regular: 1.6rem/1.6rem;
-  --font-semibold: 1.6rem/2.1rem;
+  --font-semibold: 600;
```
tresdonicf, line 21: I'm not certain of the spec here with the `/` but I think that if we want to use these weights in tailwind we'll need to use `--font-weight-*` as it is according to the [Theme variable namespace](https://tailwindcss.com/docs/theme#theme-variable-namespaces).

client/src/router/DemosRouter.tsx
```diff
@@ -1,26 +1,48 @@
...
-      <Route path="/components" element={<ComponentLibrary />} />
-      <Route path="/hooks" element={<TestHooks />} />
-    </>
-  ) : null;
+  // Authenticated routes with shared layout
+  const authenticatedRoutes = (
+    <Route
+      element={
+        <AppLayout>
+          <Outlet />
```
tresdonicf, line 31: What is `<Outlet />` here?

tresdonicf (commented):
Nice work on this! Please add some unit tests 🙏  Copilot can generate some pretty decent testcases for some components!

client/src/router/DemosRouter.tsx
nunesmlucas: `<Outlet />` is just how `react-router-dom` knows where to stick the page content inside a parent route.

In our setup, it basically tells React: “Hey, render the sidebar and nav from AppLayout, and then drop whatever page (like `Demonstrations` or `Actions`) the route matches right here.”

Without it, we’d get the layout but none of the actual page content would show up.

client/src/router/DemosRouter.tsx
```diff
@@ -1,26 +1,48 @@
 import React from "react";
+
+import AuthComponent from "components/auth/AuthComponent";
+import AppLayout from "components/layout/AppLayout";
+import Actions from "pages/Actions";
+import Budget from "pages/Budget";
```
tresdonicf, line 6: Just noticing these should perhaps be called `ActionsPage`, `BudgetPage`, etc - but personally I would leave these out for now until we define what we want to show under them to reduce noise for those working on it next. If you do want some pages for now to test the switching perhaps we can use the `/components` and `/hooks` pages

client/src/router/DemosRouter.tsx
tresdonicf: I see, could we also do something like:
```
<Route path="/demonstrations" element={<PrimaryLayout><DemonstrationsPage /></PrimaryLayout>} />
```

I think this may be a bit more explicit and perhaps easier to understand

dustbuster (approved):
This looks great! I would suggest screenshot for all client PRs

PR #8, opened 2025-05-28
DEMOS-63: Adds kafka server and demo of sending message

james-becwar-cms:
Adds a kafka server to the devcontainer setup and some example code on sending a message to kafka from node.


tresdonicf (approved):
This is cool to see, very simple code for these operations! How would I go about running this `index.js` file?

PR #11, opened 2025-05-30
DEMOS-252: added prisma and graphql for new data model begin

connor-parke-icf:
Base Server code. A starting point. Handles the Prisma and Graphql endpoints for Demonstration, DemonstrationStatus, Permission, Role, State and User. Also creates the prisma definitions for the join tables rolePermission, userRole, userState, and userStateDemonstration. Basic CRUD operations were implemented in the resolvers for each of the main data types. Each main type has a corresponding Typescript model definition for itself, and its Add and Update function inputs. 

Also includes a Seeder function, for quickly populating the database with dummy data. 

Integrates the existing Auth structure, also providing a bypass .env flag.


server/src/model/user/userResolvers.ts
```diff
@@ -0,0 +1,128 @@
...
+          }),
+          ...(roleIds && {
+            userRoles: {
+              create: roleIds.map((roleId: string) => ({ roleId })),
+            },
+          }),
+          ...(demonstrationIds && {
+            userStateDemonstrations: {
+              create: (
+                await prisma.demonstration.findMany({
```
connor-parke-icf, line 73: need to grab the state of the demonstration to insert into the userStateDemonstrations table.

server/src/prismaClient.ts
```diff
@@ -0,0 +1,4 @@
+import { PrismaClient } from "@prisma/client";
+import prismaRandom from "prisma-extension-random";
+
+export const prisma = new PrismaClient().$extends(prismaRandom());
```
connor-parke-icf, line 4: ran into an interesting user when trying out a super big query. But, theres a limit to how many concurrent connections you can have, and it would fail when there were too many. So relegated all the connections to one prisma client. I also included a prismaRandom extension used in the seeder.

server/src/seeder.ts
connor-parke-icf, line 1: clears the tables, then loops through all the tables and adds random data to it. Then, chooses pairs of tables and joins them through the join tables. If a constraint is breached, it just ignores it and moves on.

server/.env.example
```diff
@@ -0,0 +1,2 @@
+DATABASE_URL=""
```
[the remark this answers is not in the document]
james-becwar-cms: I think this is fine or localhost version

server/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "depcheck": "^1.4.7",
+    "graphql-scalars": "^1.24.2",
+    "graphql-tools": "^9.0.18",
+    "jsonwebtoken": "^9.0.2",
+    "jwks-rsa": "^3.2.0",
+    "prettier": "3.5.3",
+    "prisma-extension-random": "^0.2.2",
+    "react": "^19.1.0"
+  },
+  "scripts": {
```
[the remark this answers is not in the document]
james-becwar-cms: there are some npm utilities that can find unused packages too.  Maybe create a story for it?

server/src/auth/auth.config.ts
```diff
@@ -0,0 +1,15 @@
+interface AuthConfig {
+  jwksUri: string;
+  issuer: string;
+  audience: string;
+};
+
+export const LOCAL_AUTH_CONFIG: AuthConfig = {
+  jwksUri: 'https://cognito-idp.us-east-1.amazonaws.com/us-east-1_A7CaR2Wo3/.well-known/jwks.json',
```
james-becwar-cms, line 8: We should talk to devops and security about these not sure how we want to handle them

server/src/auth/auth.util.ts
```diff
@@ -0,0 +1,151 @@
...
+
+/**
+ * Verifies the Cognito JWT and extracts user info.
+ */
+export const getCognitoUserInfo = (
+  req: IncomingMessage,
+): Promise<DecodedJWT> => {
+
+  // Bypass authentication for testing purposes
+  if (process.env.BYPASS_AUTH === "true") {
```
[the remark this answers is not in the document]
james-becwar-cms: we should add this to the .env.example

server/src/auth/auth.util.ts
```diff
@@ -0,0 +1,151 @@
...
+};
+
+/**
+ * Fetches a user's role from the DB using their Cognito sub.
+ */
+export const getUserRoles = async (
+  cognitoSubject: string,
+): Promise<string[] | null> => {
+
+  if (process.env.BYPASS_AUTH === "true") {
```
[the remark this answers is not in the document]
james-becwar-cms: We should add this to the .env.example

server/src/model/demonstration/demonstrationSchema.ts
```diff
@@ -0,0 +1,84 @@
...
+    deleteDemonstration(id: ID!): Demonstration
+  }
+
+  type Query {
+    demonstrations: [Demonstration]!
+    demonstration(id: ID!): Demonstration
+  }
+`;
+
+export type DateTime = Date;
```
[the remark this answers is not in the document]
james-becwar-cms: Lets create a follow up story for this, because PG also has a datetimez (with timezone) that is really nice to use.

server/src/model/graphql.ts
```diff
@@ -0,0 +1,48 @@
...
+import { permissionResolvers } from "./permission/permissionResolvers";
+
+import { roleSchema } from "./role/roleSchema";
+import { roleResolvers } from "./role/roleResolvers";
+
+import { stateSchema } from "./state/stateSchema";
+import { stateResolvers } from "./state/stateResolvers";
+
+import { userSchema } from "./user/userSchema";
+import { userResolvers } from "./user/userResolvers";
```
james-becwar-cms, line 17: Lets create a story to use glob to autoload these.

server/src/model/state/state.test.ts
james-becwar-cms, line 1: Do we need this empty file?

server/src/prismaClient.ts
james-becwar-cms: We might need to look into connection pooling etc... we can create a story when we have more details and examples of it failing.

server/src/seeder.ts
james-becwar-cms: This is awesome, but lets create a follow up story to make sure this isn't run in prod.

james-becwar-cms (approved):

server/src/auth/auth.util.ts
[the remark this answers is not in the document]
tresdonicf: Alternatively, we use `process.env.NODE_ENV` which is canonical for tooling to implement as "production", "development", or "test" (+ probably some more which I'm not familiar with). This way we just compare it to `process.env.NODE_ENV == "development"` and change app behavior based on this rather than requiring a bespoke environment variable. Not important for this PR, just want to bring this up as we can probably get it for free

server/src/model/_rolePermission/rolePermission.prisma
tresdonicf, line 1: What does the underscore mean in front of some of these models?

server/src/model/demonstration/demonstrationSchema.ts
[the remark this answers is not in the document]
tresdonicf: Adding some context for this - currently the linting rule I made expects non primitive GQL types to just match the type name of the referenced field (aka: `DemonstrationStatus` should be called the same in both declarations) and `Date` isn't a [primitive concept in GQL (or a Scalar Type)](https://spec.graphql.org/draft/#sec-Scalars) so I had to do this type alias to make the linting rule happy

This is to say we can also adjust the behavior of `server/eslint-rules/validate-graphql-typescript-match.js` to handle `Dates` differently if we want.

tresdonicf (approved):
Thanks for moving this over, it looks great!

server/src/model/_rolePermission/rolePermission.prisma
connor-parke-icf: i used that to indicatge that these were Join tables. Not really important, i just wanted them to appear separate from the rest of the models. The only have prisma definitions, and dont have graphql endpoints.

server/src/auth/auth.util.ts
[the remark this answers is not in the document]
connor-parke-icf: i think i'd rather have this as a separate variable so that it can be enabled/disabled without affecting anything else that we might tie to the environment flag, like minification or banners or whatever (when the time comes for those things)

PR #12, opened 2025-05-30
DEMOS-160: DEMOS Login Page

tresdonicf:
DEMOS-160: This PR bolsters the work related to the [cognito login page](https://us-east-1a7car2wo3.auth.us-east-1.amazoncognito.com/login?client_id=5km9thunj8g6qd32s5et2i8pga&code_challenge=a2yIcl0HznGFn-xc62uBVJBKBItrEm4z046dqmlUQ1o&code_challenge_method=S256&redirect_uri=http://localhost:3000&response_type=code&scope=openid+email+phone&state=e0f4cc868c224bb8ac4b251e9e0b2027) by removing some routes that aren't needed, adding tests, and documenting next steps for when we have a better understanding of IDM integration.

As of now we're a bit limited in our ability to add anything else to this but I believe all the pieces are in place now to make the necessary changes when we know a bit more about the requirements

https://github.com/user-attachments/assets/aaa9b771-5e18-4230-85d5-e3ee5a137a11

(Slighly out of date, added our placeholder buttons back in side nav)


client/eslint.config.js
```diff
@@ -23,8 +28,8 @@ export default tseslint.config(
       "no-trailing-spaces": "error", // Disallow trailing spaces
       "no-tabs": "error", // Disallow tabs for indentation
       "eol-last": [ "error", "always" ], // Newline at the end of files
-      // Trailing commas in multiline arrays and objects
-      "comma-dangle": [ "error", { "arrays": "always-multiline", "objects": "always-multiline" } ],
+      "comma-dangle": [ "error", { "arrays": "always-multiline", "objects": "always-multiline" } ], // Trailing commas in multiline arrays and objects
+      "no-restricted-exports": ["error", { "restrictDefaultExports": { "direct": true } }], // Disallow default exports
```
tresdonicf, line 32: `export default` for us is keeping us in a weird place of sometimes you have to use named exports and sometimes you have to use default exports and I think we should make this more consistent towards named exports for now.

The difference is in being able to name it anything when importing or having to import by name and importing by name is a bit more rigorous:
```
import lkadjsfkljadsl from "./LandingPage" // Default export, any name we give it on importing is acceptable
import { LandingPage } from "./LandingPage" // Have to use the name the author of LandingPage set out for it
```

tresdonicf: Ironic that this is after the `comma-dangle` rule which is supposed to reduce noise in these sorts of changes because the commas should already be there for previous lines

client/src/layout/SideNav.tsx
```diff
@@ -62,48 +119,12 @@ export const SideNav: React.FC<SideNavProps> = ({ collapsed, setCollapsed }) =>
...
-                  }>
-                    {React.cloneElement(link.icon, {
-                      className: "w-[14px] h-[14px]",
-                    })}
-                  </span>
+      <NavLinks collapsed={collapsed} navLinks={navLinks} />
+      <DebugOnly>
+        <hr className="my-2 border-t border-gray-200" />
+        <NavLinks collapsed={collapsed} navLinks={debugNavLinks} />
+      </DebugOnly>
```
tresdonicf, line 126: These changes are a bit noisy in github - essentially this just wraps the code that was here in a component so we can call it with these different set of nav links (one for dev only)

client/src/pages/LandingPage.tsx
```diff
@@ -1,19 +1,12 @@
 import React, { Fragment } from "react";
 import { Main } from "components";
-import { PrimaryButton } from "components/button/PrimaryButton";
-import { useNavigate } from "react-router-dom";
 import { Demonstrations } from "pages/Demonstrations";
 
+/** TODO: Look into if the Demonstration Page should just be the landing page */
```
tresdonicf, line 5: I've added this because the more we work on this the more I think we can have either the `DemonstrationsPage` or the `LandingPage` but right now they just render the same thing. It seems almost more like the user should be navigated to `/demonstrations` upon going to `/` but didn't want to scope creep too much

client/src/router/DemosRouter.tsx
```diff
@@ -1,59 +1,43 @@
...
-import { getCognitoConfig } from "../router/cognitoConfig";
 import { userMocks } from "hooks/userMocks";
+import { Demonstrations } from "pages/Demonstrations";
 
-export function DemosRouter() {
+export const DemosRouter = () => {
+  // TODO: When we know what IDM integration looks like
+  // We will want to read the JWT claims and
+  // add it to the AuthProvider (specifically the user object)
   const cognitoConfig = getCognitoConfig();
```
tresdonicf, line 17: Documented where we'll want the IDM integration stuff to live. For now we don't really need a concept of "protected routes" - since the app either will or will not be accessible based on IDM authentication. Pages like the `/admin` page might not be and we can revisit this when we have this integration figured out more and non-local environments to test it on.

client/src/layout/SideNav.tsx
```diff
@@ -3,32 +3,89 @@ import React from "react";
...
-const navLinks = [
+interface NavLink {
+  label: string;
+  href: string;
+  icon: React.ReactElement;
+}
+
+const navLinks: NavLink[] = [
   { label: "Demonstrations", href: "/demonstrations", icon: <DemonstrationIcon /> },
-  { label: "Actions", href: "/?actions", icon: <ActionIcon /> },
```
connor-parke-icf, line 17: I think its still worth keeping these links around even if they dont point to anything yet. I think theres value in letting there be a current state of things to comment on/critique, rather than having empty space.  Kinda just shows how its projected to look in the future so we have a baseline to adjust

client/src/layout/SideNav.tsx
```diff
@@ -3,32 +3,89 @@ import React from "react";
...
-  { label: "Tasks", href: "/?tasks", icon: <ListIcon /> },
-  { label: "Dashboards", href: "/components", icon: <DashboardIcon /> },
-  { label: "Reports", href: "/?reports", icon: <FolderIcon /> },
-  { label: "Budget", href: "/?budget", icon: <BudgetIcon /> },
 ];
 
+const debugNavLinks: NavLink[] = [
+  { label: "Hooks", href: "/hooks", icon: <ListIcon /> },
```
connor-parke-icf, line 25: love these links here, tired of typing in "/components" every time

client/src/layout/SideNav.tsx
```diff
@@ -3,32 +3,89 @@ import React from "react";
...
+  { label: "Components", href: "/components", icon: <FolderIcon /> },
+  { label: "Authentication", href: "/auth", icon: <ActionIcon /> },
+];
+
+interface NavLinkProps {
+  collapsed: boolean;
+  navLinks: NavLink[];
+}
+
+const NavLinks = (props: NavLinkProps) => {
```
connor-parke-icf, line 35: would this be worth breaking out into its own file?

connor-parke-icf: or a component for each navlink itself, and here theyre just stacked up and rendered

client/src/pages/debug/ComponentLibrary.tsx
```diff
@@ -1,23 +1,18 @@
...
-        <Collapsible title="Button Grid (Click to expand)">
-          <ButtonGrid />
-        </Collapsible>
-        <Collapsible title="Input Sandbox (Click to expand)">
-          <InputSandbox />
-        </Collapsible>
-      </Main>
-      <Footer />
-    </Fragment>
+    <>
```
connor-parke-icf, line 8: i can never tell when its appropriate to use <> vs just a regular ol <div>. I found myself moving a container div back and forth within and surrounding a component many times during building the header. Would be nice to establish some kind of best-practice... but also it doesnt really matter.

nunesmlucas (approved):

connor-parke-icf (approved):

client/src/layout/SideNav.tsx
tresdonicf: That's how I would implement it for sure, only really putting actual markup in a "leaf node" type thing but I didn't wanna overreach here

client/src/pages/debug/ComponentLibrary.tsx
tresdonicf: Yeah! So it's almost always best to return a fragment since it simplifies the CSS drastically to have less elements on the page and less nesting but I honestly forget that `<>` is a shorthand for `<Fragment>` hahah.

I'll bring this up in our office hours at some point. It's hard to know when to enforce this I guess since `<div>` can be helpful but generally I find people just fighting with that extra degree of nesting

dustbuster (approved):
Looks fine to me!

PR #13, opened 2025-06-02
DEMOS-167: DEMOS Header - MERGED IN https://github.com/Enterprise-CMCS/demos/pull/17

connor-parke-icf:
Created the Header and the blue subheader. Currently none of the links point to anything, nor any dropdown buttons. But built out the components. 

Utilized mockedProvider to grab some dummy data to insert where applicable, namely User fullName. Added some icons, and an extra button configuration. This does not include the demonstration glance quickview.


client/src/components/button/BaseButton.tsx
```diff
@@ -1,6 +1,6 @@
 import React from "react";
 
-export type ButtonSize = "standard" | "large";
+export type ButtonSize = "small" | "standard" | "large";
```
connor-parke-icf, line 3: added a "small" configuration as the button looked too big for the header when set to standard.

client/src/components/header/Header.tsx
```diff
@@ -1,10 +1,15 @@
...
+const Header: React.FC<{
+  userId?: number;
+}> = ({ userId }) => {
   return (
-    <header className="flex items-center h-fit bg-white text-black border-b-2">
-      <h1 className="text-4xl p-1">DEMOS</h1>
-    </header>
+    <div id="header-container" className="top-0 left-0 w-full z-11">
+      <HeaderUpper userId={userId} />
+      <HeaderLower userId={userId} />
```
connor-parke-icf, line 11: currently am passing in the userId through these components. I want to change this eventually to use a Context store, where we can have things like the current userId, or the current demonstration (if on a demonstration page). could be a separate story, or could be part of the lower-header's demonstration quickview.

client/src/hooks/userMocks.ts
```diff
@@ -41,4 +43,56 @@ export const userMocks = [
       data: { user: squidward },
     },
   },
+  {
```
connor-parke-icf, line 46: added some mocks for the queries used. I think as part of the ApolloProvider vs MockedProvider ticket, this will get cleaned up a bit.  I like the idea of queries being defined in the components themselves rather than in a central location, and being as specific as possible to their task at hand. So even though the HEADER_LOWER_QUERY and the PROFILE_BLOCK_QUERY are the same query (fetch the fullName from the Id), I think explicitly defining them in their component, named like their component, for each component that hits the graphql server, is the way to go about this. Yes, tat may lead to some redundancy, but it shifts the responsibility of asking exactly what is needed from a central location to directly next to the thing that is doing the requesting, the component.

client/src/layout/PrimaryLayout.tsx
```diff
@@ -9,14 +10,20 @@ const PrimaryLayout: React.FC<PrimaryLayoutProps> = ({ children }) => {
...
-    <div className="flex">
-      <SideNav collapsed={collapsed} setCollapsed={setCollapsed} />
-      <main
-        className={`transition-all duration-300 ${collapsed ? "ml-20" : "ml-64"
-        } w-full min-h-screen p-6 bg-gray-50`}
-      >
-        {children}
-      </main>
+    <div className="h-screen flex flex-col">
+      <Header userId={2}/>
```
connor-parke-icf, line 14: wasnt sure the best way to go about this. But since this is currently controlled by a bucket-brigade of id-passing instead of a Context parameter, i just set the userId at the same place the header is initialized.

client/src/router/DemosRouter.tsx
```diff
@@ -46,14 +48,16 @@ export default function DemosRouter() {
...
     <AuthProvider {...cognitoConfig}>
-      <BrowserRouter>
-        <Routes>
-          <Route path="/" element={<LandingPage />} />
-          <Route path="/login" element={<AuthComponent />} />
-          <Route path="/login-redirect" element={<AuthComponent />} />
-          {authenticatedRoutes}
-        </Routes>
-      </BrowserRouter>
+      <MockedProvider mocks={userMocks} addTypename={false}>
```
connor-parke-icf, line 51: The MockedProvider is what allows for us to query for data but not actually reach the graphql server. In theory, all we have to do is swap this out with an ApolloProvider and we will ping the server. The task in a different ticket is to wrap both in a component such that an env variable can control the swapping. 

Ideally, that would allow us to have perfectly consistent development data when desired, and that same data can be used in the Unit tests.

dustbuster:
Resolved conflcits by merged @connor-parke-icf 's code into my code an pushing a 3rd PR. 

https://github.com/Enterprise-CMCS/demos/pull/17

PR #18, opened 2025-06-04
DEMOS-300: pushing up artillery configuration and install for future use

connor-parke-icf:
Adding the artillery configuration used in the testing of DEMOS-300. Shouldnt effect anything, but will be good to have this in the codebase for future reference.


dustbuster (approved):
Sure! I'm not sure i know any better. But 🚀

PR #21, opened 2025-06-04
DEMOS-27, DEMOS-31, DEMOS-29: Some clean up on My demos table, added the ALL Demos table, tabs system, fixed up pagination and some style!

dustbuster:
Added this little tabbing and suggestions for th PR. 
Also made tabs a component for re-use

Screenshot 2025-06-05 at 15 56 10

There is one error in the console. But it's sidenav related, and i think we let is slide till we get those page built out. 

Screenshot 2025-06-06 at 09 22 17

PR #22, opened 2025-06-05
Replace default exports with named exports

tresdonicf:
Replaces `export default` with `export {namedVariable}` and fixes a couple of linting rules in the server to get these ready for CI/CD (both just unused vars)

image

### DIFF
--- a/client/src/components/header/Avatar.tsx
+++ b/client/src/components/header/Avatar.tsx
@@ -1,6 +1,6 @@
 import React from "react";
 
-const Avatar: React.FC<{ character?: string }> = ({ character }) => {
+export const Avatar: React.FC<{ character?: string }> = ({ character }) => {
   return (
     <div
       id="profile-letter"
@@ -18,5 +18,3 @@ const Avatar: React.FC<{ character?: string }> = ({ character }) => {
     </div>
   );
 };
-
-export default Avatar;
--- a/client/src/components/table/filters/ColumnFilterSelect.tsx
+++ b/client/src/components/table/filters/ColumnFilterSelect.tsx
@@ -70,5 +70,3 @@ export function ColumnFilterByDropdown<T extends object>({
     </div>
   );
 }
-
-export default ColumnFilterByDropdown;
--- a/client/src/components/table/pagination/PaginationControls.tsx
+++ b/client/src/components/table/pagination/PaginationControls.tsx
@@ -1,4 +1,3 @@
-// src/components/table/tables/PaginationControls.tsx
 import * as React from "react";
 
 export interface PaginationControlsProps {
@@ -101,5 +100,3 @@ export const PaginationControls: React.FC<PaginationControlsProps> = ({
     </div>
   );
 };
-
-export default PaginationControls;
--- a/client/src/components/table/tables/DemonstrationTable.tsx
+++ b/client/src/components/table/tables/DemonstrationTable.tsx
@@ -14,8 +14,8 @@ import {
   RowSelectionState
 } from "@tanstack/react-table";
 
-import PaginationControls from "components/table/pagination/PaginationControls";
-import ColumnFilterByDropdown from "components/table/filters/ColumnFilterSelect";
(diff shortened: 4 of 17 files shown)


client/vite.config.ts
```diff
@@ -21,3 +21,6 @@ export default defineConfig({
     },
   },
 });
+
+// eslint-disable-next-line no-restricted-exports
+export default config;
```
tresdonicf, line 26: An example of ignoring the rule when we need to and a good illustration of where tooling might need to support export default.

Since vite is going to be pluggable with other tooling `export default` is a good call because it means that libraries which use it will be able to continue using it even if the name changes

client/src/components/auth/AuthButtons.test.tsx
dustbuster, line 1: I'm sure you meant to put something here?

client/src/components/header/Avatar.tsx
```diff
@@ -1,6 +1,6 @@
 import React from "react";
 
-const Avatar: React.FC<{ character?: string }> = ({ character }) => {
+export const Avatar: React.FC<{ character?: string }> = ({ character }) => {
```
dustbuster, line 3: I have been resisting the urge to clean these up.

client/src/components/auth/AuthButtons.test.tsx
tresdonicf: Oh didn't realize this was still in my workspace! Good catch, maybe need a new linting rule for no empty files hahah

client/src/components/table/filters/ColumnFilterSelect.tsx
```diff
@@ -70,5 +70,3 @@ export function ColumnFilterByDropdown<T extends object>({
     </div>
   );
 }
-
-export default ColumnFilterByDropdown;
```
dustbuster, line 74: I already have this guy fixed [and i may have probably will cause a conflict. ](https://github.com/Enterprise-CMCS/demos/pull/21/files#diff-0e3de78436e20050af2ab4aee169573d651ff78494e7b6ab7b3895dbdb512ce9)

client/src/components/table/pagination/PaginationControls.tsx
```diff
@@ -1,4 +1,3 @@
-// src/components/table/tables/PaginationControls.tsx
```
dustbuster, line 1: This for sure is completely different in my PR. 100% conflict.

client/src/components/table/tables/DemonstrationTable.tsx
```diff
@@ -14,8 +14,8 @@ import {
   RowSelectionState
 } from "@tanstack/react-table";
 
-import PaginationControls from "components/table/pagination/PaginationControls";
-import ColumnFilterByDropdown from "components/table/filters/ColumnFilterSelect";
+import { PaginationControls } from "components/table/pagination/PaginationControls";
+import { ColumnFilterByDropdown } from "components/table/filters/ColumnFilterSelect";
```
dustbuster, line 18: Already fixed here https://github.com/Enterprise-CMCS/demos/pull/21/files

client/vite.config.ts
dustbuster: Rule are meant to be borked!

client/src/components/table/filters/ColumnFilterSelect.tsx
tresdonicf: If the changes are the same it should merge them fine! As in both lines are removed rather than some are removed and some are added creating a discrepancy

client/src/components/table/tables/DemonstrationTable.tsx
dustbuster: WAYYYY Different now.

dustbuster (approved):
i would ask that you leave the Datatable files out of the fix. I have it already fixed, and will reduce conflicts. 

But great work! Apprecaite the follow up!

PR #26, opened 2025-06-10
Demos-316: Set up the lambda entry point

james-becwar-cms:
Adds server.ts as entry point for the lambda.  Also refactors token decoding to work with the standalone and lambda entries.

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -6,6 +6,7 @@
     "schema": "./src/model"
   },
   "dependencies": {
+    "@as-integrations/aws-lambda": "^3.1.0",
     "@graphql-tools/merge": "^9.0.24",
     "@prisma/client": "^6.7.0",
     "depcheck": "^1.4.7",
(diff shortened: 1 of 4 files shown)


server/src/auth/auth.util.ts
```diff
@@ -33,22 +34,7 @@ type DecodedJWT = {
...
-): Promise<DecodedJWT> => {
-
-  // Bypass authentication for testing purposes
-  if (process.env.BYPASS_AUTH === "true") {
-    return Promise.resolve({
-      sub: "bypassed-user",
-      email: "bypassedUser@email.com"
-    });
-  }
-
```
dustbuster, line 50: So, we don't need this "stuff" anymore?

server/src/server.ts
connor-parke-icf, line 1: does this live side-by-side with the existing index.ts or was this meant to replace it?

server/package.json
```diff
@@ -26,7 +27,7 @@
     "seed": "npx prisma db push --force-reset && tsx --inspect ./src/seeder.ts"
   },
   "devDependencies": {
-    "@apollo/server": "^4.12.1",
+    "@apollo/server": "^4.12.2",
```
dustbuster, line 30: Upgrade!

connor-parke-icf (approved):

PR #29, opened 2025-06-10
DEMOS-353: Update server config to work on lambda

cms-jesse:


deployment/lib/lambda.ts
```diff
@@ -142,9 +142,9 @@ export class Lambda extends Construct {
             ? undefined
             : aws_apigateway.AuthorizationType.COGNITO,
           authorizer: props.isLocalstack ? undefined : props.authorizer,
-          authorizationScopes: props.isLocalstack
-            ? undefined
-            : ["demosApi/read", "demosApi/write"],
+          // authorizationScopes: props.isLocalstack
+          //   ? undefined
+          //   : ["demosApi/read", "demosApi/write"],
```
cms-jesse, line 147: Is there any reason we would want to have specific authorization scopes for the API?

server/package.json
```diff
@@ -16,18 +16,19 @@
     "jwks-rsa": "^3.2.0",
     "prettier": "3.5.3",
```
cms-jesse, line 17: I assume that `prettier` and `depcheck` should not be runtime dependencies? Is that correct, and are there any others in here that can be moved to devDependencies?

PR #30, opened 2025-06-10
DEMOS-291/293/294 - Modal for Creating a new demonstration

nunesmlucas:
https://jiraent.cms.gov/browse/DEMOS-293
https://jiraent.cms.gov/browse/DEMOS-291
https://jiraent.cms.gov/browse/DEMOS-294

### Major Changes

- Implemented fully functional `CreateNewModal` component with:
  - New Demonstration form fields
  - Inline validation logic (expiration date cannot be before effective date)
  - Form submission simulation with success & error states
  - Submission button loading state
  - Cancel confirmation modal (shared logic for both Cancel + X button close)

#### `client/src/components/modal/CreateNewModal.tsx`
- Full modal layout implementation matching Figma designs
- Used proper input components and select dropdowns
- Added form state management, success/error handlers, and cancel confirmation dialog

#### `client/src/components/modal/CreateNewModal.test.tsx`
- New test coverage for CreateNewModal using `vitest` + `@testing-library/react`
- Tests:
  - Rendering
  - Cancel confirmation behavior
  - Form validations
  - Expiration date validation
  - Successful submission simulation

#### `client/src/components/input/select/AutoCompleteSelect.tsx`
- Added proper `flex flex-col gap-sm` container for consistent label-input stacking
- Updated label rendering to use `LABEL_CLASSES` with full color + spacing consistency
- Full width (`w-full`) fix to allow responsive sizing in grid layouts

#### `client/src/components/input/select/SelectUSAStates.tsx`
- Refactored to match new `AutoCompleteSelect` structure
- Now accepts:
  - `label` prop
  - `isRequired`, `isDisabled`, `defaultValue`
- Fully consistent with other selects used in modal

#### `client/src/components/input/Input.tsx` + `TextInput.tsx`
- No structural changes but leveraged throughout `CreateNewModal` for consistent inputs
- Ensure consistent className behavior matches `tw` style approach

#### `client/src/components/header/HeaderLower.tsx`
- Integrated modal opening logic from dropdown into new `CreateNewModal` component
- Simplified state management for triggering modal

### What’s Working

- Full functional Create New Demonstration form flow
- Validation and error handling
- Success and error messaging on submit
- Cancel confirmation prompts
- Proper test coverage in place
- Fully responsive inside modal layout

### Known / Future Enhancements (Can Be Split Into Future Tickets)

- Plug backend API call into `handleSubmit`
- Wire up real user list for Project Officer dropdown
- Default Project Officer to currently logged in user
- Persist submitted demonstration records
- Reusable `CancelConfirmation` modal could be componentized

Related files modified
- `CreateNewModal.tsx`
- `CreateNewModal.test.tsx`
- `AutoCompleteSelect.tsx`
- `SelectUSAStates.tsx`
- `TextInput.tsx`
- `Input.tsx`
- `HeaderLower.tsx`


client/src/components/input/select/SelectUSAStates.tsx
```diff
@@ -1,26 +1,47 @@
...
-      label="Us State or Territory"
+      id={`state-${label.toLowerCase().replace(/\s+/g, "-")}`}
+      label={label}
       options={options}
-      placeholder="Select"
+      placeholder={`Select ${label.toLowerCase()}…`}
       onSelect={onStateChange}
       isRequired={isRequired}
       isDisabled={isDisabled}
+      defaultValue={defaultLabel}
```
dustbuster, line 44: This makes sense!

client/src/components/input/TextInput.tsx
```diff
@@ -1,15 +1,23 @@
 import React from "react";
-import { Input, InputProps } from "./Input";
+
+import {
+  Input,
+  InputProps
+} from "./Input";
```
dustbuster, line 6: Apparently we do trailing commas now. So okay then!

dustbuster (approved):
This looks good to me! 
Just get that trailing comma thing and make the CI/CD happy and let'r rip!

client/src/components/input/select/SelectUSAStates.tsx
```diff
@@ -1,26 +1,47 @@
...
     value: state.abbrev,
   }));
 
+  const defaultLabel = options.find((o) => o.value === currentState)?.label ?? "";
+
   return (
     <AutoCompleteSelect
-      id="us-state"
-      label="Us State or Territory"
+      id={`state-${label.toLowerCase().replace(/\s+/g, "-")}`}
```
james-becwar-cms, line 37: I'm not following why this should be dynamic?

client/src/components/modal/CreateNewModal.tsx
```diff
@@ -0,0 +1,200 @@
...
+        setFormStatus("error");
+      }
+    }, 2000);
+  };
+
+  return (
+    <div className="fixed inset-0 z-50 flex items-center justify-center bg-black/20">
+      <div
+        className="bg-[var(--color-surface-white)] border border-[var(--color-border-rules)] rounded shadow-lg w-[880px] max-w-[95vw]"
+        style={{ color: "var(--color-text-font)" }}
```
james-becwar-cms, line 55: Is this going to cause a CSP error?
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/style-src

james-becwar-cms (changes requested):

client/src/components/input/select/SelectUSAStates.tsx
james-becwar-cms: I'm not saying anything is wrong, just I didn't understand it.

dustbuster: Screenshot 2025-06-11 at 16 11 23

dustbuster: yeah aren't like IDs kidna like important for looking things up? The way remember is is that your option needs to be unique but it deosn't need to be the datas. `key=${`index` + 'name'}` or something. but there can't be more than one state and if there was it would still be the same state or a bug.

client/src/components/input/select/AutoCompleteSelect.tsx
```diff
@@ -103,27 +110,29 @@ export const AutoCompleteSelect: React.FC<AutoCompleteSelectProps> = ({
...
+            disabled:bg-[var(--color-surface-disabled)] disabled:text-[var(--color-text-placeholder)]
+            placeholder-[var(--color-text-placeholder)] focus:outline-none 
+            focus:border-[var(--color-border-focus)] focus:ring-1 focus:ring-[var(--color-border-focus)]
+            appearance-none text-sm
           `}
         />
 
         <div className="pointer-events-none absolute inset-y-0 right-3 flex items-center">
-          <ChevronDownIcon className="text-gray-500 w-2 h-1" />
+          <ChevronDownIcon className="text-[var(--color-text-placeholder)] w-2 h-1" />
```
tresdonicf, line 123: It would be nice to use the tailwind syntax here rather than the catch-all with the `[var(...)]` Nice job getting these to the colors in the designs though!

client/src/components/input/select/SelectUSAStates.tsx
tresdonicf: Should the value be the `abbrev` rather than the `name`? Since the former is being used as the "value" of the input

client/src/components/modal/CreateNewModal.test.tsx
tresdonicf, line 1: Nice tests!

client/src/components/input/Input.tsx
```diff
@@ -44,55 +46,67 @@ const getInputColors = (value: string, validationMessage: string) => {
...
+  label,
+  isRequired,
+  isDisabled,
+  placeholder,
+  defaultValue,
+  value: controlledValue,
+  onChange,
+  getValidationMessage,
+}) => {
+  const [uncontrolledValue, setUncontrolledValue] = useState(defaultValue ?? "");
```
tresdonicf, line 72: I think we should try to take this to just the single value. I think this would look like just calling this "value" and getting rid of the value in the props and using defaultValue as it's being used here (without passing the defaultValue into the plain `<input>`)

client/src/components/modal/CreateNewModal.tsx
tresdonicf, line 1: For a while I was confused by the name of this modal since it sounds like it is going to create a modal rather than be the modal itself but then I realized it's for creating a new demonstration hahah. Do you think it would be worth to add that piece to the name? I imagine maybe the plan was to re-use some of this for CreateNewExtension, CreateNewAmendment, etc which is probably a good idea and maybe a bridge we can cross when we come to it.

client/src/components/modal/CreateNewModal.tsx
```diff
@@ -0,0 +1,200 @@
...
+      if (isSuccess) {
+        setFormStatus("success");
+      } else {
+        setFormStatus("error");
+      }
+    }, 2000);
+  };
+
+  return (
+    <div className="fixed inset-0 z-50 flex items-center justify-center bg-black/20">
```
tresdonicf, line 52: It would be nice if some of this stuff was encapsulated in a generic `<Modal>` component and then just the content was inside as a child - that way we could essentially test the popout, close, etc, separate from the specific form that just happens to be in a modal.

Also there is a well supported [`<dialog>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dialog) in HTML at this point in case you want to check it out - might help to handle some open / close behavior for free

client/src/components/modal/CreateNewModal.tsx
```diff
@@ -0,0 +1,200 @@
...
+  const [showCancelConfirm, setShowCancelConfirm] = useState(false);
+  const [formStatus, setFormStatus] = useState<"idle" | "loading" | "success" | "error">("idle");
+
+  const isFormValid =
+    state &&
+    title &&
+    projectOfficer &&
+    effectiveDate &&
+    expirationDate &&
+    expirationDate >= effectiveDate;
```
tresdonicf, line 33: Not all of these are required for the form to be valid right? Based on the mocks I thought it might just be like 3 of them

client/src/components/modal/CreateNewModal.tsx
```diff
@@ -0,0 +1,200 @@
...
+      // Randomly succeed or fail for now
+      const isSuccess = Math.random() > 0.2;
+
+      if (isSuccess) {
+        setFormStatus("success");
+      } else {
+        setFormStatus("error");
+      }
+    }, 2000);
+  };
```
tresdonicf, line 49: I think this `setTimeout` piece can be deleted

client/src/components/modal/CreateNewModal.tsx
```diff
@@ -0,0 +1,200 @@
...
+          {formStatus === "success" && (
+            <div className="text-[var(--color-text-success)] text-sm pt-2">
+              Demonstration created successfully. (Placeholder Message)
+            </div>
+          )}
+          {formStatus === "error" && (
+            <div className="text-[var(--color-text-warn)] text-sm pt-2">
+              Failed to create demonstration. (Placeholder Message)
+            </div>
+          )}
```
tresdonicf, line 180: I think this can use the new toasts! Imported as below and then can either listen to `formStatus` change to call the appropriate one (or just call it directly in getting the return value from the api call)

```
const { showInfo, showSuccess, showWarning, showError } = useToast();
```

tresdonicf (commented):
Nice work on this! A few things I noticed that might be nice to change but overall it looks great 🥳 

1. The name of the modal could perhaps be more descriptive so it doesn't seem like a generic way to create a new modal
2. The tailwind names would be nice to use rather than `var()`
3. Could be nice to separate out the outer modal from the specific form contents

tresdonicf:
If you're having trouble with the linting errors lmk - Seems like there's still some odd interactions with the formatter and linter 🤦‍♀️

client/src/components/modal/CreateNewModal.tsx
nunesmlucas: Should be safe — React’s style={{ ... }} prop doesn’t count as unsafe inline styles under CSP, unless we had very strict style-src rules that block all inline styles entirely..

client/src/components/modal/CreateNewModal.tsx
nunesmlucas: Yep — makes sense, this was just simulating an async call during development. Once we hook it up to real API logic, we can safely drop the `setTimeout`.

tresdonicf (approved):
This looks great, nice work!

james-becwar-cms (approved):

PR #33, opened 2025-06-11
Eslint: Fix comma-dangle config

tresdonicf:
The prettier config setting for `trailingComma: es5` was at odds with our `comma-dangle` config as imports need to be specifically declared as being expected to have a trailing comma where I thought this was covered by the `objects` key before.

Formatted some files while kicking the tires

### DIFF
--- a/client/eslint.config.js
+++ b/client/eslint.config.js
@@ -27,7 +27,7 @@ const eslintConfig = tseslint.config(
       "no-trailing-spaces": "error", // Disallow trailing spaces
       "no-tabs": "error", // Disallow tabs for indentation
       "eol-last": ["error", "always"], // Newline at the end of files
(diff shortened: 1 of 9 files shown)


client/eslint.config.js
```diff
@@ -27,10 +27,16 @@ const eslintConfig = tseslint.config(
...
       "comma-dangle": [
         "error",
-        { arrays: "always-multiline", objects: "always-multiline" },
-      ], // Trailing commas in multiline arrays and objects
+        {
+          arrays: "always-multiline",
+          objects: "always-multiline",
+          imports: "always-multiline",
+          exports: "always-multiline",
+        },
```
tresdonicf, line 38: And prettier put them all on newlines - which is prettier!

nunesmlucas (approved):

PR #37, opened 2025-06-13
Demos 295: Conditional MockedProvider / ApolloProvider

tresdonicf:
This change adds support for conditionally using Apollo's `MockedProvider` versus the main `ApolloProvider` depending on environment and a new environment variable: `VITE_USE_MOCKS`

- Test: Always use mocks
- Dev: Use mocks if VITE_USE_MOCKS is true
- Prod: Never use mocks

### DIFF
--- a/client/src/hooks/useUserOperations.ts
+++ b/client/src/hooks/useUserOperations.ts
@@ -1,5 +1,5 @@
 import { gql, useLazyQuery } from "@apollo/client";
-import { User } from "models/index";
+import { User } from "demos-server";
 
 export const GET_ALL_USERS = gql`
   query GetUsers {
--- a/client/src/hooks/useUserOperations.ts
+++ b/client/src/hooks/useUserOperations.ts
@@ -4,8 +4,7 @@ import { User } from "demos-server";
 export const GET_ALL_USERS = gql`
   query GetUsers {
     users {
-      firstName
-      lastName
+      fullName
     }
   }
 `;
@@ -13,9 +12,7 @@ export const GET_ALL_USERS = gql`
 export const GET_USER_BY_ID = gql`
   query GetUser($id: ID!) {
     user(id: $id) {
-      id
-      firstName
-      lastName
+      fullName
     }
   }
 `;
--- a/client/src/mock-data/index.test.ts
+++ b/client/src/mock-data/index.test.ts
@@ -1,0 +1,24 @@
+import { describe, it, expect } from "vitest";
+import { print } from "graphql";
+import { ALL_MOCKS } from "./index";
+
+describe("ALL_MOCKS", () => {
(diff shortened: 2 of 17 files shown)


client/src/mock-data/userMocks.ts
connor-parke-icf, line 1: As the site gets bigger and we want more things to ping against, it may be a good idea to create a centralized json file that stores a bunch of mock data as a kind of pseudo-database, that the mocked data providers can pull from. So for example, we would have a Spongebob user with all the possible user fields having some kind of data in it, then in the mocked provider we just pull the things we need to match the mocked provider's example returns. May be a good thing to lump into the idea of unifying the style/theme of our test data so it appears more fluid on quick glance

client/src/router/DemosApolloProvider.tsx
connor-parke-icf, line 1: huge fan of this; being able to switch the entire site from "real" mode to "fake" mode is excellent

connor-parke-icf (approved):

client/src/hooks/useUserOperations.ts
```diff
@@ -1,21 +1,18 @@
 import { gql, useLazyQuery } from "@apollo/client";
-import { User } from "models/index";
+import { User } from "demos-server";
```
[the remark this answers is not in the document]
tresdonicf: Yeah exactly! Along with the `export` changes in the server's `package.json`

client/src/mock-data/index.test.ts
```diff
@@ -0,0 +1,24 @@
...
+
+describe("ALL_MOCKS", () => {
+  it("should have a request and result for each mock", () => {
+    ALL_MOCKS.forEach((mock) => {
+      expect(mock.request).toBeDefined();
+      expect(mock.result).toBeDefined();
+    });
+  });
+
+  it("should not have overlapping queries", () => {
```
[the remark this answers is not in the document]
tresdonicf: Interesting - I couldn't find mention of this concept in the docs but it sounds interesting - could you share where you found this by chance? I imagine the same query with the same parameters at the same time should return the same thing but it could feasibly return different results at different points in the application lifecycle for sure

client/src/mock-data/userMocks.ts
tresdonicf: I think this would be cool too! Would love to get a spike on how we can declare this to keep us from a few time-sinks that I could see happening:

1. Having to reference existing entities to make sure new entities are ontologically consistent with that system
2. Having to cascade mock changes on db / schema migration

There might be some other things to this but yeah it would be really cool and we _should_ be able to use some of our `User` types in our `Demonstration` (etc) schemas to get the cascaded changes for free! I do wonder if the effort of building the consistent ontology is worth the squeeze but super open to it

PR #41, opened 2025-06-16
DEMOS-74-duckdb-experimentation

TARehman:
This is the result of researching the use of DuckDB as part of the broader ETL of the legacy system to the new PostgreSQL option. This is a spike, not an actual implementation, but I've tagged folks who may want to review it. I will also discuss tomorrow at our office hours. It took me about 5 minutes to rip the whole old database over to the new PG one.

I was using Python 3.13 for this - there's no DevContainer so we are probably going to be facing some "runs on my machine" issues until we create one. But, for a spike I felt it was acceptable.


data/research/duckdb_exporter/duckdb_exporter.py
```diff
@@ -0,0 +1,314 @@
...
+        trace (Optional[types.TracebackType]): The traceback object.
+    """
+    logger.error(
+        "Unhandled exception",
+        exc_info=(e_type, val, trace)
+    )
+    return None
+
+
+sys.excepthook = custom_excepthook
```
tresdonicf, line 306: This is super cool, had no idea you could do this!

data/research/duckdb_exporter/duckdb_exporter.py
```diff
@@ -0,0 +1,314 @@
...
+    "smallint": "smallint",  # Putting this here avoids issues with unsigned items in MySQL
+    "timestamp": "timestamptz",
+    "tinyint": "smallint",
+}
+DUCKDB_MYSQL_DB_NAME = "mysql_db"
+DUCKDB_POSTGRES_DB_NAME = "postgres_db"
+PG_SCHEMA = "legacy_pmda_raw"
+MYSQL_SCHEMA = "cma_dev_11_1_000"
+
+Path("logs").mkdir(parents=True, exist_ok=True)
```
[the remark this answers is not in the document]
TARehman: Yeah, same. Kudos to Claude/CoPilot for that one. 😛

data/research/duckdb_exporter/duckdb_exporter.py
TARehman: It was annoying to discover this is apparently the "right" way to route stuff that goes to `stderr` to your logs, but the more you know! 🤷🏻

PR #42, opened 2025-06-16
Demos 304: Clean up Client Env Configuration

tresdonicf:
**tl;dr: Calls to `process.env` are no longer scattered about in the client, calls to `import.meta.env` (Vite's environment variable system) are consolidated, `npm run preview` works again 🥳**

In my exploration of the spike regarding user authentication state I learned a bit more about the proper way to [manage environment variables in vite](https://vite.dev/guide/env-and-mode). While we could keep using `process.env.NODE_ENV` this gives us a bit more consistency in usage and flexibility with adding variables and documenting them in the codebase.

Talked to Jesse about the `window._env_` approach for some of the cognito vars to see if we can incorporate this same fold since this seems to be not working when we run `npm run preview` (the deploy file doesn't exist locally). For now I've put these variables into the vite's environment variable system but there's some figuring out to do regarding whether we know the values we'll need to use for the cognito configuration before the deploy happens, which would be required for build-time deploys.


client/src/config/env.ts
```diff
@@ -0,0 +1,33 @@
...
+  return getAppMode() === "test";
+};
+
+export const isProductionMode = (): boolean => {
+  return getAppMode() === "production";
+};
+
+export const shouldUseMocks = (): boolean => {
+  return (
+    isTestMode() || (isDevelopmentMode() && import.meta.env.VITE_USE_MOCKS)
```
nunesmlucas, line 31: Should we add a guard before `shouldUseMocks` to check that `VITE_USE_MOCKS` exists first? or is it something that will always be defined? Similar to this, if not the same, `isTest || (isDevelopment && import.meta.env.VITE_USE_MOCKS === "true");` which was in `DemosApolloProvider.tsx`

tresdonicf: Good catch! This was set up to use the mere presence of it as truthiness and I wish there was a way that was supported natively in environment variables but it's probably more principled to compare this to `"true"`

PR #43, opened 2025-06-17
DEMOS 355

JeffreyM-GA:
Adds CheckboxGroup.tsx to component library. Accepts configurations for options, disabled/enabled, validation, and inline display.

image


client/src/components/checkboxGroup/CheckboxGroup.test.tsx
```diff
@@ -0,0 +1,110 @@
...
+    const input = screen.getByLabelText("Option B");
+    expect(input).toBeChecked();
+  });
+
+  it("updates values on change", () => {
+    render(<CheckboxGroup {...requiredProps} />);
+    const input = screen.getByLabelText("Option B");
+    fireEvent.click(input);
+    expect(input).toBeChecked();
+  });
```
tresdonicf, line 60: nit: A test for checking our `selectedValues` field is accurate after a sequence of changes could be helpful since I think this will mostly just be handled by the browser

tresdonicf (approved):
Looks good, nice work!

PR #51, opened 2025-06-23
DEMOS-253 - Updates Demos sort order

nunesmlucas:
###  DEMOS-253: Update Default Sort Order for Demonstrations Table

This PR implements the default sorting behavior for both **My Demonstrations** and **All Demonstrations** tabs based on user feedback.

###  Changes Made
- Applied default sorting state in `DemonstrationTable.tsx`:
  - Primary sort: `State/Territory (stateId)` ascending
  - Secondary sort: `Title (title)` ascending

```ts
const [sorting, setSorting] = React.useState<SortingState>([
  { id: "stateId", desc: false },
  { id: "title", desc: false },
]);

### DIFF
--- a/client/src/components/table/tables/DemonstrationTable.tsx
+++ b/client/src/components/table/tables/DemonstrationTable.tsx
@@ -1,23 +1,33 @@
 import * as React from "react";
+
 import {
-  flexRender,
-  useReactTable,
-  getCoreRowModel,
-  getSortedRowModel,
-  getFilteredRowModel,
-  getExpandedRowModel,
-  getPaginationRowModel,
-  SortingState,
+  DemonstrationColumns,
+} from "components/table/columns/DemonstrationColumns";
+import {
+  ColumnFilterByDropdown,
+} from "components/table/filters/ColumnFilterSelect";
+import {
+  PaginationControls,
+} from "components/table/pagination/PaginationControls";
+import {
+  DemoWithSubRows,
+  groupByDemoNumber,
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Review by tresdonicf — APPROVED** [review]

File: client/src/components/table/tables/DemonstrationTable.tsx [hunk 1]
```diff
@@ -124,7 +137,7 @@ export function DemonstrationTable({
                 >
                   {flexRender(header.column.columnDef.header, header.getContext())}
                   {{
-                    asc:  " ↑",
+                    asc: " ↑",
```
connor-parke-icf, line 140: i believe we have a couple of icons for arrows like this. Not sure if its applicable here though, whether we can put Components in this spot.

client/src/components/table/tables/DemonstrationTable.tsx
```diff
@@ -1,24 +1,34 @@
...
+import {
+  PaginationControls,
+} from "components/table/pagination/PaginationControls";
+import {
+  DemoWithSubRows,
+  groupByDemoNumber,
+} from "components/table/preproccessors/GroupByDemoNumber";
+
 import {
+  ColumnFiltersState,
```
connor-parke-icf, line 18: curious why this was moved around, was it a prettier configuration/auto-lint?

connor-parke-icf (approved):
nothing big, but one comment about those arrow characters and the arrow icon

PR #61, opened 2025-07-01
DEMOS-263 add DemonstrationDetail page, make lowerHeader context sens…

JeffreyM-GA:
…itive

- Adds "usePageHeader" hook to set the page level header
- Adds demonstrationDetail route, with initial setup for viewing demonstration details in header

image

### DIFF
--- a/client/src/components/header/DefaultHeaderLower.test.tsx
+++ b/client/src/components/header/DefaultHeaderLower.test.tsx
@@ -1,4 +1,4 @@
-// src/components/HeaderLower.test.tsx
+// src/components/DefaultHeaderLower.test.tsx
 import React from "react";
 
 import { vi } from "vitest";
@@ -10,7 +10,7 @@ import {
   screen,
 } from "@testing-library/react";
 
-import { HeaderLower } from "./HeaderLower";
+import { DefaultHeaderLower } from "./DefaultHeaderLower";
 
 // Mock Apollo
 vi.mock("@apollo/client", async () => {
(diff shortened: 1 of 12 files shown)


nunesmlucas:
Looks like the header is changed and being squished a bit. 
image
Main branch:
image

nunesmlucas (changes requested):
I'm not sure if it's because I am running this with `npm run dev:mocks` but I am unable to display the details in the top header. Also for what it's worth, maybe we an define/adjust the styling to remain the same `height` outside of these failure messages so that doesn't change. You can see its shorter, more narrow. 

Sorry if I'm being a pain Jeff, let me know what you think. 

image

JeffreyM-GA:
> I'm not sure if it's because I am running this with `npm run dev:mocks` but I am unable to display the details in the top header. Also for what it's worth, maybe we an define/adjust the styling to remain the same `height` outside of these failure messages so that doesn't change. You can see its shorter, more narrow.
> 
> Sorry if I'm being a pain Jeff, let me know what you think.
> 
> image

JeffreyM-GA:
"Also for what it's worth, maybe we an define/adjust the styling to remain the same height outside of these failure messages so that doesn't change. You can see its shorter, more narrow."

I did think the same thing regarding height, we should probably set it to a static height. I think that would make sense to handle as a separate story IMO.

nunesmlucas:
Looks great overall — clean context setup and solid integration with the new page and routes. Nice use of `setHeaderConfig` and fallback content.

PR #62, opened 2025-07-01
DEMOS-423-add-lambda-secretsmanager-items

TARehman:
This PR is essentially two features in one: updating our local development environment to use `localstack` for the server, and adding support to the server Lambda to get the authentication details from SecretsManager and use them to talk to the database. I'm writing a bit more here just for context.

# Devcontainer Changes

This changes the devcontainer setup in a few ways.

- Installs the `awscli` package, which is used by the `localstack` setup.
- Removes port forwarding from the actual devcontainer setup and moves it to the Docker compose.
- Adds a proper post-create script, needed for AWS configuration. In that script, we set up the `awscli` profile for `localstack` and alias the `aws` command to include the endpoint (for some reason, none of the other ways of setting that up are honored).
- Adds a setting regarding auto forwarding of ports to the devcontainer. This just makes the VSCode interface less cluttered with extra ports added.

# Localstack Implementation

Localstack is set up using the script `server/localstack/localstack.sh`. Most of this is pretty straightforward configuration of the `localstack` environment.

- It checks to see if the deployment bucket, the Lambda function, the API Gateway, and the secret exist, and if they do, it drops them and recreates them.
- Note that in the build process, it is necessary to grab both the minified JSON file, _and_ the binaries used by Prisma; these do not exist in the Lambda runtime, and so that is the reason to grab them from the `server/node_modules/.prisma/client` folder.

When running `localstack.sh`, the last thing output is the URL of the API endpoint. Putting this into Bruno or something similar will allow access to run queries. Note that if you were using it from _inside_ the `app` machine, you would want to make the URL `localstack`.

# Lambda Implementation

The changes to the server code add a new function to retrieve a secret from SecretsManager, and then modify the Lambda to set the `DATABASE_URL` environment variable.

Note that at the time of Lambda startup, Prisma will be unhappy if there is not _some_ value of `DATABASE_URL` set - this is why the `localstack.sh` file sets it to `postgres://placeholder`. Note also that Prisma will be unhappy if the value doesn't start with `postgres://` which is why the version above is used.

The function that retrieves the secret performs very basic caching of the secret with a 1-hour timeout. This will only be necessary if the Lambda context is reused, where it will decrease the amount of calls to SecretsManager that we'll need to make. We may want to think about what will happen when secrets rotate, as a cached value could remain incorrect for up to an hour depending.

In the Lambda, the `getDatabaseUrl` function is used to create the `databaseUrlPromise` which basically just sets the `DATABASE_URL` environment variable. Arguably, there's no need to return the `url` value there in the anonymous function on Line 15 - not sure if we want to remove that or not. Then, we just await the `databaseUrlPromise` in the Lambda setup.

# Other Changes

The changes to these three files revert a previous change which had broken the server from working. This still needs attention per comments from @connor-parke-icf, but just was not in scope for this ticket.

- `server/src/model/demonstrationStatus/demonstrationStatusSchema.ts`
- `server/src/model/state/stateSchema.ts`
- `server/src/model/user/userSchema.ts`

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,6 +1,13 @@
 FROM mcr.microsoft.com/devcontainers/javascript-node:1-22-bookworm
 
-RUN apt-get update && export DEBIAN_FRONTEND=noninteractive  && apt-get -y install python3 python3-pip python-is-python3 && pip3 install pre-commit --break-system-packages
+RUN apt-get update && \
+    export DEBIAN_FRONTEND=noninteractive && \
+    apt-get -y install \
+        python3 \
+        python3-pip \
+        python-is-python3 \
+        awscli && \
+    pip3 install pre-commit --break-system-packages
 
 # [Optional] Uncomment if you want to install an additional version of node using nvm
 # ARG EXTRA_NODE_VERSION=10
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -19,10 +19,10 @@
 
     // Use 'forwardPorts' to make a list of ports inside the container available locally.
     // This can be used to network with other containers or with the host.
-    "forwardPorts": [3000, 4000, 5432, 9092, 9093],
+    "forwardPorts": [3000, 4000, 4566, 5432, 9092, 9093],
 
     // Use 'postCreateCommand' to run commands after the container is created.
-    "postCreateCommand": "pre-commit install",
+    "postCreateCommand": "zsh .devcontainer/post_create.sh",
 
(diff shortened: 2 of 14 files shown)


TARehman:
I put this as a draft while I investigated why the configuration of the localstack stuff was not persisting between restarts of the devcontainer. The simple answer is that persistence turns out to be a pro feature of localstack and just doesn't work if you are running on the free version. This means that for regular use, everyone will need to run the `localstack.sh` script every time they run the devcontainer, regardless of whether there are build differences or not.

dustbuster (approved):

PR #72, opened 2025-07-09
Demos-449: follow up, update frontend tests. 

dustbuster:
There was a comment to include my dumb json tables in the unit tests. So i did that and updated the unit tests. 

Screenshot 2025-07-09 at 16 25 47


dustbuster (commented):
Thanks Tres!

PR #73, opened 2025-07-09
DEMOS-322: Event Logging Prototype

tresdonicf:
JIRA: DEMOS-322

This PR does a few things:

1. Add prisma tables / GQL models for  `event / event_type` (+`Event / EventType`)
2. Adds some plaintext exploration around this concept in the `spikes/` folder. Definitely looking for some input on this!

Additionally:

- Add a `.code-workspace` file the server to focus on the files needed when editing server code and minimizing some noise.
- I ran `npx prisma format` to add some relations and it re-ordered some of the stuff in the other prisma files. I left as is but can revert if desired. I figured using the tool means that others will get consistent results in future runs.


server/src/model/event/eventSchema.ts
```diff
@@ -0,0 +1,60 @@
...
+  eventTypeId: string;
+  createdAt: Date;
+  eventData: object;
+}
+
+export interface CreateEventInput {
+  userId: string;
+  eventTypeId: string;
+  eventData: object;
+}
```
dustbuster, line 60: Newline! 
I have an .editorconfig file should auto add newlines at the end of files on save.  
```
# install editorconfig extension to use
[*]
insert_final_newline = true
indent_style = space
indent_size = 2
```
Pretty hand little tool.

server/src/model/event/index.ts
```diff
@@ -0,0 +1,4 @@
+export * from "./eventSchema.js";
+export * from "./eventResolvers.js";
+export * from "./eventTypeSchema.js";
+export * from "./eventTypeResolvers.js";
```
dustbuster, line 4: Can you do like `export * from "./*"`? Maybe is safer to explicitly define them. Just curious.

server/src/model/state/state.prisma
```diff
@@ -1,3 +1,9 @@
+/**
+TODO: We probably don't want to use UUIDs for state IDs.
+We can use the state code as the primary key instead.
+This will allow us to use natural keys, which are more meaningful and easier to work with.
+This also avoids the problem of regenerating these IDs and invalidating existing records that reference them.
+*/
```
dustbuster, line 6: This is the way! We talked about this at some point and came to same conclusion.

server/src/model/event/index.ts
tresdonicf: It's not natively supported by ESModules though I do think some build tools should be able to do this, I wonder what effects that would have on the intellisense since it would need to `ls` the files in the directory rather than just walk the tree and build references

server/src/model/event/eventSchema.ts
tresdonicf: That is awesome! I think we should be able to set up prettier on the server here soon too if we want to go that way, now that folks seem to have it set up on the client! I didn't add it here when I added it to the client since I have less idea of the coding standards the team wants to maintain here

server/src/model/event/eventResolvers.ts
```diff
@@ -0,0 +1,89 @@
...
+  Mutation: {
+    createEvent: async (_: undefined, { input }: { input: CreateEventInput }, context: GraphQLContext) => {
+      const { eventTypeId, route, eventData } = input;
+      
+      const userId = await getCurrentUserId(context);
+      const withRoleId = await getCurrentUserRoleId(context);
+      
+      return await prisma().event.create({
+        data: {
+          userId,
```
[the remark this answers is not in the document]
tresdonicf: This is a good question, it's probably worthwhile to if we're returning all these others.

On the other hand, I don't know if we should return any of these fields since this will be ran in `logEvent()` and maybe the user shouldn't be able to inspect any of the returned details from that request for security reasons. I suppose I'm kind of leaning towards that now since the user / client shouldn't need to react to an event being logged on their behalf. Kinda seems like leaving it out is good need-to-know practice but curious what you think regarding these two approaches.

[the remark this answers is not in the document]
tresdonicf: Yeah valid! I'll add a change regarding this for the next PR addressing `DEMOS-315` to return the ID or another type of success measurement. I do think that's important to be able to tell it was written correctly but I wonder if we want to give the user a handle this or in the resolver just return some type of `"Event was written" / "Sadly it was not"`

PR #75, opened 2025-07-11
DEMOS-265 Edit Demonstration Details from Top Nav

JeffreyM-GA:
Adds Edit Functionality to CreateNewModal including autopopulating controlled fields
Renames CreateNewModal
Associated Tests


client/src/components/modal/DemonstrationModal.tsx
```diff
@@ -9,14 +8,26 @@ import { SelectUsers } from "components/input/select/SelectUsers";
...
 
 const LABEL_CLASSES = tw`text-text-font font-bold text-field-label flex gap-0-5`;
 const DATE_INPUT_CLASSES = tw`w-full border rounded px-1 py-1 text-sm`;
 
-export const CreateNewModal: React.FC<{ onClose: () => void }> = ({ onClose }) => {
+type DemonstrationModalMode = "add" | "edit";
+
+type Props = {
+  onClose: () => void;
+  demonstration?: Demonstration; // TODO: replace with actual type
```
connor-parke-icf, line 22: is this still TODO? i believe we have a Demonstration type somewhere in server, though ideally i think id like to get a client version of the type

client/src/components/modal/DemonstrationModal.tsx
```diff
@@ -28,10 +39,21 @@ export const CreateNewModal: React.FC<{ onClose: () => void }> = ({ onClose }) =
   const [formStatus, setFormStatus] = useState<"idle" | "pending">("idle");
 
   const { showSuccess, showError } = useToast();
-  const { addDemonstration } = useDemonstration();
+  const { addDemonstration, updateDemonstration } = useDemonstration();
```
connor-parke-icf, line 42: this is fine; though imo i would still like to lean us away from centralized queries and mutators, to align better with the design patterns of graphql.  This is part of a larger discussion on the architecture of the project, so no need to change anything here.

client/src/hooks/useDemonstration.test.ts
connor-parke-icf, line 1: Same idea here. I would like to have the responsibility placed on each component to use and test its own fetching rather than having a centralized collection of common functions. Not every component may use the same saving behavior or require the same data from a query. That being said, this is totally fine; just my opinion overall

client/src/components/modal/DemonstrationModal.tsx
dustbuster: The power of a ORM is pulling giant queries out of models! But agree with Conner. We can come back to this.

PR #76, opened 2025-07-11
Demos-503: Add non-prod private hosted zones

cms-jesse:
There are 2 changes here:

1. This adds the private hosted zones for all non-prod environments. These will be synced with CMS Infoblox so that they can be resolved while on ZScaler. Since these will never change and will likely be shared between main and ephemeral environments, I put them in the bootstrap stack that is only run at the account level
2. There is another change included for a Jenkins role. This is the role assumed by the Jenkins pipelines and has been in use, but the changes were never pr'd

### DIFF
--- a/deployment/stacks/bootstrap.ts
+++ b/deployment/stacks/bootstrap.ts
@@ -1,7 +1,8 @@
-import { Stack, StackProps, aws_iam, aws_apigateway } from "aws-cdk-lib";
+import { Stack, StackProps, aws_iam, aws_apigateway, aws_ec2 } from "aws-cdk-lib";
 import { Construct } from "constructs";
 
 import { DeploymentConfigProperties } from "../config";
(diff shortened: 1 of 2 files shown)


deployment/stacks/bootstrap.ts
```diff
@@ -113,5 +135,27 @@ export class BootstrapStack extends Stack {
...
+    createPHZ(commonProps.scope, "impl")
   }
 }
+
+function createPHZ(scope: Construct, env: string) {
+  const dnsSuffix = "demos.internal.cms.gov"
+
+    const devVpc = aws_ec2.Vpc.fromLookup(scope, `${env}Vpc`, {
+          tags: {
+            Name: `demos-east-${env}`,
```
dustbuster, line 154: No cross region redundancies? lol

PR #79, opened 2025-07-11
explicitly passing in dates as date objects rather than strings

connor-parke-icf:


server/src/model/demonstration/demonstrationResolvers.ts
```diff
@@ -22,10 +22,20 @@ export const demonstrationResolvers = {
...
+        projectOfficerUserId, 
+        evaluationPeriodStartDate, 
+        evaluationPeriodEndDate, 
+        ...rest 
+    } = input;
       return await prisma().demonstration.create({
         data: {
           ...rest,
+          evaluationPeriodStartDate: new Date(evaluationPeriodStartDate),
+          evaluationPeriodEndDate: new Date(evaluationPeriodEndDate),
```
tresdonicf, line 38: Do we need to do this for `@db.Timestamptz()` fields like `createdAt` and `updatedAt` or just the `@db.Date` fields? I wonder if there's a way to do tell prisma / GQL to handle everything of these types in this way without doing it for each resolver

tresdonicf (approved):

server/src/model/demonstration/demonstrationResolvers.ts
connor-parke-icf: the original issue with these fields specifically was that theyre "Date" fields and not timestamptz. The input for the timestamptz fields would by nature contain all of the information that prisma was expecting (a time and timezone).  I expect the only time we need to do this is when we are storing a simple date. Furthermore, the createdAt and updatedAt fields are generally being handled without user input, so they shouldnt be a concern. I dont know if theres a better way to handle this globally with a middleware or something or not, but if we end up having to do this for multiple endpoints then we can take another pass to see if we can have a centralized data coercion function

PR #82, opened 2025-07-14
DEMOS-305-doc-bundle-experiment

TARehman:
- Adds in the concept of a document "bundle" and implements it for demonstrations.
   - Adds relevant tables and triggers.
   - Modifies the API so that when a new demonstration is created, the bundle is created first and then linked.
   - Adds in some constants and types based on them, to reflect a move toward natural keys for some things (this can be expanded on in future PRs).
   - Fixes the seeder to reflect this change as well.
 - Updates the `npm run build` command to generate the Prisma client first, which helps to avoid mismatch issues (this was already part of `npm build:ci`).
 - Incorporates the data model documentation into the repository.
   - Adds a package to the devcontainer for dealing with Mermaid files.
   - Adds the version 5 of the data model here for future tracking. Note that this is "aspirational" and needs to be reviewed and checked in future PRs (to both match what we currently have, and clearly mark what is implemented and what is not yet).
   - Sets the pre-commit hook to 1024KB so I could upload the PNG file.


server/src/model/bundle/bundle.prisma
```diff
@@ -0,0 +1,10 @@
+model Bundle {
+  id           String @id @default(uuid()) @db.Uuid
+  bundleTypeId String @map("bundle_type_id")
+
+  bundleType    BundleType     @relation(fields: [bundleTypeId], references: [id])
+  demonstration Demonstration?
```
james-becwar-cms, line 6: Should this have an @relation?  I find prisma relationships so weird.

server/src/model/bundleType/bundleTypeHistory.prisma
```diff
@@ -0,0 +1,12 @@
+model BundleTypeHistory {
+  revisionId   Int          @id @default(autoincrement()) @map("revision_id")
+  revisionType RevisionType @map("revision_type")
+  modifiedAt   DateTime     @default(now()) @map("modified_at") @db.Timestamptz()
+  id           String
```
james-becwar-cms, line 5: I'm confused by this line, isn't revisionId marked as the PK?

server/src/model/demonstration/demonstration.prisma
```diff
@@ -1,5 +1,6 @@
 model Demonstration {
-  id                        String   @id @default(uuid()) @db.Uuid
+  id                        String   @db.Uuid
```
james-becwar-cms, line 2: How come we don't want the @default(uuid())?

server/src/model/demonstration/demonstration.prisma
```diff
@@ -1,5 +1,6 @@
...
+  bundleTypeId              String   @map("bundle_type_id")
```
james-becwar-cms, line 3: I'm curious if prisma adds this to the demonstration?

james-becwar-cms (commented):
Had a few questions.  I promise I will read the prisma docs some day.

Also I think we should remove the data model png since the mmd is there and github it self can render it. (and the vs code extension)

server/src/model/bundle/bundle.prisma
TARehman: No, because it's the receiving end of a relationship, not the providing end. I don't claim to fully understand it myself, but I think its just saying that demonstrations will have bundles, because there's an FKEY from demonstration back to bundle.

server/src/model/demonstration/demonstration.prisma
TARehman: Now that you need a bundle first, you no longer need or want a default value to be generated here; we need to get the bundle ID first and then insert it. That being said, I assumed this did something with the SQL that is generated, and it turns out it is entirely internal to Prisma, so it might not matter at all.

server/src/model/demonstration/demonstration.prisma
TARehman: Adds what? If you mean the bundleTypeId line, the answer is no, I had to add that. But I don't know if it automagically would produce ID columns if you set the relations correctly; I just sort of circumvent the process.

server/src/model/bundleType/bundleTypeHistory.prisma
james-becwar-cms: since ID is the PK in every other table, I think we should rename this column to avoid confusion.

james-becwar-cms (approved):

PR #83, opened 2025-07-14
DEMOS-315 - Add client function for `logEvent`

tresdonicf:
Addressing DEMOS-315 - adds a function `logEvent` that will allow us to log events across the client-side application. Add an `events` tab to our below-the-fold development tools that will show all events that have been logged in a given user session

https://github.com/user-attachments/assets/1f84713b-1197-4a2c-8f4f-89652493f614


tresdonicf:
Putting this back in draft to reflect changes made in #87

server/src/seeder.ts
TARehman, line 1: Hey: I made a bunch of changes to the seeder in the PR I just opened. It might make sense to pull / merge my changes in here to avoid colliding with the ones I saw you making too. I like the CONST approach, would be cool to merge it into what I had put together today.

tresdonicf, line 1: Think this will change a bit after the updates from #89

client/src/hooks/event/eventTypes.ts
```diff
@@ -0,0 +1,22 @@
+export type LogLevel = "INFO" | "ERROR";
```
TARehman, line 1: Relevant nit relating to the other point I'm raising about log levels - check the current `main` for the contents of the table, which should be aligned to the syslog standards. Assuming we actually want to have the log level table per @james-becwar-cms.

server/src/model/migrations/20250721204840_event_updates/migration.sql
```diff
@@ -0,0 +1,41 @@
...
+DROP COLUMN "active_user_id",
+DROP COLUMN "event_type_id",
+DROP COLUMN "log_level_id",
+DROP COLUMN "role_id",
+ADD COLUMN     "event_type" TEXT NOT NULL,
+ADD COLUMN     "log_level" TEXT NOT NULL,
+ADD COLUMN     "with_role_id" UUID NOT NULL;
+
+-- DropTable
+DROP TABLE "log_level";
```
TARehman, line 35: My understanding is that @james-becwar-cms wanted there to be a `log_level` constraint table so that we could facilitate things like "Filter by all levels greater than WARN" or something like that, which is what it supports by implementing the `syslog` numeric levels.

server/src/model/event/event.prisma
```diff
@@ -1,20 +1,17 @@
...
-  activeUserId String?  @map("active_user_id") @db.Uuid
-  eventTypeId  String   @map("event_type_id")
-  roleId       String   @map("role_id") @db.Uuid
-  activeRoleId String?  @map("active_role_id") @db.Uuid
-  logLevelId   String   @map("log_level_id")
-  route        String
-  createdAt    DateTime @default(now()) @map("created_at") @db.Timestamptz()
-  eventData    Json?    @map("event_data") @db.Json
+  id         String   @id @default(uuid()) @db.Uuid
+  userId     String?   @map("user_id") @db.Uuid
```
TARehman, line 5: Nit: we should run `prisma format` and it will do things like ensure these align correctly.

server/src/model/event/event.prisma
```diff
@@ -1,20 +1,17 @@
...
+  withRoleId String?   @map("with_role_id") @db.Uuid
+  eventType  String   @map("event_type")
+  logLevel   String   @map("log_level")
+  route      String
+  createdAt  DateTime @default(now()) @map("created_at") @db.Timestamptz()
+  eventData  Json?    @map("event_data") @db.Json
```
TARehman, line 11: Should we consider removing the nullability of this now that the `userId` and `roleId` values are going to appended to every payload?

server/src/model/event/event.prisma
```diff
@@ -1,20 +1,17 @@
...
-  roleId       String   @map("role_id") @db.Uuid
-  activeRoleId String?  @map("active_role_id") @db.Uuid
-  logLevelId   String   @map("log_level_id")
-  route        String
-  createdAt    DateTime @default(now()) @map("created_at") @db.Timestamptz()
-  eventData    Json?    @map("event_data") @db.Json
+  id         String   @id @default(uuid()) @db.Uuid
+  userId     String?   @map("user_id") @db.Uuid
+  withRoleId String?   @map("with_role_id") @db.Uuid
+  eventType  String   @map("event_type")
```
TARehman, line 7: Can we still name this column `eventTypeId`? I know that it doesn't formally link to another table anymore, but the implicit relation still exists in that we'll probably have the "reportable event types" or what not, and I would like to make it clear that this is a potentially joinable field otherwise (similar as with `withRoleId` and `logLevelId` depending).

server/src/model/event/eventSchema.ts
```diff
@@ -9,52 +9,59 @@ export const eventSchema = gql`
...
 `;
 
 type DateTime = Date;
 export interface Event {
   id: string;
-  user?: User; // TODO: I think this isn't right but eslint doesn't handle User | null at the moment
-  eventTypeId: string;
-  role?: Role; // TODO: I think this isn't right but eslint doesn't handle User | null at the moment
-  logLevelId: string;
+  user?: User;
```
TARehman, line 48: So, I think we should leave the TODO here unless we've fixed this. Strictly speaking, `user?: User;` isn't quite as accurate to the GQL type as `User | null`, but the `eslint` additions we have don't play nice with that right now (I think it's fixable, just not currently supported). We don't have to fix it for this but I would like to make sure we leave the annotation or otherwise get the work on the board to fix this later.

server/src/model/migrations/20250721204840_event_updates/migration.sql
```diff
@@ -0,0 +1,41 @@
...
+
+-- AddForeignKey
+ALTER TABLE "event" ADD CONSTRAINT "event_with_role_id_fkey" FOREIGN KEY ("with_role_id") REFERENCES "role"("id") ON DELETE RESTRICT ON UPDATE CASCADE;
```
TARehman, line 38: The foreign keys didn't get produced correctly by Prisma. These should be `ON DELETE SET NULL`. Not sure why they are not coming out correctly; they generate correctly on my end when I do it (I just checked and got a small diff when I ran it again).

My approach is always:
 - delete the most recent migration that I'm trying to replace
 - `prisma migrate reset --force` which brings the database and schema back into alignment with the migration right before the one being replaced
 - `prisma migrate dev --name temporary` which generates a new version of the migration that I am trying to recreate
 - Use git to restore the deleted original version, and move the migration.sql file from the `temporary` migration back to the original one. Then, review the Git diff to make sure that the old file changed in expected ways.

server/src/seeder.ts
```diff
@@ -13,46 +29,41 @@ export function clearDatabase() {
...
-    prisma().userRole.deleteMany(),
-    prisma().userState.deleteMany(),
-    prisma().userStateDemonstration.deleteMany(),
-    prisma().demonstration.deleteMany(),
-    prisma().demonstrationStatus.deleteMany(),
-    prisma().permission.deleteMany(),
     prisma().role.deleteMany(),
     prisma().state.deleteMany(),
     prisma().user.deleteMany(),
+
```
TARehman, line 35: This delete order really shouldn't work, given that you are deleting users before you delete the tables that users are in. I don't understand how it could have been working; the only context where it would have worked would have been in the `npm run seed`, since that resets the DB to the most up-to-date migration first.

TARehman (dismissed):
Left my feedback strictly on the back-end / DB stuff, not on the front-end parts. A few things needing tweaks, etc.

server/src/seeder.ts
tresdonicf: Heard! I don't usually try to get ahead of the git auto-merge algorithm and I just typically try to see what conflicts arise and resolve those then. I find it's actually pretty good and doesn't result in as many conflicts as folks sometimes expect! I didn't change lots of lines in that file so I think it'll be maybe two conflicts, the const and if our delete re-ordering changes are slightly different (I would just accept yours) but I don't like to get ahead of myself guessing.

I do think it's important to optimize for avoiding conflicts through using more, smaller files and encapsulating logic in functions!

server/src/model/event/event.prisma
tresdonicf: I prefer `eventType` since it's just a string on the table. The way I see it using it a table like "reportable event types" is another way we can use this field but not something we should design around.

You can always potentially join on a field if you make a field in another table match up but it's more confusing to see this table with `id` in it because that does suggest it always references an ID

client/src/hooks/event/eventTypes.ts
tresdonicf: Consolidating this to the other `syslog` thread

server/src/model/migrations/20250721204840_event_updates/migration.sql
tresdonicf: I think we'll only ever be logging out 3 levels in this system and we shouldn't be trying to implement syslog here.

INFO and ERROR are already implemented, maybe we'll do WARN for something like a successful deletion) but I think it's overkill to build a "comparable" for these 3 values. If we ever did need to do that, I think the more principled way would be to create it as a view rather than a prisma model. These can be queried just like a table but without all of the overhead.

```
CREATE VIEW log_level_syslog_view AS
SELECT *
FROM (
  VALUES
    ('INFO', 6),
    ('ERROR', 3),
    ('WARN', 4)
) AS log_levels(log_level, syslog_value);
```

But I'll probably stick with something simple like `WHERE log_level IN ('INFO', 'ERROR');`

Screenshot 2025-07-23 102712

server/src/model/event/event.prisma
tresdonicf: I don't think it's necessary unless it gains us something as we may end up changing this decision.

server/src/model/event/eventSchema.ts
tresdonicf: You're right that `User?` is technically `User | undefined` when it comes to typescript and not `User | null` but GQL doesn't have an `undefined` type so the true distinction isn't quite the same.

In JS, `null` is sort of considered something that a developer might intentionally set a value to in order to signal its absence whereas `undefined` is more like "this value doesn't exist but probably should / will". Some people say that having this distinction at all is confusing for this reason and I kinda agree, they're both `Nullable` so they match the nullability of the GQL field.

Since this is a queried type and not an object we're manipulating in our program - it's quite alright  for it to be `undefined` since we're not planning to set it back to `null` in our code. Hopefully that makes sense.

server/src/model/migrations/20250721204840_event_updates/migration.sql
tresdonicf: Good catch! I'll regenerate these

server/src/model/event/eventSchema.ts
TARehman: I'm confused why we wouldn't just fix this, though. The issue isn't with using `User | null` - as noted, that is arguably the most accurate representation of the GQL type in TS. The reason it is this way is that the extra stuff in `validate-graphql-typescript-match.js` breaks when given something like `User | null`.

Obviously doesn't block here, but I'm confused why we wouldn't fix the rules to properly parse the correct notation. I'll resolve this and maybe we can throw a conversation into an office hours soon.

server/src/model/event/event.prisma
TARehman: Well, it gains us that you can't insert an event that has a null eventData payload, and then subsequently lose information because we delete a role and the SET NULL constraints affect the actual columns. _That being said_, the payload not being null doesn't actually mean it will have the userId and roleId.

It's more that we know that there should never be an empty eventData payload in the new implementation, so if there ever is one, it should fail because something is broken. If we later want to change that, it's a lot easier to take away a NOT NULL constraint than it is to add one later.

server/src/model/migrations/20250721204840_event_updates/migration.sql
tresdonicf: This is updated with `ON DELETE SET NULL`

server/src/model/event/event.prisma
tresdonicf: `eventType`s are always in the system because they don't or never have referenced an ID from another table

`userId` and `roleId` explicitly reference a field which was an ID in another table.

server/src/seeder.ts
tresdonicf, line 1: In this file just removed the eventType and event seeding as eventType is no longer a table and events don't really need to be seeded since it's an audit log concept rather than something that should exist in order to move around in the application

server/src/model/event/eventSchema.ts
```diff
@@ -9,52 +9,59 @@ export const eventSchema = gql`
   type Event {
     id: ID!               # From B/E (generated by Prisma)
     user: User            # From B/E
-    eventTypeId: String!  # From F/E
-    role: Role            # From B/E
-    logLevelId: String!   # From F/E
+    eventType: String!    # From F/E
```
connor-parke-icf, line 12: these "From F/E" comments here, were they each generated by prisma? Might be worth removing those comments unless theyre critical. It kinda gives the impression that this is more transient than it may be, or that this is a file that regularly gets generated.

connor-parke-icf (approved):
Dont really have much to highlight here, just left a few comments as discussion points but nothing critical or blocking. Looks good!

server/src/model/event/eventResolvers.ts
```diff
@@ -33,36 +33,29 @@ export const eventResolvers = {
...
+
+      try {
+        await prisma().event.create({
+          data: {
+            userId: userId,
+            eventType: eventType,
+            logLevel: logLevel,
+            withRoleId: roleId,
+            route: route,
+            eventData: eventData
```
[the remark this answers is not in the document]
TARehman: I don't know this for sure, but I think there's probably a need to have a logical name for the "physical column" and the "logical column" or whatever Prisma calls it - when these are actually enforced.

e.g. the distinction between `role` and `roleId` in the `User` Prisma model seems like it might matter. In some contexts, you are passing in the actual value, and in other contexts, you're passing in a `Role` object, and I think it would be hard to tell which is which if they were both named the same. Assuming you even _can_ name them the same.

In this context, none of the join tables are enforced, so there's nothing stopping us from omitting the `Id` suffix.

server/src/model/event/eventSchema.ts
```diff
@@ -9,52 +9,59 @@ export const eventSchema = gql`
...
   """
-  input CreateEventInput {
-    eventTypeId: String!
-    logLevelId: String!
+  input LogEventInput {
+    eventType: String!
+    logLevel: String!
     route: String!
-    eventData: JSONObject!
+    eventData: JSONObject
```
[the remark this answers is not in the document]
TARehman: If the JSON object is null, something else broke in the code, because the code explicitly attaches the `userId` and `roleId` when you use the `logEvent` process.

james-becwar-cms (changes requested):
Please ping me once you update to eventData to not be null and I will approve and lets get this merged.

server/src/model/migrations/20250721204840_event_updates/migration.sql
james-becwar-cms: I liked the syslog idea for sorting/reporting and I like standards.  I also though we had agreed to use the external table for severity level, but it wasn't written down so that's on me.  

I also see Tresdon's point that its a enum on the client and its the only one creating logs right now.  So I'm fine leaving it, at worst it will be a 1 point story to refactor in the future.

server/src/model/event/eventSchema.ts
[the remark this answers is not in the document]
james-becwar-cms: The eventData is going to hold the userId and roleId and will be needed if the user is removed form the system, so every object will need at least {userIdD, roleId}

server/src/seeder.ts
TARehman: Nevermind, I convinced myself 😄

server/src/model/event/event.prisma
tresdonicf: Updated `eventData` to be nonNull

PR #84, opened 2025-07-15
DEMOS 267: Add Dummy Document Table | Implement Filtering

JeffreyM-GA:
Adds Documents table to Demo Details page with filtering capability
Uses dummy data to populate documents table


client/src/mock-data/demonstrationMocks.ts
```diff
@@ -69,4 +70,34 @@ export const demonstrationMocks: MockedResponse[] = [
...
+          ...testDemonstration,
+          name: "Updated Demo Name",
+          description: "Updated description",
+          evaluationPeriodStartDate: new Date("2024-07-01T00:00:00.000Z"),
+          evaluationPeriodEndDate: new Date("2024-07-31T00:00:00.000Z"),
+          updatedAt: new Date("2024-07-01T00:00:00.000Z"),
+        },
+      },
+    },
+  },
```
dustbuster, line 102: I have a much better demo mocks coming thru the pike with update field names

client/src/queries/demonstrationQueries.ts
```diff
@@ -80,3 +80,30 @@ export const ADD_DEMONSTRATION_QUERY = gql`
...
+      }
+      state {
+        id
+        stateName
+        stateCode
+      }
+      users {
+        id
+        fullName
+      }
```
dustbuster, line 106: Need to add ProjectOfficer to this i think. maybe? Not sure about update.

JeffreyM-GA: Hm this is strange, I dont think this changes should be showing in this PR. I branched from my previous feature branch to avoid conflicts, I expected this to go away after merging it but it looks like it didnt

PR #85, opened 2025-07-15
Demos-363 Adjusted table columns, added project officer, made seeder better, and corrected unit tests. 

dustbuster:
So as soon as i put project officer in the table, it sploded, because the Frontend was not prepared to deal with project officer. So i had to add that, and while i was at it. I made the dmeo mocks better and used better data for the seeder so we could get real data back and not 15 states. 

Mocks that match the api

Screenshot 2025-07-15 at 11 06 53

Screenshot 2025-07-15 at 14 48 48


client/src/components/table/columns/DemonstrationColumns.tsx
```diff
@@ -7,24 +7,10 @@ import {
...
-  evalPeriodStartDate: string;
-  evalPeriodEndDate: string;
-  demonstrationStatusId: number;
-  stateId: string;
-  projectOfficer: string;
-  demoNumber: string;
-  createdAt: string;
-  updatedAt: string;
-};
-
```
dustbuster, line 26: All types come from the pages/Demonstration.tsx

client/src/components/table/tables/DemonstrationTable.tsx
```diff
@@ -1,25 +1,13 @@
...
-  PaginationControls,
-} from "components/table/pagination/PaginationControls";
-import {
-  DemoWithSubRows,
-  groupByDemoNumber,
-} from "components/table/preproccessors/GroupByDemoNumber";
+import { RawDemonstration } from "../../../pages/Demonstrations";
+import { DemonstrationColumns } from "components/table/columns/DemonstrationColumns";
+import { ColumnFilterByDropdown } from "components/table/filters/ColumnFilterSelect";
+import { PaginationControls } from "components/table/pagination/PaginationControls";
```
dustbuster, line 5: No more red lines!

client/src/mock-data/demonstrationMocks.ts
```diff
@@ -1,102 +1,102 @@
...
-import { johnDoe } from "./userMocks";
-import {
-  ADD_DEMONSTRATION_QUERY,
-  GET_ALL_DEMONSTRATIONS_QUERY,
-  GET_DEMONSTRATION_BY_ID_QUERY,
-  UPDATE_DEMONSTRATION_MUTATION,
-} from "queries/demonstrationQueries";
+import rawDemoData from "../faker_data/demonstrations.json";
+import { states } from "../data/StatesAndTerritories";
+import { demonstrationStatuses } from "./demonstrationStatusMocks";
```
dustbuster, line 3: The mock imports the types from the actual page.

client/src/mock-data/demonstrationMocks.ts
```diff
@@ -1,102 +1,102 @@
...
+  name?: string;
+  description?: string;
+  evaluationPeriodStartDate?: string;
+  evaluationPeriodEndDate?: string;
+  createdAt?: string;
+  updatedAt?: string;
+  demonstrationStatusId?: string | number;
+  stateId?: string;
+  projectOfficerUser?: string;
+}
```
dustbuster, line 24: WE are using the demo.json file as a data seed. So i did have to make this type. But generally the outputting to RawDemonstration.

client/src/mock-data/demonstrationStatusMocks.ts
```diff
@@ -1,9 +1,36 @@
...
+    updatedAt: new Date("2025-01-03"),
+    demonstrations: [],
+  },
+  {
+    id: "12345678-1234-1234-1234-000000000004",
+    name: "Withdrawn",
+    description: "Demonstration was withdrawn before completion.",
+    createdAt: new Date("2025-01-04"),
+    updatedAt: new Date("2025-01-04"),
+    demonstrations: [],
```
dustbuster, line 34: Um. We should really make a seeder. And set up GUIDs so we can at least make our mocks a bit more matchy.

client/src/pages/Demonstrations.tsx
```diff
@@ -1,38 +1,116 @@
...
+  projectOfficerUser: ProjectOfficerUser | null;
+  state: StatesAndTerritories | null;
+  demonstrationStatus: DemoStatus | null;
+}
+
+export interface DemonstrationsQueryData {
+    demonstrations: DemoFromQuery[];
+  }
+
+export const DEMONSTRATIONS_TABLE = gql`
```
dustbuster, line 59: Exported to Mock for consistency.

server/src/SWUsersSeeder.json
```diff
@@ -0,0 +1,2702 @@
+[
+    {
+        "id": 1,
+        "name": "Admiral Gial Ackbar",
+        "email": "a.ackbar@rebellion.org",
+        "username": "gialackbar",
+        "role": "Rebel Admiral",
+        "status": "pending",
+        "created_at": "2024-11-11 13:27:28"
+    },
```
dustbuster, line 10: I moved this over to server because i needed a seed file for the um seeder. 
And @connor-parke-icf wanted some star wars chars. so who am i to argue?

client/src/components/table/columns/DemonstrationColumns.tsx
tresdonicf: I think we should be able to consolidate these types with what's used in the server by doing `import {Demonstration} from demos-server`. This way we have only one type that is agreed on by all the components (client, server, DB) - let me know what you think about this.

client/src/data/StatesAndTerritories.ts
```diff
@@ -1,68 +1,68 @@
-// client/src/faker_data/usStates.ts
+// client/src/data/StatesAndTerritories.ts
```
tresdonicf, line 1: nit: Don't think we need the comment for the filename since it can become outdated

client/src/data/StatesAndTerritories.ts
```diff
@@ -1,68 +1,68 @@
...
 
 export interface USState {
   name: string;
-  abbrev: string;
+  abbreviation: string;
```
tresdonicf, line 5: nice change!

client/src/hooks/useDemonstration.test.ts
```diff
@@ -4,7 +4,7 @@ import { AddDemonstrationInput } from "demos-server";
 import {
   mockAddDemonstrationInput,
   testDemonstration,
-} from "mock-data/demonstrationMocks";
+} from "../mock-data/demonstrationMocks";
```
tresdonicf, line 7: should be able to just import from `mock-data` (provided in `tsconfig.json - paths`). The reason to do this is to only have to declare the path relative to the `src` root in one place and reduce changes related to moving files around.

At some point should probably make a linting rule to disallow relative imports that start with `../`

client/src/mock-data/demonstrationMocks.ts
tresdonicf: FWIW I think it's more straightforward to make source code JSON a `.ts` or `.js` file and we can just put a `const varName: OptionalTypeHint = {the existing json file contents}` and get a file that uses our existing types.


JSON is still needed in the case of things like requests and stuff where you don't know the shape beforehand but if it's in source code we do know the shape of the data and we can type it directly in the file

client/src/pages/Demonstrations.tsx
```diff
@@ -1,38 +1,116 @@
...
+  name: string;
+  description: string;
+  evaluationPeriodStartDate: string;
+  evaluationPeriodEndDate: string;
+  createdAt: string;
+  updatedAt: string;
+  projectOfficerUser: ProjectOfficerUser | null;
+  state: StatesAndTerritories | null;
+  demonstrationStatus: DemoStatus | null;
+}
```
tresdonicf, line 53: I'd encourage the use of the server `Demonstration` type - maybe with some typescript utility types like `Pick` or `Partial`. I know it can be kind of a pain but it's less drift than if we keep redeclaring many fields of the type

client/src/pages/Demonstrations.tsx
tresdonicf: Could we move this to the `queries` folder by chance? The idea being that we can then consolidate all of our GQL queries in one place and eventually move to using [Apollo Fragments](https://www.apollographql.com/docs/react/data/fragments).

Also that way the export will make more sense and things like the mocker won't need to pull directly from the component but rather both the mock query and the component query will pull the same value from `queries`

client/src/mock-data/demonstrationStatusMocks.ts
```diff
@@ -1,9 +1,36 @@
 import { DemonstrationStatus } from "demos-server";
 
-export const activeDemonstrationStatus: DemonstrationStatus = {
-  id: "1",
-  name: "Active",
-  createdAt: new Date("2025-01-01"),
-  updatedAt: new Date("2025-01-01"),
-  description: "Active Demonstration Status",
-};
```
tresdonicf, line 9: I think it's best to keep this around, sometimes we'll want to use a demonstration status from this file for another object and it's quite a bit less harder to interpret comparisons to `demonstrationStatuses[0]` .

If anything I would specify all of the values as `const` and then have an array that just compiles them:
```
export const demonstrationStatus: DemonstrationStatus = [
activeDemonstrationStatus,
inactiveDemonstrationStatus,
onHold,
...
]
```

client/src/mock-data/demonstrationStatusMocks.ts
tresdonicf: This is what it looks like to seed data on the client side and the value-gain from the keys matching up here is not nearly what's gained by doing it in the DB. Here it's just for getting enough records to make sure the component behavior is alright - on the DB those constraints are actually checked so it's different and from my testing that's where we should focus our efforts on matching up keys. Not to mention if we have a good DB seeder script we can just use that instead of this mock response.

client/src/pages/Demonstrations.tsx
```diff
@@ -1,38 +1,116 @@
...
+        id
+        name
+      }
+    }
+  }
+`;
+
 export const Demonstrations: React.FC = () => {
-  // Dummy value - replace me!
   const currentUserId = 123;
```
tresdonicf, line 88: I think we should keep this comment around with a TODO to remove this or just remove it entirely

tresdonicf (commented):
Nice work on this! I'd like to see if we can centralize more of the types to be shared with those that are exported from the `server` folder as these have the linting set up to make sure they're the same as the GQL types.

PR #87, opened 2025-07-15
task/tweak-event-table-implementation

TARehman:
UPDATED: See the additional context for the most recent changes in [this comment](https://github.com/Enterprise-CMCS/demos/pull/87#issuecomment-3090199297).

I was asking @tresdonicf a bunch of questions, and eventually decided it might be easier to just raise a PR and discuss it here. This PR does a bunch of _mostly_ housekeeping items, including the below.

 - Adds an ERD that reflects the _current state_ of the data model, rather than a proposed version. I will eventually clean up the other one to be in harmony with this, or figure out a better way of representing what is implemented vs. what is future state.
 - For a number of Prisma schemas, I formatted them consistent with the way @connor-parke-icf and I started to do them: all the columns at the top, then all the constraints, then any unique or mapping statements at the end. I ran `prisma format` after this so indentation changed on a few of them. This affects:
   - rolePermission
   - userRole
   - userState
   - state
   - user
   - role
 - Moved the eventType model out of the event folder and into its own folder. This is consistent with the way that @connor-parke-icf and I initially had figured on doing this - one table to a folder. If there's a reason to do it another way, I'm open to hearing it - that being said, I think there's good reasons to go with the "one table per folder" approach. In doing this, I moved over the resolvers, and redid the index.ts files accordingly.
 - I don't know how it worked before, but the with_role_id wasn't coming up as a UUID when I generated because the `@db.Uuid` tag wasn't there. 🤔
 - I added a history table for the event types as well as an updated timestamp. I agree with the reasoning of not having these things on the event table itself, since that is a strictly append-only table. However, I could imagine we would want to see changes / updates to the types over time, and these things facilitate that.
 - I moved the definition of the logging level ENUM into the schema.prisma file, where we also put the revision type one, and named it with underscores and lowercase to avoid needing to quote it continuously within the database.
 - The migrations are odd; I had to truncate the existing table before I could run the migration correctly. This could be cleaner but again, as we get closer to a final model, I foresee us going back and just smooshing all the revisions together.


server/src/model/event/eventType.prisma
tresdonicf, line 1: nit: Since event types are a part of the "event system" and not a top-level entity of its own I think it's best to leave them in `event`. I think that this separate folder undermines some of the ontology of how these should be used without any clear gain. Not a deal (hence the `nit:`) but just wanted to share that as adding a folder for every entity seems to move back toward a non-nested structure

server/src/model/eventType/eventTypeHistory.prisma
```diff
@@ -0,0 +1,12 @@
...
+  revisionType RevisionType @map("revision_type")
+  modifiedAt   DateTime     @default(now()) @map("modified_at") @db.Timestamptz()
+  id           String
+  description  String
+  logLevel     LogLevel     @map("log_level")
+  createdAt    DateTime     @map("created_at") @db.Timestamptz()
+  updatedAt    DateTime     @map("updated_at") @db.Timestamptz()
+
+  @@map("event_type_history")
+}
```
tresdonicf, line 12: I don't think this is quite necessary.

My understanding is that the plan for this table is that it will only be changed via migrations and that should give us a history of these changes through `git`. I can see some instances of migration consolidation that blow this history away and leave us without it but I think the answer for all of these event type changes will be: "We identified a new type of event we want to log" or "We realized we don't need to log a certain type of event".

I sort of think these history tables should be more so for entities that the users can change via the UI

server/src/model/migrations/20250715182703_empty_event_type/migration.sql
```diff
@@ -0,0 +1,3 @@
+-- Migration to empty the event_type table
+-- Allows the changes in the next migration to succeed
+TRUNCATE TABLE "event_type" CASCADE;
```
tresdonicf, line 3: Can this not run as the first command of the `..._update_event_implementation` migration?

Just thinking about this pattern as we continue to migrate this table.

server/src/model/migrations/20250715183022_update_event_implementation/migration.sql
```diff
@@ -0,0 +1,110 @@
...
+    "created_at" TIMESTAMPTZ NOT NULL,
+    "updated_at" TIMESTAMPTZ NOT NULL,
+
+    CONSTRAINT "event_type_history_pkey" PRIMARY KEY ("revision_id")
+);
+
+-- AddForeignKey
+ALTER TABLE "event" ADD CONSTRAINT "event_with_role_id_fkey" FOREIGN KEY ("with_role_id") REFERENCES "role"("id") ON DELETE RESTRICT ON UPDATE CASCADE;
+
+CREATE OR REPLACE FUNCTION demos_app.log_changes_event_type()
```
tresdonicf, line 40: Awesome capturing the changes and inserting the old values here!

tresdonicf (commented):
Thanks for making these changes! I'm working on some similar modifications to this system as a part of #83 .

The more I think about this the more I think `EventType` living in the DB is a bit heavy-handed. The goal is that we want to log a `string` of the type of event it is and another `string` for the level of event it is and these events will (by definition) always come from the UI.

It seems like the UI code should actually declare what type of events it is able to log and what the relevant `logLevel` and `description` should be for these and we reconstruct this mapping when needed

Another way to say this is that it's in a weird place because it's a reference table but one that changes more than "state" or something and making those changes might become a bit of unneeded toil.

server/src/model/user/user.prisma
```diff
@@ -1,12 +1,13 @@
...
-  updatedAt                    DateTime                 @updatedAt @map("updated_at") @db.Timestamptz()
+  id             String   @id @default(uuid()) @db.Uuid
+  cognitoSubject String   @unique @map("cognito_subject") @db.Uuid
+  username       String
+  email          String
+  fullName       String   @map("full_name")
+  displayName    String   @map("display_name")
+  createdAt      DateTime @default(now()) @map("created_at") @db.Timestamptz()
+  updatedAt      DateTime @updatedAt @map("updated_at") @db.Timestamptz()
+
```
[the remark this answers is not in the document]
TARehman: It does add the relations if they are missing on the other end, which is nice. But mostly in this case, because I broke the various sections apart, it just changed the indentation as a result. We have been breaking them into separate chunks because it helps to understand section 1 as columns, section 2 as relations, etc.

server/src/model/event/eventType.prisma
TARehman: This is worth a team discussion as to what the models represent. I had taken the stance that they are tables, as had @connor-parke-icf originally. If they are intended to represent something higher level, I can definitely see the argument. This is a place where the ORM is having to do two different kinds of lifting for us - probably an example of the [ORM impedance mismatch](https://en.wikipedia.org/wiki/Object%E2%80%93relational_impedance_mismatch).

server/src/model/migrations/20250715183022_update_event_implementation/migration.sql
TARehman: There's a python script found in `data/utilities/scripts` which generates the history triggers easily on demand, which makes it a lot easier to add them in. It does, however, depend on each table being its own folder (re: your comment about nesting model structures).

TARehman:
> The more I think about this the more I think `EventType` living in the DB is a bit heavy-handed. The goal is that we want to log a `string` of the type of event it is and another `string` for the level of event it is and these events will (by definition) always come from the UI.
> 
> It seems like the UI code should actually declare what type of events it is able to log and what the relevant `logLevel` and `description` should be for these and we reconstruct this mapping when needed
> 
> Another way to say this is that it's in a weird place because it's a reference table but one that changes more than "state" or something and making those changes might become a bit of unneeded toil.

This is definitely valid and worth a discussion. I think this touches on two things we've discussed previously.

 - We should not recreate a logging system in the event system - we should use the logging systems in place for logging. To that regard, the event system has some requirements that are different from a logging system. I completely agree that a logging system should just accept whatever messages come from the application, because telemetry is their entire purpose. But the event system is supposed to fulfill a customer requirement for reporting and analysis of user activity. So, it makes more sense (to me at least) that it should be more discerning about what messages it accepts, precisely because its purpose is _not_ to accept anything that is sent, but rather, to capture specific, requirements-driven information from the application that is surfaced later to the business users.
 - To that end, having the types defined in advance is also helpful for reporting, especially if we need to report zero-count events (this is much easier if there's a list of canonical "things that should happen" we can join against).

@james-becwar-cms, probably more stuff to discuss in office hours here.

server/src/model/eventType/eventTypeHistory.prisma
dustbuster: I would also prefer a migration. Because then when there are breaking changes i can easily track them down.

server/src/model/migrations/20250715182703_empty_event_type/migration.sql
dustbuster: yea i think this could be dangerous if we forget about it....

server/src/model/migrations/20250715183022_update_event_implementation/migration.sql
```diff
@@ -0,0 +1,110 @@
...
+            'D'::demos_app.revision_type_enum,
+            OLD.id,
+            OLD.description,
+            OLD.log_level,
+            OLD.created_at,
+            OLD.updated_at
+        );
+        RETURN OLD;
+    END IF;
+    RETURN NULL;
```
dustbuster, line 83: I done would have wrote a script for this nicely done! lol

server/src/model/role/role.prisma
```diff
@@ -1,9 +1,10 @@
...
-  name            String
-  description     String
-  createdAt       DateTime         @default(now()) @map("created_at") @db.Timestamptz()
-  updatedAt       DateTime         @updatedAt @map("updated_at") @db.Timestamptz()
+  id          String   @id @default(uuid()) @db.Uuid
+  name        String
+  description String
+  createdAt   DateTime @default(now()) @map("created_at") @db.Timestamptz()
+  updatedAt   DateTime @updatedAt @map("updated_at") @db.Timestamptz()
+
```
dustbuster, line 7: Is there any changes here? or just white space?

server/src/model/schema.prisma
```diff
@@ -16,3 +16,12 @@ enum RevisionType {
 
   @@map("revision_type_enum")
 }
+
+enum LogLevel {
+  DEBUG
+  INFO
+  WARN
+  ERROR
+
```
dustbuster, line 25: Laravel also had an EMERGENCY level. This is for when the logging is broked and it needs to write a darn error like now! Could be overkill. Just i thing to think about

server/src/model/role/role.prisma
TARehman: A lot of that is just whitespace when we reformatted; there shouldn't be any meaningful changes.

server/src/model/user/userSchema.ts
```diff
@@ -42,7 +44,7 @@ export const userSchema = gql`
 
   type Mutation {
     addUser(input: AddUserInput!): User
-    updateUser(input: UpdateUserInput!): User
+    updateUser(id: ID!, input: UpdateUserInput!): User
```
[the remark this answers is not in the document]
TARehman: I don't know if there's any clear preference one way or another either. In other places, we put it in as an argument separately from the input, but all the stuff I've seen online says you could also put it in the input and it would be just as valid.

TARehman:
Okay, so based on a talk with @tresdonicf he pointed out the possibility of deleting a user or role, and somehow the previous Prisma migration produced what I think was an impossible scenario: a column that could not be NULL linked to a Prisma field that was optional.

Spotting that in my final review made me realize this would be an issue, so I've updated the `event` table to capture both the "active" link between a user or role and an event, and the value at the point in time when the event is created. If a role or user is deleted, the values in `active_user_id` and `active_role_id` get set to `NULL` in the `event` table, but the original values remain in `user_id` and `role_id`. `user_id` and `role_id` cannot be NULL, but are not constrained to another table, meaning we can safely delete a user or role without losing that information. I updated the API to handle this during event creation.

I think this is probably the simplest compromise for this type of append-only logging table. Hopefully it makes sense.

PR #88, opened 2025-07-16
Demos-533: custom authorizer

cms-jesse:
This change adds a custom authorizer to the API gateway that validates the existence of "roles" prior to allowing a user through. It also sets the request context values that are passed to the backend.

Due to the existing dynamic outputs and inputs, some manual steps will be required to deploy this change. However, after the initial manual change, future deploys will proceed without issues.

### DIFF
--- a/client/src/router/cognitoConfig.ts
+++ b/client/src/router/cognitoConfig.ts
@@ -46,8 +46,8 @@ const PRODUCTION_COGNITO_CONFIG: CognitoConfig = {
   authority: import.meta.env.VITE_COGNITO_AUTHORITY!,
   domain: import.meta.env.VITE_COGNITO_DOMAIN!,
   client_id: import.meta.env.VITE_COGNITO_CLIENT_ID!,
-  post_logout_redirect_uri: import.meta.env.BASE_URL,
-  redirect_uri: import.meta.env.BASE_URL,
+  post_logout_redirect_uri: import.meta.env.BASE_URL == "/" ? `${window.location.origin}/` : import.meta.env.BASE_URL,
+  redirect_uri: import.meta.env.BASE_URL == "/" ? `${window.location.origin}/` : import.meta.env.BASE_URL,
   scope: "openid email profile",
 };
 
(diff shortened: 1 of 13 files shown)


client/src/router/cognitoConfig.ts
```diff
@@ -46,8 +46,8 @@ const PRODUCTION_COGNITO_CONFIG: CognitoConfig = {
   authority: import.meta.env.VITE_COGNITO_AUTHORITY!,
   domain: import.meta.env.VITE_COGNITO_DOMAIN!,
   client_id: import.meta.env.VITE_COGNITO_CLIENT_ID!,
-  post_logout_redirect_uri: import.meta.env.BASE_URL,
-  redirect_uri: import.meta.env.BASE_URL,
+  post_logout_redirect_uri: import.meta.env.BASE_URL == "/" ? `${window.location.origin}/` : import.meta.env.BASE_URL,
+  redirect_uri: import.meta.env.BASE_URL == "/" ? `${window.location.origin}/` : import.meta.env.BASE_URL,
```
cms-jesse, line 50: Potentially not ideal to use window.location.origin here, however its basically the only option at the moment until we can get a static url

deployment/config.ts
```diff
@@ -21,6 +22,10 @@ export const determineDeploymentConfig = async (
 ): Promise<DeploymentConfigProperties> => {
   const project = process.env.PROJECT || "demos";
 
+  execSync("npm ci", {
+    cwd: "../lambda_authorizer"
+  })
```
dustbuster, line 27: I can't say I am a fan of this. 
This can be used for command injection. It is JS and can be inspected perhaps?

Can you bundle this from build? 

I tossed this in AI and it is suggesting something like this:
```
import * as lambdaNode from 'aws-cdk-lib/aws-lambda-nodejs';

const lambdaAuthorizer = new lambdaNode.NodejsFunction(this, 'LambdaAuthorizer', {
  entry: '../lambda_authorizer/index.ts', // or handler file
  handler: 'handler',
  runtime: lambda.Runtime.NODEJS_18_X,
  bundling: {
    minify: true,
    externalModules: ['aws-sdk'], // AWS SDK is pre-included in Lambda
  }
});
```

cms-jesse: Let me look into this more in another ticket. I don't like the execSync call either

PR #97, opened 2025-07-24
DEMOS-554:  Update authorizer to use CDK bundler

cms-jesse:
Rather than using execSync call during the CDK deploy, use the built-in CDK bundling functionality


dustbuster:
> Rather than using execSync call during the CDK deploy, use the built-in CDK bundling functionality

Nice update!

dustbuster (approved):
Looks good to me! Thanks for addressing that thing!

PR #98, opened 2025-07-24
Fix Error: Query.event defined in resolvers, but not in schema

tresdonicf:
I removed these unused query types and the GQL server was yelling at me because the resolvers still exist


TARehman (approved):
LGTM 🥳

PR #100, opened 2025-07-24
DEMOS-508-natural-key-replacement

TARehman:
Refactors a number of tables to use natural / human-readable keys, rather than UUIDs. I tried to carry these changes up into the client where encountered to avoid being too disruptive - currently, the unit tests pass successfully and the `dev:mocks` version appears to run as expected.

This also fixes the `evaluationPeriodStartDate` and `evaluationPeriodEndDate` fields to be `effectiveDate` and `expirationDate` instead.

Some notes:
 - I removed history tables from a number of places based on some of our conversations.
 - The state table is refactored so that the ID column is just the state code, and the stateName is just the name (consistent with other tables where we don't repeat the meaning of the thing twice).
 - I've tried to consistently make all the parameter / constraint tables have ID, name, and description, even in contexts where the ID and name may be somewhat similar (e.g. 'APPLICATION' and 'Application').
 - The seeder is updated and successfully produces data for the updated structure, and the API appears to work without issues during my experimentation.
 - I tried to bypass `prettier` changes when doing things on the client-side, simply because I thought it would make it much harder to see the substantive changes. `prettier` ran as normal for server-side stuff.
 - `npm run build` works fine in the `client` but I did notice that `tsc` gives some type errors. That being said, they don't seem to be different in my branch vs. main, except insofar as variable names have changed on types that are already not lined up with the backend. I don't think I added / caused any of that.


james-becwar-cms (approved):

server/src/model/documentType/documentTypeResolver.ts
```diff
@@ -11,50 +14,52 @@ export const documentTypeResolvers = {
...
+          name: input.name,
+          description: input.description,
+        },
       });
     },
 
     updateDocumentType: async (
       _: undefined,
-      { id, input }: { id: string, input: UpdateDocumentTypeInput }
+      { id, input }: { id: string; input: UpdateDocumentTypeInput },
```
connor-parke-icf, line 36: huh, was this breaking before? feel like this shouldnt ever have worked with the syntax error

server/src/model/role/roleResolvers.ts
```diff
@@ -64,12 +64,6 @@ export const roleResolvers = {
         },
       });
     },
-
-    deleteState: async (_: undefined, { id }: { id: string }) => {
-      return await prisma().state.delete({
-        where: { id: id },
-      });
-    },
```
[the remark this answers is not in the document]
connor-parke-icf: guessing it was supposed to be deleteRole, and we decided we didnt need an endpoint to delete a role, which i can understand

connor-parke-icf (approved):
no issues, couple of comments. Most of this is the same change applied to 50 files, but the change itself is fairly simple. Like the cleanup and removal of some mutators in the case of states.  and improvements on the Seeder. 

There is a built-in prisma seeding thing, but i dont know how much we want to lean into that; it could be really annoying or cumbersome. I would imagine we would like to use our graphql endpoint to seed the backend eventually, but continuing it in prisma directly is totally fine for now

client/src/components/modal/DemonstrationModal.test.tsx
```diff
@@ -206,8 +206,8 @@ describe("DemonstrationModal", () => {
       expect(mockTrigger).toHaveBeenCalledWith({
         name: "Test Demonstration",
         description: "",
-        evaluationPeriodStartDate: expect.any(Date),
-        evaluationPeriodEndDate: expect.any(Date),
+        effectiveDate: expect.any(Date),
+        expirationDate: expect.any(Date),
```
dustbuster, line 210: I like it.

client/src/mock-data/stateMocks.ts
```diff
@@ -1,7 +1,6 @@
 import { State } from "demos-server";
 
 export const california: State = {
-  id: "1",
-  stateCode: "CA",
-  stateName: "California",
+  id: "CA",
+  name: "California",
```
dustbuster, line 5: Much gooder.

server/src/model/documentType/documentTypeResolver.ts
TARehman: I think it might have just disregarded it? That looks like a `prettier` change, I don't remember making it.

server/src/model/event/eventResolvers.ts
```diff
@@ -28,27 +36,35 @@ export const eventResolvers = {
...
-    }
+    },
   },
 
   Event: {
     user: async (parent: Event) => {
-      return await prisma().user.findUnique({
-        where: { id: parent.userId },
-      });
+      if (!parent.userId) {
```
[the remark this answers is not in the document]
TARehman: No, but if you delete a user, the events it was attached to will have the column get nulled out. The original value is always there as part of the eventData payload, though.

server/src/model/state/stateResolvers.ts
```diff
@@ -14,50 +13,6 @@ export const stateResolvers = {
     },
   },
 
-  Mutation: {
```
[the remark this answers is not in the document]
TARehman: So in my view the seeder is entirely making fake data (also will discuss this below). When it comes time to load or migrate data into the DB for "real" we'll put it in migrations. States are a good example of that.

PR #101, opened 2025-07-25
WIP: DEMOS-268: Edit Document Details Modal

tresdonicf:
Edit Document Details Modal for DEMOS-268


tresdonicf:
Unit tests are failing on main as well. I'll look to put up a separate PR for that unless somebody is already on it

client/src/components/button/PrimaryButton.tsx
```diff
@@ -1,10 +1,8 @@
 import React from "react";
 
-import {
-  BaseButton,
-  ButtonSize,
-} from "./BaseButton";
+import { BaseButton, ButtonSize } from "./BaseButton";
 
+// TODO: Let's rename this to `ButtonProps` and export it for the other buttons
```
james-becwar-cms, line 5: Please create a backlog item for the button changes and remove the TODOs

client/src/components/modal/document/DocumentModal.tsx
```diff
@@ -0,0 +1,321 @@
...
+import { PrimaryButton, SecondaryButton } from "components/button";
+import { AutoCompleteSelect } from "components/input/select/AutoCompleteSelect";
+import { BaseModal } from "components/modal/BaseModal";
+import { useToast } from "components/toast";
+import { tw } from "tags/tw";
+import { TextInput } from "components/input";
+
+type DocumentModalType = "add" | "edit";
+
+// TODO: get these from the server or a shared constants file.
```
james-becwar-cms, line 17: Please have a backlog item created and remove the TODO

client/src/components/modal/document/DocumentModal.tsx
```diff
@@ -0,0 +1,321 @@
...
+  "application/vnd.ms-excel.sheet.macroEnabled.12",
+  "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
+  "application/zip",
+];
+
+const ACCEPTED_EXTENSIONS = ".pdf,.doc,.docx,.xls,.xlsx,.zip";
+
+const MAX_FILE_SIZE_MB = 600;
+const MAX_FILE_SIZE_BYTES = MAX_FILE_SIZE_MB * 1024 * 1024;
+const MAX_FILENAME_DISPLAY_LENGTH = 60;
```
james-becwar-cms, line 51: Should these get moved to the env.ts file?  Seems more like an app global setting.

client/src/components/modal/document/DocumentModal.tsx
```diff
@@ -0,0 +1,321 @@
...
+  onClose = () => {},
+  forDocumentId,
+}) => {
+  const { showSuccess } = useToast();
+
+  const documentModalType: DocumentModalType = forDocumentId ? "edit" : "add";
+  const modalTitle =
+    documentModalType === "edit" ? "Edit Document" : "Add New Document";
+
+  // TODO: get this from GQL hook
```
james-becwar-cms, line 233: Please create a backlog item and remove the TODO

client/src/components/modal/AddDocumentModal.tsx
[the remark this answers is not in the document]
james-becwar-cms: If you use the `git mv` command git will know about the move and the history and changes will show up correctly. You don't need to change anything in this PR, just a FYI.

client/src/config/env.ts
```diff
@@ -1,3 +1,4 @@
+// TODO: probably best to rename this to `environment.ts` to be more explicit
```
james-becwar-cms, line 1: I'm allergic to todo's.  env.ts is prob ok with the popularity of .env files.

client/src/layout/Tabs.tsx
```diff
@@ -28,7 +28,7 @@ export const Tabs: React.FC<TabsProps> = ({
               <button
                 onClick={() => onChange(tab.value)}
                 className={
-                  "inline-block px-2 py-2 font-medium " +
+                  "inline-block px-2 py-2 font-medium cursor-pointer " +
```
james-becwar-cms, line 31: Don't need to change anything, but I was thinking... Should put the cursor-pointer stuff higher up in the CSS so we don't have to apply it as much?

client/src/TODO.md
```diff
@@ -0,0 +1,9 @@
+A running list of TODO items that might not point at any part of the code in particular but that we want to capture
+
+1. Ensure no fake / mocked data makes it to prod
+   - For the `faker_data/` folder - remove this if seeding volume allows to test things like pagination. Otherwise move it to a place in our below the fold dev environment to test it.
+   - Add linting rule to make sure that components cannot import from `mock-data/`. Hook these up to the GQL queries to use that data in the proper way.
+
+2. Look into importing just `types.ts` from `server/`. Currently in the package.json we're installing it the same as the rest of our modules. This has the effect of bringing everything to the node_modules when we just need the one file. There's a couple ways to go about this:
+   - Make a "package" out of the `types.ts` folder.
+   - Make a file `client/server-types.ts` that just pulls in the types directly from the filepath (`../server/src/types.ts` or something) and re-exports them for use in the client, removing the dependency.
```
james-becwar-cms, line 9: Lets move this to Jira so it can be put in to sprints.

server/src/model/event/eventSchema.ts
```diff
@@ -58,7 +58,7 @@ export interface LogEventInput {
   eventType: string;
   logLevel: string;
   route: string;
-  eventData: object;
+  eventData?: object;
```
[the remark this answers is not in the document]
james-becwar-cms: eventData should be required since it includes the userid and roleid and its our last resort if the user/role is deleted.

We keep coming back to this, perhaps we should move the static userid and static roleid to there own fields?

nunesmlucas (approved):

james-becwar-cms (changes requested):
Code looks awesome.  Lets get the TODO's in to jira.

client/src/components/modal/document/DocumentModal.tsx
tresdonicf: I would consider moving this to `useFileDrop.ts` as it applies there but it's not really an environment consideration - For now I think it's okay here as we don't quite know where this will be shared yet (after all the DocumentModal makes sense as the place we'll be most concerned with this).

I'm confused by the choice to support `600MB` - but I didn't change that hahah

PR #104, opened 2025-07-25
DEMOS-561: add IDM configuration to CDK

cms-jesse:
This change adds the IDM configuration to the CDK code. It had been set up manually for initial testing


dustbuster (approved):
Looks good to me.

PR #107, opened 2025-07-28
Demos 518 new column for applications

connor-parke-icf:


client/src/components/table/columns/useDemonstrationColumns.tsx
```diff
@@ -0,0 +1,210 @@
...
+            type="button"
+            size="small"
+            onClick={handleClick}
+            className="px-2 py-0 text-sm font-medium"
+          >
+            View
+          </SecondaryButton>
+        );
+      },
+    }),
```
connor-parke-icf, line 187: modified to make sure the view button links to the correct tab on the demo details page.

client/src/pages/DemonstrationDetail.tsx
```diff
@@ -19,21 +15,49 @@ import { ExtensionTable } from "components/table/tables/ExtensionTable";
...
 import { mockAmendments } from "mock-data/amendmentMocks";
 import { mockExtensions } from "mock-data/extensionMocks";
-import { useParams } from "react-router-dom";
+import { useLocation, useParams } from "react-router-dom";
 
 import { isTestMode } from "../config/env";
 
 export const DemonstrationDetail = () => {
   const { id } = useParams<{ id: string }>();
+  const location = useLocation();
```
connor-parke-icf, line 27: sets the tab based on the query param, like 

/demonstrations/1?extension=true or /demonstrations/1?amendment=true

client/src/hooks/useDemonstration.ts
```diff
@@ -13,25 +16,55 @@ import {
...
 } from "queries/demonstrationQueries";
 
+type Demonstration = ServerDemonstration & {
+  extensions: {
+    name: string;
+    projectOfficer: { fullName: string };
+    extensionStatus: { name: string };
+  }[];
+};
+
```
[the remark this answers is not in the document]
tresdonicf: Awesome job with the union type! Maybe we should add a TODO here to just make sure it's revisted. At some point we'll be able to `ctrl + f` for all these `TODO`s and group them into themes and make some tickets

client/src/components/table/columns/useDemonstrationColumns.tsx
tresdonicf, line 1: I don't think the `hook` pattern is quite what we're looking for here - it's alright that it's named as one, but it's moreso just a function (which is still an improvement imo). In react terms, "hooks" are typically things that will be used by multiple components whereas this will be primarily used to clean up a component (still a great thing to use it for) - just wanted to mention this distinction

client/src/mock-data/demonstrationMocks.ts
```diff
@@ -116,6 +116,25 @@ export const demonstrationMocks: MockedResponse[] = [
             state: { name: "Montana" },
             projectOfficer: { fullName: "John Doe" },
             users: [{ id: "1" }],
+            amendments: [
```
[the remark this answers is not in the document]
tresdonicf: That's something I was wondering about as well - I'm working on a `TODO.md` file in #101 with an item like this in it to revisit this and make sure it's cleaned up.

I think we should declare a much slimmer group of mock data (1-3 entities of all the important concepts - demonstration, user, etc)  and when we're doing something like testing pagination, we just render that same record in a loop instead of declaring multiple records. We could do this in a new tab in our "Dev Mode" sidebar area called "Tables". That way we don't need to declare a bunch of these and we still get what we need out of testing pagination

client/src/pages/DemonstrationDetail.tsx
tresdonicf: When are these query params populated? Do we still need to do this work for if somebody was looking to `ctrl + c` their current address and share something at their current tab or does this already exist?

tresdonicf (approved):

connor-parke-icf:
Renamed useDemonstrationColumns component to not use the "hook" naming convention. There is likely more to do here to make it a bit smoother, but this should suffice for now. Added TODO comments to various places where we could revisit.

PR #114, opened 2025-07-30
DEMOS-577-gql-list-return-types

TARehman:
Very simple PR that just changes the query returns in the GraphQL schemas to use `[Thing!]!` instead of `[Thing]!`. This is a small tweak, but makes the results more consistent with the other types throughout the system. `[Thing]!` means that an array is always returned, but that the elements can either be `Thing` or `null`. `[Thing!]!` means that the array must have only non-null elements. Both allow for empty arrays, to be clear.

Unless there's an actual reason to expect / return null elements from a query (i.e. the null values communicate some meaning), everything I've seen indicates the `[Thing!]!` type is more appropriate in GQL.

### DIFF
--- a/server/src/model/demonstration/demonstrationSchema.ts
+++ b/server/src/model/demonstration/demonstrationSchema.ts
@@ -54,7 +54,7 @@ export const demonstrationSchema = gql`
   }
 
   type Query {
-    demonstrations: [Demonstration]!
+    demonstrations: [Demonstration!]!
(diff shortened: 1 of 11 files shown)


connor-parke-icf (approved):
i have a nagging feeling that there was a reason i had made many of these nullable during my first pass at the server, i cant think of that reason now. Im guessing it was a mistake, as it seems like a good change now!

PR #122, opened 2025-08-01
DEMOS-586: updated server mutators to use create instead of add

connor-parke-icf:
Changes the addX syntax to createX to align with the standards we're looking to enforce on the backend. Updated a few places in the client to where the type is used or the mutator usage is defined.

### DIFF
--- a/client/src/components/modal/DemonstrationModal.test.tsx
+++ b/client/src/components/modal/DemonstrationModal.test.tsx
@@ -30,6 +30,7 @@ vi.mock("hooks/useDemonstration", () => ({
   })),
 }));
 
+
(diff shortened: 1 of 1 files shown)


.devcontainer/devcontainer.json
```diff
@@ -32,7 +32,8 @@
     "customizations": {
         "vscode": {
             "extensions": [
-                "tomoyukim.vscode-mermaid-editor"
+                "tomoyukim.vscode-mermaid-editor",
+                "apollographql.vscode-apollo"
```
connor-parke-icf, line 36: added the apollographql extension to the devcontainer for gql syntax highlighting.

PR #129, opened 2025-08-05
Demos 378

connor-parke-icf:
This PR encapsulates the file uploading process. The flow is as follows (in broad terms):

- the client requests to upload a document. This request contains all the needed fields for a document, except for the s3path. 
- the document resolver will save the document to a document_pending_upload table. Afterward, the system will access the CDK and generate an S3Path point to a "upload" bucket to return to client. 
- The client will then upload the file to the s3 clean bucket path
- when the s3 bucket recieves the file, it is picked up by a SQS, which will trigger a GuardDuty process to scan the file for malware
- after the file is successfully scanned, a EventBridge is triggered which triggers the main lambda (fileprocess/index.js)
- that lambda will determine if the scan resulted in clean or infected. 
- - If infected, the lambda will leave the file in the "upload" bucket. 
- - If clean, the lambda will 
- - - move the file from the upload bucket to the clean bucket, assigning it a new s3path of form "bundle_id/document_id"
- - - trigger a stored procedure in the database to move the file data from the document_pending_upload table to the document table, assigning it the new s3path prefixed with the clean bucket's id. 

This PR includes all of the processes needed for this to happen, though no frontend component has been linked yet. This upload process is available on the gql server as uploadDemonstrationDocument, uploadAmendmentDocument, or uploadExtensionDocument. 

It also includes a set of scripts which can be invoked locally to initialize and test the fileuploading via localstack. Some components are not testable on localstack, however the inputs/outputs are simulated via the script.


lambdas/fileprocess/setup.sh
connor-parke-icf, line 1: script for setting up the s3, secretmanager, and lambda on localstack.

lambdas/fileprocess/setup.sh
```diff
@@ -0,0 +1,235 @@
...
+# Create S3 buckets (skip if they already exist)
+aws --endpoint-url=$LOCALSTACK_ENDPOINT s3 mb s3://upload-bucket --region $REGION 2>/dev/null || echo -e "${YELLOW}⚠️  upload-bucket already exists${NC}"
+aws --endpoint-url=$LOCALSTACK_ENDPOINT s3 mb s3://clean-bucket --region $REGION 2>/dev/null || echo -e "${YELLOW}⚠️  clean-bucket already exists${NC}"
+
+echo -e "${GREEN}✅ S3 buckets ready${NC}"
+
+echo -e "${YELLOW}📋 Step 2: Setting up Secrets Manager${NC}"
+
+# Check if secret already exists
+if aws --endpoint-url=$LOCALSTACK_ENDPOINT secretsmanager describe-secret \
```
connor-parke-icf, line 48: i am assuming that the db connection string will look like the example. with username password, host, port, and dbname. The dbpassword had to be passed in as an arg to make the commit checker happy

lambdas/fileprocess/setup.sh
```diff
@@ -0,0 +1,235 @@
...
+
+# Install dependencies if node_modules doesn't exist or package.json is newer
+if [ ! -d "node_modules" ] || [ "package.json" -nt "node_modules" ]; then
+    echo "Installing npm dependencies..."
+    npm install
+fi
+
+# Create deployment package
+echo "Creating deployment package..."
+zip -r fileprocess.zip index.js package.json package-lock.json node_modules/ -x "*.git*" "*.DS_Store*" >/dev/null
```
connor-parke-icf, line 109: assuming this is how the lambda will be deployed

lambdas/fileprocess/setup.sh
```diff
@@ -0,0 +1,235 @@
...
+        RETRY_COUNT=$((RETRY_COUNT + 1))
+    fi
+done
+
+if [ $RETRY_COUNT -eq $MAX_RETRIES ]; then
+    echo -e "${RED}❌ Lambda function did not become active within expected time${NC}"
+    exit 1
+fi
+
+# Create environment variables file
```
connor-parke-icf, line 157: passing in custom variables that i assume will be handled outside the lambda

lambdas/fileprocess/process-file.sh
connor-parke-icf, line 1: script to actually run the lambda. given a url and id, creates a test file, uploads it, and builds an EventBridge input to trigger it in either infected state or not. then checks the return of the lambda and state of s3. also propts user to check the db

lambdas/fileprocess/.gitignore
connor-parke-icf, line 1: ignoring test files and build artifacts.

lambdas/fileprocess/check-logs.sh
connor-parke-icf, line 1: quick log checker

lambdas/fileprocess/index.js
```diff
@@ -0,0 +1,243 @@
...
+
+const s3Config = process.env.AWS_ENDPOINT_URL
+  ? {
+      region: process.env.AWS_REGION,
+      endpoint: process.env.AWS_ENDPOINT_URL,
+      forcePathStyle: true,
+    }
+  : undefined;
+
+const s3 = new S3Client(s3Config);
```
connor-parke-icf, line 17: at the moment i have this conditional, since i couldnt get it to work with localstack as i had it before. Not sure why. This can be removed if needed.

lambdas/fileprocess/index.js
```diff
@@ -0,0 +1,243 @@
...
+  const secretData = JSON.parse(response.SecretString);
+  databaseUrlCache = `postgresql://${secretData.username}:${secretData.password}@${secretData.host}:${secretData.port}/${secretData.dbname}?schema=${dbSchema}`;
+  cacheExpiration = now + 60 * 60 * 1000;
+
+  return databaseUrlCache;
+}
+
+function isGuardDutyScanClean(guardDutyEvent) {
+  try {
+    if (guardDutyEvent["detail-type"] !== "GuardDuty Malware Protection Object Scan Result") {
```
connor-parke-icf, line 57: these checks might need updating, not sure.

lambdas/fileprocess/index.js
```diff
@@ -0,0 +1,243 @@
...
+    console.log("Processing summary:", results);
+
+    return { 
+      statusCode: 200, 
+      body: `Processed ${results.processedRecords} records: ${results.cleanFiles.length} clean files processed, ${results.infectedFiles.length} infected files skipped`
+    };
+
+  } catch (error) {
+    console.error("Lambda execution failed:", error);
+    throw new Error(`Lambda failed: ${error.message}`);
```
connor-parke-icf, line 233: tried to make it always throw error if actually fails, but report on status of infected or clean in the return if doesnt fail.

server/src/model/migrations/20250804171333_add_documents_pending_upload_table/migration.sql
```diff
@@ -0,0 +1,73 @@
...
+DECLARE
+    v_bundle_id UUID;
+BEGIN
+    SELECT bundle_id INTO v_bundle_id
+    FROM demos_app.document_pending_upload 
+    WHERE id = p_document_id;
+    
+    RETURN v_bundle_id;
+END;
+$$;
```
TARehman, line 73: Tiniest of nits but maybe add a newline here?

lambdas/fileprocess/.gitignore
```diff
@@ -0,0 +1,11 @@
...
+guardduty-event-*.json
+response-*.json
+processed-*
+test-file.txt
+
+# Lambda deployment artifacts
+fileprocess.zip
+lambda-env.json
+node_modules/
+package-lock.json
```
TARehman, line 11: Should probably have this be a pre-commit hook (ending files with a newline). I think there is one.

server/src/model/migrations/20250804171333_add_documents_pending_upload_table/migration.sql
TARehman, line 1: Blocker: You need to actually add the SQL to run the trigger and update the history table. The file in `/data/utilities` can be used to generate it, but you will also need to update the Python to allow that new model name.

server/src/model/document/documentResolvers.ts
```diff
@@ -1,21 +1,20 @@
 import { GraphQLError } from "graphql";
 
-import {
-  Bundle,
-  Document,
-} from "@prisma/client";
+import { Bundle, Document } from "@prisma/client";
```
TARehman, line 3: NB: I am still confused why some things get imported from `@prisma/client` and others don't. It feels inconsistent throughout. We should get something added to the README for the server explaining this behavior and expectations better.

server/src/model/document/documentResolvers.ts
```diff
@@ -1,21 +1,20 @@
...
 import { prisma } from "../../prismaClient.js";
 import { BundleType } from "../../types.js";
 import {
-  CreateAmendmentDocumentInput,
   CreateDemonstrationDocumentInput,
   CreateExtensionDocumentInput,
   UpdateAmendmentDocumentInput,
   UpdateDemonstrationDocumentInput,
   UpdateExtensionDocumentInput,
+  UploadAmendmentDocumentInput,
```
TARehman, line 14: Why is this only implemented for amendments? Wouldn't it make more sense to just write a general function for it and then alias it to all three, since they're all effectively the same process?

server/src/model/migrations/20250804171333_add_documents_pending_upload_table/migration.sql
```diff
@@ -0,0 +1,73 @@
...
+    FROM 
+        demos_app.document_pending_upload 
+    WHERE id = p_id;
+    
+    DELETE FROM demos_app.document_pending_upload
+    WHERE id = p_id;
+END;
+$$;
+
+CREATE FUNCTION demos_app.get_bundle_id_for_document(p_document_id UUID)
```
TARehman, line 60: I don't think you want to do this in SQL. This function is perfectly fine but this kind of thing should just be in the Prisma, where it is more testable anyway.

server/src/model/migrations/20250804171333_add_documents_pending_upload_table/migration.sql
```diff
@@ -0,0 +1,73 @@
...
+-- AddForeignKey
+ALTER TABLE "document_pending_upload" ADD CONSTRAINT "document_pending_upload_document_type_id_fkey" FOREIGN KEY ("document_type_id") REFERENCES "document_type"("id") ON DELETE RESTRICT ON UPDATE CASCADE;
+
+-- AddForeignKey
+ALTER TABLE "document_pending_upload" ADD CONSTRAINT "document_pending_upload_owner_user_id_fkey" FOREIGN KEY ("owner_user_id") REFERENCES "users"("id") ON DELETE RESTRICT ON UPDATE CASCADE;
+
+-- AddForeignKey
+ALTER TABLE "document_pending_upload" ADD CONSTRAINT "document_pending_upload_bundle_id_fkey" FOREIGN KEY ("bundle_id") REFERENCES "bundle"("id") ON DELETE RESTRICT ON UPDATE CASCADE;
+
+CREATE OR REPLACE PROCEDURE move_document_from_processing_to_clean(
```
TARehman, line 41: I don't know where this procedure would get stuck automagically without giving it a fully-qualified name. It would probably end up in `demos_app`, but if your search path doesn't contain `demos_app`, then invoking it might not work right. I think you might want to use a fully-qualified name here like you do in the actual code.

server/src/model/migrations/20250804171333_add_documents_pending_upload_table/migration.sql
```diff
@@ -0,0 +1,73 @@
...
+    p_id UUID,
+    p_s3_path TEXT
+)
+LANGUAGE plpgsql
+AS $$
+BEGIN
+    INSERT INTO demos_app.document (id, title, description, s3_path, owner_user_id, document_type_id, bundle_id, created_at, updated_at) 
+    SELECT 
+        id, title, description, p_s3_path, owner_user_id, document_type_id, bundle_id, created_at, updated_at
```
TARehman, line 50: Should we pass the S3 path here or just the bucket, and then construct the S3 path from it? The "risk" I see is that you could pass in an S3 path not aligned with the bundle_id and input ID, and so you could insert something nonsensical. We know everything that we need to get the path except the bucket already. `s3://{BUCKET}/{BUNDLE_ID}/{ID}`

server/src/model/migrations/20250804171333_add_documents_pending_upload_table/migration.sql
```diff
@@ -0,0 +1,73 @@
...
+ALTER TABLE "document_pending_upload" ADD CONSTRAINT "document_pending_upload_bundle_id_fkey" FOREIGN KEY ("bundle_id") REFERENCES "bundle"("id") ON DELETE RESTRICT ON UPDATE CASCADE;
+
+CREATE OR REPLACE PROCEDURE move_document_from_processing_to_clean(
+    p_id UUID,
+    p_s3_path TEXT
+)
+LANGUAGE plpgsql
+AS $$
+BEGIN
+    INSERT INTO demos_app.document (id, title, description, s3_path, owner_user_id, document_type_id, bundle_id, created_at, updated_at)
```
TARehman, line 48: We should probably wrap this into a transaction, which sort of weirdly requires you to do something like:

```
BEGIN
    BEGIN;
```

We can also rollback the transaction automatically if there's an issue, which is what you'd want here. The delete and the insert should happen as an atomic operation.

server/src/model/document/documentResolvers.ts
```diff
@@ -170,11 +156,39 @@ export const documentResolvers = {
           },
         },
       });
+
+      const uploadResult = documentPendingUpload as Document;
+      // Generate a presigned S3 upload URL
+      const s3ClientConfig = process.env.S3_ENDPOINT_LOCAL
```
TARehman, line 162: I don't love this but I don't know what the alternative is.

TARehman (changes requested):
I'm gonna pause at the moment, I started looking over the Lambda but I had several comments before that about the other section and I think it makes more sense to just leave those to start.

server/src/model/document/documentResolvers.ts
connor-parke-icf: updated to work with demonstrations and extensions

server/src/model/migrations/20250804171333_add_documents_pending_upload_table/migration.sql
connor-parke-icf: removed function, now is defined in lambda

server/src/model/migrations/20250804171333_add_documents_pending_upload_table/migration.sql
connor-parke-icf: i actually ran into this issue a bunch! there were a few changes in the lambda to attempt to force stuff to be in demos_app, Big find to simplify that mess

server/src/model/migrations/20250804171333_add_documents_pending_upload_table/migration.sql
TARehman: Nevermind. I realize now that this is downstream of the Lambda already having put the file somewhere. So it makes sense to store the actual value.

lambdas/fileprocess/check-logs.sh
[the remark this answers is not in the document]
connor-parke-icf: yeah thats basically it; just wanted a quick way to check the logs of the lambda output. Not really needed, but is useful

server/src/model/migrations/20250804171333_add_documents_pending_upload_table/migration.sql
connor-parke-icf: registered documentPendingUpload in python script, ran it to create triggers in migration.

lambdas/fileprocess/.gitignore
dustbuster: Another good one to add is `.DS_Store` it's a mac thing. 

```
.DS_Store files are hidden metadata files that Finder creates in each folder to store how that folder should be displayed.
```

deployment/lib/guardDutyS3.ts
```diff
@@ -0,0 +1,164 @@
...
+    const guardDutyPassRole = new aws_iam.Role(
+      this,
+      "GuardDutyMalwareProtectionPassRole",
+      {
+        roleName: `GuardDutyMalwareProtectionPassRole-${props.account}`,
+        assumedBy: new aws_iam.ServicePrincipal(
+          "malware-protection-plan.guardduty.amazonaws.com"
+        ),
+        description:
+          "An iam pass role for guardduty malware protection service to assume",
```
dustbuster, line 117: Huh. it is one word. Silly AWS.

Screenshot 2025-08-13 at 13 30 36

lambdas/fileprocess/check-logs.sh
```diff
@@ -0,0 +1,197 @@
...
+echo ""
+echo -e "${YELLOW}💡 Usage tips:${NC}"
+echo "  • Use './check-logs.sh 100' to see more lines"
+echo "  • Use './check-logs.sh 20 all' to see logs from all streams"
+echo "  • Logs are updated in real-time after Lambda invocations"
+echo ""
+echo -e "${YELLOW}🔧 Debugging tips:${NC}"
+echo "  • If logs are empty, try invoking the Lambda first"
+echo "  • Use 'docker logs demos_devcontainer-localstack-1' to see LocalStack logs"
+echo "  • Check that your Lambda function has console.log statements"
```
dustbuster, line 197: newline? nbd tho

server/src/server.ts
```diff
@@ -29,13 +29,17 @@ export const graphqlHandler = startServerAndCreateLambdaHandler(
...
         context: async ({ event, context }) => {
             // Add any shared context here, e.g., user authentication
+            console.log("ENTRY...")
             await databaseUrlPromise;
+            console.log("AFTER DB URL PROMISE")
             
             // Values from the user's id_token are set in /lambda_authorizer/index.mjs
             // and available here from event.requestContext.authorizer.<key>
-
+
```
dustbuster, line 38: Clean up this white space maybe? Still on NBD level.

dustbuster (approved):
Couple very minor comments. NBDs. But this is good man! Nicely done!

PR #136, opened 2025-08-07
DEMOS-557-new-fields

TARehman:
This PR adds two new columns to demonstration, CMCS Division and Signature Level. There's no enforcement of the proper values on the DB or server side; I added a custom TypeScript type with the allowable values, which should be usable to populate a dropdown as well. I looked into having the custom type be enforced in GQL, but GQL doesn't really have an equivalent to TS for a union of quoted strings that have spaces. The best it seems to do is either custom scalars, which is a fair bit of work, or enums, which cannot have spaces or quotes. We could use the enums, but I suspect that makes more sense if we are using them for natural keys in the DB context, which would involve adding constraints in the DB.


server/src/model/demonstration/demonstrationSchema.ts
```diff
@@ -7,12 +7,27 @@ import { State } from "../state/stateSchema.js";
...
+  - Division of Eligibility and Coverage Demonstrations
+  """
+  scalar CmcsDivision
+  """
+  A string representing a signature level. Expected values are:
+  - OA
+  - OCD
+  - OGD
+  """
+  scalar SignatureLevel
```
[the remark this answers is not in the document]
tresdonicf: Nice I'm a big fan of that stuff, for instance `type ErrorMessage = string` and then passing things around as `ErrorMessage` adds so much information to the codebase without requiring any extra infrastructure!

server/src/model/demonstration/demonstrationSchema.ts
```diff
@@ -62,12 +81,18 @@ export const demonstrationSchema = gql`
 `;
 
 export type DateTime = Date;
+export type SignatureLevel = "OA" | "OCD" | "OGD";
```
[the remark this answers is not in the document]
tresdonicf: 100% agree with adding a check for this to the resolvers

TARehman:
The fruit of some conversations between me and @connor-parke-icf on a path that hopefully reduces complexity, but still includes some of the data integrity stuff I think that @james-becwar-cms wants. This adds simplified constraints in the database, but maintains the export in the server layer too, so that those values are easy to make use of in the front-end. It also adds in documentation / an approach in the README. I think if this makes sense and seems tractable, we should move some of the other things that we added as more dynamic constraints toward this (for instance, it's increasingly seeming likely we won't need to have statuses that are configurable by the application).

server/src/model/demonstration/demonstrationSchema.ts
```diff
@@ -62,12 +81,25 @@ export const demonstrationSchema = gql`
...
+
+// Note: If changing either of these, be sure to update the related DB table as well!
+export const SIGNATURE_LEVEL = ["OA", "OCD", "OGD"] as const;
+export type SignatureLevel = (typeof SIGNATURE_LEVEL)[number];
+
+export const CMCS_DIVISION = [
+  "Division of System Reform Demonstrations",
+  "Division of Eligibility and Coverage Demonstrations",
+] as const;
+export type CmcsDivision = (typeof CMCS_DIVISION)[number];
```
tresdonicf, line 93: Could we add these into the `types.ts` file so they can be used by the frontend?

tresdonicf: Additionally just curious - why this pattern over `type SignatureLevel = "OA" | "OCD" | "OGD"`?

server/src/model/demonstration/demonstrationResolvers.ts
```diff
@@ -216,5 +234,13 @@ export const demonstrationResolvers = {
...
       });
     },
+
+    cmcsDivision: async (parent: Demonstration) => {
+      return parent.cmcsDivisionId;
+    },
+
+    signatureLevel: async (parent: Demonstration) => {
+      return parent.signatureLevelId;
+    },
```
connor-parke-icf, line 244: i feel like we still may be missing something here, but this is fine if it works. We may discover a more direct way to return these fields in the future

connor-parke-icf (approved):

server/src/model/demonstration/demonstrationSchema.ts
tresdonicf: Ahh okay - yeah the only file that's currently used by the frontend is `types.ts` (as sort of a place to see at a glance what is shared) and I like the idea of a `constants.ts` too for declaring some more of these enumerable values! We could open this up to use more files but figured for now this simplifies the code-sharing between the two. For that second item that is a great point that we'd need the full array before we can declare the types, good call!

TARehman: OK, I moved the definitions to the existing `types.ts` and `constants.ts` files.

connor-parke-icf:
This looks fine to me. I think we need to revisit how we're exporting types, as it appears instead of copying the server code to the node_modules of client, we are instead symlinking the folder to server. Which means, we still have a dependency on server during runtime, which is nonideal. Or at least, that is what appears to be the case.  We should go back and see if we can export the types and constants a different way as it appears something is off with the current config.

PR #139, opened 2025-08-10
DEMOS 512 Navigate to Specific Amendment/Extension

JeffreyM-GA:
Adds Amendment/Extension ID to query param when select
Activates relevant dropdown by default
image


client/src/pages/DemonstrationDetail/DemonstrationDetail.tsx
```diff
@@ -53,12 +53,12 @@ export const DemonstrationDetail: React.FC = () => {
   const location = useLocation();
 
   const queryParams = new URLSearchParams(location.search);
-  const [tab, setTab] = useState<TabType>(() => {
-    const amendmentParam = getQueryParamValue(queryParams, "amendment", "amendments");
-    const extensionParam = getQueryParamValue(queryParams, "extension", "extensions");
+  const amendmentParam = getQueryParamValue(queryParams, "amendment", "amendments");
```
connor-parke-icf, line 56: i think we should look to make this a single applicable value, i.e. instead of amendment or amendments, we should just use amendment (unless somewhere we are using both). singular makes more since to me, since we are now passing an id in.

client/src/components/table/tables/AmendmentTable.tsx
```diff
@@ -20,19 +21,32 @@ import {
 interface AmendmentTableProps {
   data: RawAmendment[];
   demonstrationId: string;
+  initiallyExpandedId?: string;
 }
 
-export function AmendmentTable({ data }: AmendmentTableProps) {
-  const [expanded, setExpanded] = React.useState({});
+export function AmendmentTable({ data, initiallyExpandedId }: AmendmentTableProps) {
+  const [expanded, setExpanded] = React.useState<ExpandedState>(() => {
```
connor-parke-icf, line 28: nice typing here

client/src/components/table/tables/ExtensionTable.tsx
```diff
@@ -20,19 +21,32 @@ import {
...
   const table = useReactTable({
     data,
     columns: ExtensionColumns,
     state: { expanded },
-    onExpandedChange: setExpanded,
+    onExpandedChange: handleExpandedChange,
     getCoreRowModel: getCoreRowModel(),
     getExpandedRowModel: getExpandedRowModel(),
     getRowCanExpand: () => true,
+    getRowId: (row) => row.id,
```
connor-parke-icf, line 49: I suspect this might already be available somewhere or defaulted to row.id in some way. Are we calling on getRowId anywhere?

connor-parke-icf (approved):
Looks good to me!

PR #140, opened 2025-08-11
DEMOS-525 add defaults for edit document modal flow

JeffreyM-GA:
Add default values for edit document modal

image


client/src/components/modal/document/DocumentModal.tsx
```diff
@@ -231,8 +241,9 @@ const BaseDocumentModal: React.FC<BaseDocumentModalProps> = ({
     documentModalType === "edit" ? "Edit Document" : "Add New Document";
 
   // TODO: get this from GQL hook
-  const [documentTitle, setDocumentTitle] = useState("Document title");
-  const [description, setDescription] = useState("");
+  const [documentTitle, setDocumentTitle] = useState(initialTitle);
+  const [description, setDescription] = useState(initialDescription);
+  const [selectedType, setSelectedType] = useState(initialType);
```
dustbuster, line 246: Is this TODO still accurate? 
Seems we are dynamically pulling the titles and descriptions. I think?

dustbuster (approved):
Looks good to me. I would just pull that TODO if accurate.

client/src/components/modal/document/DocumentModal.tsx
tresdonicf: I think this is moving us in the right direction - I've been doing some thinking on this and I'll share what I've been thinking about in office hours!

If anything - I would maybe just pass a `document: Document` into this component and pull the fields directly out of that to simplify the propslist

tresdonicf (approved):

PR #143, opened 2025-08-11
Demos 622 convert dayjs to date

connor-parke-icf:
this was easier than anticipated. Thankfully, MUI DatePicker has a adaptor build for Datefns, so we were able to basically swap everything out easily. Went through and removed the references of dayjs in the codebase. Now the server has no usages, and the client only has the dependencies pulled in my MUI.

### DIFF
--- a/client/src/components/modal/DemonstrationModal.test.tsx
+++ b/client/src/components/modal/DemonstrationModal.test.tsx
@@ -1,12 +1,17 @@
 import React from "react";
 
 import { ToastProvider } from "components/toast/ToastContext";
+import { Demonstration } from "demos-server";
 import { vi } from "vitest";
 
-import { fireEvent, render, screen, waitFor } from "@testing-library/react";
+import {
+  fireEvent,
+  render,
+  screen,
+  waitFor,
+} from "@testing-library/react";
 
 import { DemonstrationModal } from "./DemonstrationModal";
-import { Demonstration } from "demos-server";
 
 // Mock the useDemonstration hook
 vi.mock("hooks/useDemonstration", () => ({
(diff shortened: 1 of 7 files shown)


client/src/router/DemosRouter.tsx
```diff
@@ -24,7 +24,7 @@ export const DemosRouter = () => {
   return (
     <AuthProvider {...cognitoConfig}>
       <DemosApolloProvider>
-        <LocalizationProvider dateAdapter={AdapterDayjs}>
+        <LocalizationProvider dateAdapter={AdapterDateFns}>
```
tresdonicf, line 27: Do we need this adapter at the top level of the application? It seems `<LocalizationProvider>` comes from the `@mui/x-date-pickers` package so it seems like we should just be wrapping it at the site there

client/src/components/table/columns/DocumentColumns.tsx
```diff
@@ -1,14 +1,14 @@
...
 
-import dayjs, { Dayjs } from "dayjs";
 import { DocumentTableRow } from "hooks/useDocument";
 import { useDocumentType } from "hooks/useDocumentType";
 
 import { createColumnHelper } from "@tanstack/react-table";
 
 import { SecondaryButton } from "../../button/SecondaryButton";
 import { highlightCell } from "../KeywordSearch";
+import { format, isAfter, isBefore, isSameDay } from "date-fns";
```
tresdonicf, line 11: I do love the modularity of these - I wish the function name for `format` wasn't so generic hahh

tresdonicf (approved):
One question about `LocalizationProvider` being used in the router, otherwise looks great!

client/src/components/table/columns/DocumentColumns.tsx
connor-parke-icf: agreed; it almost makes me to want to namespace it or something. or have like datefn.format() or similar.

PR #152, opened 2025-08-15
Demos 625 consistent parsing of dates and datetime

connor-parke-icf:


server/src/model/demonstration/demonstrationSchema.ts
```diff
@@ -89,8 +88,8 @@ export interface Demonstration {
   expirationDate: Date;
   cmcsDivision?: CmcsDivision;
   signatureLevel?: SignatureLevel;
-  createdAt: DateTime;
-  updatedAt: DateTime;
+  createdAt: Date;
+  updatedAt: Date;
```
tresdonicf, line 92: It's interesting because while this makes sense and s what we talked about, seeing it like this almost makes me prefer the type alias approach - but nice to not have to declare that line over and over

client/eslint-rules/no-nonstandard-date-rendering.js
tresdonicf, line 1: Very cool to add another eslint rule for this!

client/eslint-rules/no-nonstandard-date-rendering.js
```diff
@@ -0,0 +1,64 @@
...
+
+const DATE_METHODS = [
+  "toLocaleDateString",
+  "toLocaleString",
+  "toISOString",
+  "toUTCString",
+  "toDateString",
+  "toTimeString",
+  "toJSON",
+];
```
tresdonicf, line 13: what does this list represent? Is it something like `DATE_METHODS_TO_AVOID`? I'm a little confused how this picks up on `date-fns.format`

client/src/util/RenderDate.tsx
```diff
@@ -0,0 +1,16 @@
+import { format } from "date-fns";
+
+export function renderDate(date: Date) {
+  // eslint-disable-next-line no-nonstandard-date-rendering/no-nonstandard-date-rendering
```
tresdonicf, line 4: Can instead do `/* eslint-disable no-nonstandard-date-rendering/no-nonstandard-date-rendering */` at the top of the file for it to work for the whole file

client/src/util/RenderDate.tsx
tresdonicf, line 1: This file is the only thing I might have some helpful comments on:

I would perhaps name this file `formatDate.ts` since it's not a `.tsx` file and it's not responsible for rendering but just formatting. Type annotations on the function outputs would be really nice here also.

I've been happy about our streak of not including a `utils` folder but I knew this day would come 🫠

server/src/model/demonstration/demonstrationSchema.ts
connor-parke-icf: yeah, i tried looking for a few different ways to more explicitly set this as a proper DateTime type. but ultimately, all of them wouldnt be useful for type checking. One such method was called branding, where you init an object with an extra field. but that seemed like too much, and would require that we "brand" the date as a Date or DateTIme each time we used it.

client/src/util/RenderDate.tsx
connor-parke-icf: ah good call. Was probably leftover from my first attempt at this, where i tried to make it a component that you would render as <USDate date={date}/>, but there were too many places where we were rendering explicitly as a string to be passed in somewhere for this to be worth it. So converted it to a generic function. 

Perhaps we'll find a better home for it in the future; its lonely in that folder (at least it has a test file buddy)

client/src/util/RenderDate.tsx
connor-parke-icf: ehhh went ahead and changed it to be file-wide

client/src/util/RenderDate.tsx
connor-parke-icf: changed verbiage to be 'format' instead of 'render'. Added string output Type

PR #162, opened 2025-08-18
Add `data-testid` to components

tresdonicf:
DEMOS-651: Add `data-testid` to some UI components, this is helpful for QA to develop their automation scripts and it's also helpful for us in writing unit tests.

For now I've just kept this to interactive elements like <button>, <input>, <select>, <option>, etc but in the future we may just want a process for adding these ad-hoc when QA requests a new element they'd like to grab.

I made a little script that helps to ascertain if a certain tag does not have a `data-testid` already - left some things alone for the time being but think I got to most everything

Screenshot 2025-08-18 134104

### DIFF
--- a/client/src/components/modal/document/DocumentModal.test.tsx
+++ b/client/src/components/modal/document/DocumentModal.test.tsx
@@ -180,7 +180,7 @@ describe("RemoveDocumentModal", () => {
 
   it("calls deleteDocumentsTrigger when Remove is clicked", async () => {
     setup(["test-document-id"]);
-    fireEvent.click(screen.getByRole("button", { name: /Remove/ }));
+    fireEvent.click(screen.getByTestId(CONFIRM_REMOVE_BUTTON_TEST_ID));
     await waitFor(() => {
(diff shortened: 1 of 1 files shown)


client/scripts/identify-missing-data-testid.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+TAGS_TO_SCAN = ["button", "input", "select", "option", "textarea"]
+JSX_TAG_REGEX = re.compile(
+    r'<(' + '|'.join(TAGS_TO_SCAN) + r')\b[\s\S]*?>', re.DOTALL
+)
+DATA_TESTID_REGEX = re.compile(r'data-testid\s*=')
+
+
+def scan_file(filepath):
```
dustbuster, line 14: I always think this is "mos def" a function! Or def Jam records... Python is weird.

dustbuster (approved):
Looks good! Only suggestion is to make sure there's no dup ids? If that's relevant.

PR #163, opened 2025-08-18
DEMOS-623 - stable auth, automatic user upsert, a single canonical source of user truth via UserContext, working logout, Fix profile header

dustbuster:
* **Auth + user bootstrapping**
   * In `context`, verified JWT, then **upserted** the user by `cognitoSubject` and loaded roles.
   * Guarded nullable `email` and added safe fallbacks to satisfy TS/Prisma.
* **GraphQL & resolvers**
   * Added `getCurrentUser` query and resolver.
   * Fixed resolver typings (no implicit `any`, proper return types).
   * Cleaned helper lookups to use the right ID (DB id vs Cognito sub).
* **Front-end state via context**
   * Built a `UserContext` that fetches `getCurrentUser` once and provides `hasRole`, `refresh`.
   * Refactored **PrimaryLayout**, **DefaultHeaderLower**, and **ProfileBlock** to use `UserContext` (no more per-component `user(id)` queries).
* **Mocks & tests**
   * Swapped mocks to return **`getCurrentUser`** instead of `user(id)`.
   * Removed `HEADER_LOWER_QUERY` / `PROFILE_BLOCK_QUERY` mocks.
* **Logout flow (Hosted UI)**
   * Fixed Cognito sign-out: pointed to `/logout`, ensured **trailing slash** in `post_logout_redirect_uri`, and matched allowed URLs.
   * Centralized sign-in/out: created `AuthActions` + `AuthButtons`, and used them in `ProfileBlock`.
* Updated mocks to work with the new query
* Added a check for existing auth header, e.g. `event.requestContext.authorizer`
   * If decoded claims exist, otherwise decode!
   * Having the decode makes auth work locally. 
   * On lambda the, claim should be hydrated. (further dev env testing req'd)

Screen recording of the user being generated. 
https://github.com/user-attachments/assets/fe90eae3-ad7a-4609-84c1-6c563d3c55c4

Screenshot 2025-08-20 at 11 48 50

Upon resetting this back to `server/.env` -> `BYPASS_AUTH=true`

It still works just fine with Bypass user
Screenshot 2025-08-21 at 09 08 30

Just wanted to verify that!

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -30,7 +30,6 @@
         "@types/react-dom": "^19.1.2",
         "@vitejs/plugin-react": "^4.3.4",
         "@vitest/coverage-v8": "^3.1.2",
-        "demos-server": "file:../server",
         "depcheck": "^1.4.7",
         "eslint": "^9.33.0",
         "eslint-config-prettier": "^10.1.5",
@@ -58,12 +57,14 @@
     "../server": {
       "name": "demos-server",
       "version": "0.1.0",
-      "dev": true,
+      "extraneous": true,
       "dependencies": {
         "@apollo/server": "^4.12.2",
         "@as-integrations/aws-lambda": "^3.1.0",
+        "@aws-sdk/client-s3": "^3.859.0",
         "@aws-sdk/client-secrets-manager": "^3.840.0",
-        "@prisma/client": "^6.7.0",
+        "@aws-sdk/s3-request-presigner": "^3.859.0",
+        "@prisma/client": "^6.11.1",
         "graphql-scalars": "^1.24.2",
         "graphql-tools": "^9.0.18",
         "jsonwebtoken": "^9.0.2",
@@ -72,6 +73,9 @@
         "react": "^19.1.0"
       },
       "devDependencies": {
+        "@aws-sdk/client-s3": "^3.859.0",
+        "@aws-sdk/client-secrets-manager": "^3.840.0",
+        "@aws-sdk/s3-request-presigner": "^3.859.0",
         "@eslint/js": "^9.27.0",
         "@faker-js/faker": "^9.8.0",
         "@graphql-tools/merge": "^9.0.24",
@@ -87,7 +91,7 @@
         "graphql": "^16.11.0",
         "jest": "^29.7.0",
         "prettier": "^3.5.3",
-        "prisma": "^6.6.0",
+        "prisma": "^6.11.1",
         "ts-jest": "^29.3.3",
         "tsc-watch": "^7.1.1",
         "tsx": "^4.19.4",
@@ -4464,10 +4468,6 @@
       "funding": {
         "url": "https://github.com/sponsors/ljharb"
       }
-    },
-    "node_modules/demos-server": {
-      "resolved": "../server",
-      "link": true
     },
     "node_modules/depcheck": {
       "version": "1.4.7",
@@ -10327,21 +10327,6 @@
       "integrity": "sha512-a4UGQaWPH59mOXUYnAG2ewncQS4i4F43Tv3JoAM+s2VDAmS9NsK8GpDMLrCHPksFT7h3K6TOoUNn2pb7RoXx4g==",
       "dev": true,
       "license": "ISC"
-    },
-    "node_modules/yaml": {
-      "version": "2.8.1",
-      "resolved": "https://registry.npmjs.org/yaml/-/yaml-2.8.1.tgz",
-      "integrity": "sha512-lcYcMxX2PO9XMGvAJkJ3OsNMw+/7FKes7/hgerGUYWIoWu5j/+YQqcZr5JnPZWzOsEBgMbSbiSTn/dv/69Mkpw==",
-      "dev": true,
-      "license": "ISC",
-      "optional": true,
-      "peer": true,
-      "bin": {
-        "yaml": "bin.mjs"
-      },
-      "engines": {
-        "node": ">= 14.6"
-      }
     },
     "node_modules/yargs": {
       "version": "16.2.0",


server/src/auth/auth.config.ts
```diff
@@ -5,9 +5,9 @@ interface AuthConfig {
 };
 
 export const LOCAL_AUTH_CONFIG: AuthConfig = {
-  jwksUri: 'https://cognito-idp.us-east-1.amazonaws.com/us-east-1_A7CaR2Wo3/.well-known/jwks.json',
-  issuer: 'https://cognito-idp.us-east-1.amazonaws.com/us-east-1_A7CaR2Wo3',
-  audience: '5km9thunj8g6qd32s5et2i8pga',
+  jwksUri: 'https://cognito-idp.us-east-1.amazonaws.com/us-east-1_FCc2lmZDJ/.well-known/jwks.json',
+  issuer: 'https://cognito-idp.us-east-1.amazonaws.com/us-east-1_FCc2lmZDJ',
+  audience: "5p61qososiui75cmclcift45oi",
```
dustbuster, line 10: THIS HAS BEEN THE BANE OF MY EXISTENCE!! I did not know this was here!

client/src/components/auth/AuthActions.tsx
tresdonicf, line 1: 2 things here:

1. This should be `.ts` instead of `.tsx` since it doesn't render react
2. This should be in `hooks/` rather than `components/` 

I am a bit skeptical because part of me thinks this should just live on our `SignIn` and `SignOut` buttons and that components shouldn't be calling this functionality themselves

client/src/hooks/useCurrentUser.ts
tresdonicf, line 1: This seems like it might belong in `queries` rather than `hooks`

client/src/components/header/ProfileBlock.tsx
```diff
@@ -1,74 +1,52 @@
...
     return (
-      <div>
-        <button>Log In</button>
-      </div>
+      <button
+        onClick={signIn}
+        className="text-white bg-black/20 hover:bg-black/30 rounded px-2 py-1"
+      >
+        Log In
+      </button>
```
tresdonicf, line 22: Here if we just call `<SignInButton>` we don't need to import `useAuthActions`

client/src/router/DemosApolloProvider.tsx
```diff
@@ -1,64 +1,54 @@
...
+  const authLink: ApolloLink = useMemo(
+    () =>
+      setContext((_, { headers }) => {
+        const token = auth.user?.id_token ?? auth.user?.access_token ?? null;
+        console.debug("[Apollo] has token:", !!token);
+        // dev-only breadcrumb
+        if (import.meta.env.DEV) {
+          // will spam if left in; comment out once verified
+          console.debug("[Apollo] has token:", !!token);
+        }
```
tresdonicf, line 31: if (isLocalDevelopment()) !

client/src/router/DemosApolloProvider.tsx
```diff
@@ -1,64 +1,54 @@
...
 
-  if (shouldUseMocks()) {
-    return (
-      <MockedProvider mocks={ALL_MOCKS} addTypename={false}>
-        {children}
-      </MockedProvider>
-    );
-  } else {
-    const apolloClient = createApolloClient(auth);
+  const client = useMemo(
```
tresdonicf, line 44: Hmm, how often is this dependency array set to change? I'm not certain about the increased code here but not seeing anything that seems wrong either. Just not sure why it's in a `useMemo`

server/src/auth/auth.util.ts
```diff
@@ -13,57 +15,77 @@ import {
...
+      // Lightweight, structured debug (no full token)
+      console.log("[jwt]", {
+        kid: (jwt.decode(token, { complete: true }) as any)?.header?.kid,
+        token_use: peek.token_use,
+        iss: (peek as { iss?: string })?.iss,
+        aud: (peek as { aud?: string })?.aud,
+        expectedIssuer: config.issuer,
+        expectedAudience: config.audience,
+      });
+    }
```
tresdonicf, line 69: Do we still need to decode this in the server? I thought all of this was being handled by the lambda authorizer now

I was kinda under the impression we might be able to scrap this file for the time being

server/src/local-server.ts
```diff
@@ -6,21 +6,56 @@ import {
...
+      });
+
+      const roles = await getUserRoles(sub);
+      return { user: { id: dbUser.id, roles } };
+    } catch (err) {
+      console.error("[auth] context error:", err);
+      return { user: null };
+    }
   },
 });
```
tresdonicf, line 58: Let's encapsulate some of this in a function and maybe in a different file to keep this file clean (also I think this needs to be added to `server.ts` as well

server/.env.example
```diff
@@ -1,6 +1,8 @@
 DATABASE_URL="postgresql://postgres:postgres@db:5432/demos?schema=demos_app" # pragma: allowlist secret
 BYPASS_AUTH=true
 ALLOW_SEED=true
+
 ALLOW_INTROSPECTION=true
 UPLOAD_BUCKET="upload-bucket"
 S3_ENDPOINT_LOCAL="http://localstack:4566"
+AUTH_DEBUG=false
```
tresdonicf, line 8: Probably not worth adding to the `.env.example` since I can't imagine we'll need this separately from a different type of `isLocalDevelopment` on the server

client/src/components/user/UserContext.tsx
```diff
@@ -0,0 +1,67 @@
...
+    }),
+    [currentUser, loading, error, refetch]
+  );
+  return <Ctx.Provider value={value}>{children}</Ctx.Provider>;
+}
+
+export function getCurrentUser() {
+  const ctx = useContext(Ctx);
+  if (!ctx) {
+    if (import.meta.env.DEV) {
```
tresdonicf, line 54: `if (isLocalDevelopment())` might be useful here as well

tresdonicf (commented):
This is looking great! Some suggestions on re-organizing the code to match the other sorts of declarations we have but they're pretty small. The 2 main things I'm curious about are:

1. Do we need the decode token stuff on the server still? I think the authorizer is handling all of this now and if we can get rid of it that would be nice
2. I think we should encapsulate the `createUserIfNotExists()` logic into a function and also call it in `server.ts`

client/src/components/header/ProfileBlock.tsx
```diff
@@ -1,74 +1,52 @@
...
-          className="absolute top-12 min-w-full right-0 bg-white border border-gray-300 rounded shadow-lg z-11"
-        >
+        <ul className="absolute top-12 right-0 min-w-full bg-white border border-gray-300 rounded shadow-lg z-50">
           <li className="hover:bg-gray-100 cursor-pointer p-1">
-            <a>Logout</a>
+            <button
+              onClick={(e) => { e.stopPropagation(); setOpen(false); signOut(); }}
+            >
+              Logout
+            </button>
```
dustbuster, line 46: Same thing with `<Signoutbutton />`

client/src/queries/userQueries.ts
```diff
@@ -7,3 +7,18 @@ export const USER_OPTIONS_QUERY = gql`
...
+      cognito_subject
+      username
+      email
+      full_name
+      display_name
+      created_at
+      updated_at
+    }
+  }
+`;
```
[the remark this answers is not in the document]
dustbuster: Yeah this is probably a left over.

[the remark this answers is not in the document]
dustbuster: This is for sure being used! It breaks the app without it!

Screenshot 2025-08-21 at 13 02 35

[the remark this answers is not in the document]
dustbuster: I bet we can replace `USER_OPTIONS_QUERY`. But let's do that in the next effort.

client/src/components/user/UserContext.tsx
dustbuster: Good idea ⚡

client/src/router/DemosApolloProvider.tsx
dustbuster: Great question! I do not remember! 

useMemo is a React hook that memoizes (caches) the result of a calculation so it doesn’t get recomputed on every render unless its dependencies change.

This refactor uses cache and saves overhead. 

```
Why use it?

Performance: Avoid re-running expensive calculations every render.

Stable references: Keep the same object/function reference between renders so children don’t re-render unnecessarily (or to avoid effect loops).

When not to use it:
If the calculation is cheap, skip it. useMemo adds a tiny overhead.
Not for side effects (use useEffect for those).
Common comparisons

useMemo vs useCallback: useCallback(fn, deps) is basically useMemo(() => fn, deps)—it memoizes a function instead of a value.

useMemo vs useEffect: useMemo runs during render to compute a value; useEffect runs after render to perform side effects.

File: client/src/router/DemosRouter.tsx [hunk 11]
```diff
@@ -12,6 +12,7 @@ import { IconLibrary } from "pages/debug/IconLibrary";
 import { DemosApolloProvider } from "./DemosApolloProvider";
 import { isLocalDevelopment } from "config/env";
 import { EventSandbox } from "pages/debug/EventSandbox";
+import { UserProvider } from "components/user/UserContext";
 
 export const DemosRouter = () => {
   // TODO: When we know what IDM integration looks like
```
dustbuster, line 18: We can pull this out.

server/src/auth/auth.config.ts
dustbuster: Helps to login when you get the config correct.

server/src/auth/auth.util.ts
dustbuster: I was able to remove most of this, but yeah i could not validate on localhost 4000, it was totally confounding and this helped me find the cognito config in the backend.

server/src/local-server.ts
dustbuster: I was JUST. thinking that.

client/src/components/header/Header.test.tsx
```diff
@@ -1,84 +1,74 @@
...
+vi.mock("react-oidc-context", () => ({
+  useAuth: () => ({
+    isAuthenticated: true,
+    isLoading: false,
+    user: { id_token: "fake-token" },
+    signinRedirect: vi.fn(),
+    signoutRedirect: vi.fn(),
+    removeUser: vi.fn(),
+  }),
+}));
```
dustbuster, line 29: Yeet! Got the login mock to work!

client/src/components/header/QuickLinks.test.tsx
```diff
@@ -0,0 +1,13 @@
...
+import { QuickLinks } from "./QuickLinks";
+
+describe("QuickLinks", () => {
+  it("renders all quick links", () => {
+    render(<QuickLinks />);
+    expect(screen.getByRole("link", { name: /Admin/i })).toBeInTheDocument();
+    expect(screen.getByRole("link", { name: /Notifications/i })).toBeInTheDocument();
+    expect(screen.getByRole("link", { name: /Help/i })).toBeInTheDocument();
+  });
+});
```
dustbuster, line 13: Maybe frivolous... but idk.

client/src/components/user/UserContext.tsx
```diff
@@ -0,0 +1,56 @@
...
+  return <Ctx.Provider value={value}>{children}</Ctx.Provider>;
+}
+
+export function getCurrentUser() {
+  const ctx = useContext(Ctx);
+  if (!ctx) {
+    throw new Error("getCurrentUser must be used within <UserProvider>");
+  }
+  return ctx;
+}
```
dustbuster, line 56: Let's take all this messy auth, And slap it behind a context. `getCurrentUser()` Boom!

client/src/layout/PrimaryLayout.tsx
```diff
@@ -1,31 +1,46 @@
...
+      </ToastProvider>
+    );
+  }
 
-  // Use different user ID based on environment - always string now
-  const userId = shouldUseMocks() ? "2" : "00000000-1111-2222-3333-123abc123abc";
+  // Optional: surface auth/data errors
+  if (error) {
+    // You can render something nicer or trigger a sign-out here
+    console.error("[PrimaryLayout] currentUser error:", error);
```
dustbuster, line 36: nuked alll these AI comments.

client/src/components/auth/AuthLinks.tsx
```diff
@@ -0,0 +1,56 @@
...
+          onClick?.(e);
+        }}
+        {...rest}
+      >
+        {children}
+      </a>
+    );
+  }
+);
+SigninLink.displayName = "SigninLink";
```
tresdonicf, line 32: This pattern still gets me, I've never seen it before - I wonder if there's a way to get react dev tools to work nicer with all our components and to avoid specifying this

client/src/mock-data/userMocks.ts
```diff
@@ -25,153 +24,70 @@ export const johnDoe: User = {
...
+      displayName: "John",
+      roles: [],
     },
   },
+};
+
+export const userMocks: MockedResponse[] = [
+  // It takes two of these apparently to hydrate the Profile block.
+  { request: { query: GET_CURRENT_USER_QUERY }, result: currentUserResult },
+  { request: { query: GET_CURRENT_USER_QUERY }, result: currentUserResult },
```
dustbuster, line 52: There is a dupe request thing going on. This fixes the header. But should probs fix the core issue at some point. But yeah, mocks on, bypass auth on and this populates the header as expected.

server/src/auth/auth.util.ts
```diff
@@ -1,213 +1,203 @@
...
 const config = getAuthConfig();
 
 export interface GraphQLContext {
   user: null | {
     id: string;
-    name: string;
+    sub: string;
     roles: string[] | null;
+    displayName?: string;
   };
```
tresdonicf, line 19: Maybe this should `extend BaseContext` from apollo

server/src/auth/auth.util.ts
```diff
@@ -1,213 +1,203 @@
...
-          return reject(
-            new GraphQLError("User is not authenticated", {
-              extensions: { code: "UNAUTHENTICATED", http: { status: 401 } },
-            }),
-          );
-        }
-        resolve(decoded as DecodedJWT);
-      },
-    );
+    jwt.verify(token, getKey, verifyOpts, (err, decoded) => {
```
dustbuster, line 52: Factored this array into a var `verifyOpts`

server/src/auth/auth.util.ts
```diff
@@ -1,213 +1,203 @@
...
+    update: { ...(email ? { email } : {}) },
+    create: {
+      cognitoSubject: sub,
+      username,
+      email: emailForCreate,
+      fullName,
+      displayName,
+    },
+  });
+}
```
dustbuster, line 98: If user is not in DB, let's g'head and make them.

tresdonicf (approved):
Shared some comments on thoughts I have but nothing that I can't live with / we can't revisit as we use the `Context` approach!

client/src/components/auth/AuthLinks.tsx
dustbuster: Honestly. I dislike it. I pulled them.

client/src/components/header/HeaderConfigContext.test.tsx
```diff
@@ -19,21 +20,28 @@ const TestConsumer = () => {
...
+function renderWithProviders(ui: React.ReactNode) {
+  return render(
+    <MockedProvider mocks={userMocks} addTypename={false}>
+      <UserProvider>{ui}</UserProvider>
+    </MockedProvider>
+  );
+}
+
 describe("HeaderConfigProvider", () => {
   it("renders defaultLowerContent initially inside Header", () => {
```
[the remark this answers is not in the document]
dustbuster: That would replace the MockedProvider?

server/src/server.ts
```diff
@@ -1,46 +1,39 @@
-import { ApolloServer, BaseContext } from '@apollo/server';
```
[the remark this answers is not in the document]
dustbuster: I considered. This and i asked the old robot to provides some umm. Context? 
And it suggested this. 

Short answer: you don’t need to “maintain” Apollo’s BaseContext. It’s just an empty object type from @apollo/server and isn’t useful for your app. 

What you should maintain is a single app-wide context type that your resolvers expect, and then extend it per runtime (HTTP vs Lambda) when you build the context.

server/src/auth/auth.util.ts
dustbuster: Perhaps! Let's get this chunk out and we can always come back to this.

server/src/server.ts
```diff
@@ -1,46 +1,98 @@
...
+    const auth = (rc.authorizer ?? {}) as { claims?: Record<string, unknown> };
+    const claims = auth.claims;
+    if (claims && typeof claims === "object") {
+      const sub = typeof claims.sub === "string" ? claims.sub : "";
+      if (!sub) return null;
+      const email = typeof claims.email === "string" ? claims.email : undefined;
+      return { sub, email };
+    }
+  }
+  return null;
```
dustbuster, line 55: @tresdonicf I did take one of your suggestions. Though it's hard to know if it'll work until i merge. 

but we were decoding every time, so now we check for decode claims before decoding them. I thought this important for dev since we really have no way to test without merging it.

server/src/auth/auth.util.ts
```diff
@@ -99,10 +104,26 @@ export const getCognitoUserInfoForLambda = (
...
+//       data: {
+//         cognitoSubject: sub,
+//         email,
+//         fullName: `${given_name ?? ""} ${family_name ?? ""}`.trim(),
+//         username: email, // or another field if available
+//         displayName: given_name ?? email,
+//       },
+//     });
+//   }
+// };
```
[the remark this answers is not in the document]
dustbuster: Nuked

server/src/model/user/userResolvers.ts
```diff
@@ -13,6 +13,15 @@ export const userResolvers = {
...
     },
+    currentUser: async (_p, _a, ctx) => {
+      if (!ctx.user) return null;
+      try {
+        return await prisma().user.findUnique({ where: { id: ctx.user.id } });
+      } catch (e) {
+        console.error("[currentUser] resolver error:", e);
+        throw e;
+      }
+    },
```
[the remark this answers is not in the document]
dustbuster: > yeah, i did.

client/src/queries/userQueries.ts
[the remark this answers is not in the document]
dustbuster: Nuked.

dustbuster:
> This is looking great! Some suggestions on re-organizing the code to match the other sorts of declarations we have but they're pretty small. The 2 main things I'm curious about are:
> 
> 1. Do we need the decode token stuff on the server still? I think the authorizer is handling all of this now and if we can get rid of it that would be nice
> 2. I think we should encapsulate the `createUserIfNotExists()` logic into a function and also call it in `server.ts`

Added a check to see if claims are decoded. We aren't user `createUserIfNotExists` but yes I centalized the user create code. Both good finds!

server/src/auth/auth.util.ts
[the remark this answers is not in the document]
dustbuster: yeah, revisted. local, we need to decode, but on server this should be provided. We'll prolly have an adjustment period after deploy

client/src/components/header/HeaderConfigContext.test.tsx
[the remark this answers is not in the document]
dustbuster: done

client/src/components/auth/AuthLinks.tsx
```diff
@@ -0,0 +1,57 @@
...
+import { useAuthActions } from "./AuthActions";
+import { getCognitoConfig, getCognitoLogoutUrl } from "router/cognitoConfig";
+
+type AnchorProps = React.AnchorHTMLAttributes<HTMLAnchorElement> & {
+  children?: React.ReactNode;
+};
+
+/** Link that triggers the Cognito sign-in flow. */
+export const SigninLink = forwardRef<HTMLAnchorElement, AnchorProps>(
+  ({ children = "Sign In", onClick, href, ...rest }, ref) => {
```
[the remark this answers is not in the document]
dustbuster: Nuked those children!

server/src/auth/auth.util.ts
dustbuster: Let's make a TODO. I'd like to wrap this mega PR

PR #166, opened 2025-08-19
DEMOS-594 (1/2) Dispatch Delete Document Query

tresdonicf:
For DEMOS-594 - dispatches the delete document query to the server

### DIFF
--- a/client/src/components/modal/document/DocumentModal.tsx
+++ b/client/src/components/modal/document/DocumentModal.tsx
@@ -8,6 +8,8 @@ import { BaseModal } from "components/modal/BaseModal";
 import { useToast } from "components/toast";
 import { tw } from "tags/tw";
 import { TextInput } from "components/input";
+import { useMutation } from "@apollo/client";
+import { DELETE_DOCUMENTS_QUERY } from "queries/documentQueries";
 
 type DocumentModalType = "add" | "edit";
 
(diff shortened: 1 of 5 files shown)


server/src/model/document/documentResolvers.ts
```diff
@@ -201,12 +201,6 @@ export const documentResolvers = {
       });
     },
 
-    deleteAmendmentDocument: async (_: undefined, { id }: { id: string }) => {
-      return await prisma().document.delete({
-        where: { id: id },
-      });
-    },
-
```
tresdonicf, line 209: These all used the same logic, so just simplified this replacing it with `deleteDocuments()` which is just a change to use `deleteMany` rather than `delete`

client/src/components/modal/document/DocumentModal.tsx
```diff
@@ -394,22 +396,26 @@ export const RemoveDocumentModal: React.FC<{ documentIds: string[]; onClose: ()
...
-      const isMultiple = documentIdList.length > 1;
+      setIsDeleting(true);
+      await deleteDocumentsTrigger({ variables: { ids: documentIdList } });
+
+      const isMultipleDocuments = documentIdList.length > 1;
       showWarning(
-        `Your document${isMultiple ? "s" : ""} ${isMultiple ? "have been" : "has been"} removed.`
+        `Your document${isMultipleDocuments ? "s" : ""} ${isMultipleDocuments ? "have been" : "has been"} removed.`
       );
       onClose();
```
tresdonicf, line 414: Odd, this is showing up different in GH than it actually is in the branch

Image

server/src/model/document/documentSchema.ts
```diff
@@ -74,13 +74,11 @@ export const documentSchema = gql`
...
     uploadDemonstrationDocument(input: UploadDemonstrationDocumentInput!): Document
     updateDemonstrationDocument(id: ID!, input: UpdateDemonstrationDocumentInput!): Document
-    deleteDemonstrationDocument(id: ID!): Document
     uploadAmendmentDocument(input: UploadAmendmentDocumentInput!): Document
     updateAmendmentDocument(id: ID!, input: UpdateAmendmentDocumentInput!): Document
-    deleteAmendmentDocument(id: ID!): Document
     uploadExtensionDocument(input: UploadExtensionDocumentInput!): Document
     updateExtensionDocument(id: ID!, input: UpdateExtensionDocumentInput!): Document
-    deleteExtensionDocument(id: ID!): Document
+    deleteDocuments(ids: [ID!]!): [ID!]!
```
connor-parke-icf, line 81: I think we have the ability here to return a whole document object instead of just IDs, i think we should still allow that graphql feature. It may be useful to return more attributes if someone wanted to see, for instance the s3path of the deleted document. I havnt actually played around with what prisma actually returns, so it may be just the id from deletemany. Not sure. But if prisma does allow returning the entire deleted object, i think we should still return it.

connor-parke-icf (approved):
just one comment: if possible i dont think we should restrict the return type of the deleteMany to just the ids. I would rather return an array of Documents, in case a client wants to get other fields back from their deleted document

server/src/model/document/documentSchema.ts
tresdonicf: I definitely think this could prove to be useful in the future! For now since it's just used in a modal with the count I figured it's best to keep it simple and just return the IDs of the documents that were deleted.

The other thing to mention is that If we're deleting these from a table or something we should have the document records already in memory on the client and so we could use fields like `.s3Path`, `title` directly from there, even if we just get the IDs back of the documents that were successfully deleted (to filter our list of "want to delete" documents)

PR #169, opened 2025-08-20
(2/2) DEMOS-594: Dispatch Requests for Create/Edit DocumentModal

tresdonicf:
For DEMOS-594 - this PR dispatches the requests for the Create / Edit Document Modals

I was able to simplify some of the schema / resolvers on the backend as a part of this work - I did notice that the `UploadDocument` input doesn't have an `Upload` scalar - which is what would be needed for us to actually attach the file (through the S3 presigned URL stuff)

Decided to make this a separate piece of work, will attach the ticket here when it's created. For now we're just sending the modal fields but anything additional should be simpler to plumb through! Also noticed some fields that shouldn't be part of updating or uploading inputs like `s3Path` & `ownerUserId` (these should come from the server).

### DIFF
--- a/server/src/model/document/documentSchema.ts
+++ b/server/src/model/document/documentSchema.ts
@@ -67,6 +67,7 @@ export interface UploadDocumentInput {
 }
 
 export interface UpdateDocumentInput {
+  id: string;
   title?: string;
   description?: string;
   documentType?: string;
--- a/client/src/components/modal/document/DocumentModal.tsx
+++ b/client/src/components/modal/document/DocumentModal.tsx
@@ -412,6 +412,7 @@ export const EditDocumentModal: React.FC<{
 
   const handleEdit = async (modalFields: DocumentModalFields) => {
     const updateDocumentInput: UpdateDocumentInput = {
+      id: modalFields.id,
       title: modalFields.title,
       description: modalFields.description,
       documentType: modalFields.documentType,


server/src/model/document/documentSchema.ts
```diff
@@ -6,7 +6,8 @@ import { Amendment, Extension } from "../modification/modificationSchema.js";
 import { User } from "../user/userSchema.js";
 
 export const documentSchema = gql`
-  union Bundle = Demonstration | Amendment
+  union Bundle = Demonstration | Amendment | Extension
```
connor-parke-icf, line 9: good catch here

server/src/model/document/documentSchema.ts
```diff
@@ -101,53 +60,15 @@ export interface Document {
...
-  documentTypeId: string;
-  extensionId: string;
-}
-
-export interface UpdateExtensionDocumentInput {
+export interface UpdateDocumentInput {
+  id: string;
   title?: string;
   description?: string;
-  s3Path?: string;
```
connor-parke-icf, line 149: i like this change, makes sense that the s3 path is uneditable. If for some reason it needs to be edited we should make a dedicated resolver for it.

connor-parke-icf (approved):
no real notes; we discussed a lot of the server changes already. I would look to maybe put the uploadDocument queries within the document model on the clientside, since i image the only place we would be calling these queries is from the modal, but thats not a big deal either way.

PR #171, opened 2025-08-22
Demos 623 bypass auth

dustbuster:
We have the authorizer in the code. so made it easy to tweak this to work. (probably)

### DIFF
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -90,6 +90,7 @@ export const graphqlHandler = startServerAndCreateLambdaHandler(
       await databaseUrlPromise;
 
       const claims = extractAuthorizerClaims(event as LambdaEvent);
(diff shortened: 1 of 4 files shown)


server/src/server.ts
```diff
@@ -13,52 +13,28 @@ import {
...
+  if (!sub) return null;
+
+  // Optional: show what's coming through from the authorizer
+  // (Be careful to not log PII in production)
+  console.log(
+    "[lambda] authorizer keys:",
+    Object.keys(auth).join(", ") || "<none>"
+  );
+
+  return { sub, email };
```
dustbuster, line 37: I know there's stupid comments everywhere. They will get pulled. This is all debugging and hopeful swings!

james-becwar-cms (approved):

PR #175, opened 2025-08-26
Demos temp modal

nunesmlucas:


client/src/components/header/DefaultHeaderLower.tsx
```diff
@@ -97,10 +114,16 @@ export const DefaultHeaderLower: React.FC = () => {
         )}
       </div>
 
-      {modalType === "create" && <DemonstrationModal mode="add" onClose={() => setModalType(null)} />}
-      {modalType === "document" && <AddDocumentModal onClose={() => setModalType(null)} />}
-      {modalType === "amendment" && <CreateNewModal mode="amendment" onClose={() => setModalType(null)} />}
-      {modalType === "extension" && <CreateNewModal mode="extension" onClose={() => setModalType(null)} />}
+      {modalType === "create" && (
```
connor-parke-icf, line 117: i imagine here is the bulk of the intended changes, swapping out the modal call for the new Dialog call

client/src/components/header/HeaderConfigContext.tsx
connor-parke-icf, line 1: this seems like a prettier/auto change. Hard to catch prettier sometimes, but generally (imo) we should strive to have updates only to files we're making real changes to. Not a blocker though

client/src/components/table/tables/DocumentTable.tsx
```diff
@@ -74,17 +70,19 @@ function DocumentActionButtons({
 }: DocumentActionButtonsProps) {
   return (
     <div className="flex gap-2 ml-4">
-      <CircleButton ariaLabel="Add Document" onClick={() => onShowModal("add")}>
+      <CircleButton name="add-document" ariaLabel="Add Document" onClick={() => onShowModal("add")}>
         <ImportIcon />
       </CircleButton>
       <CircleButton
+        name="edit-document"
```
connor-parke-icf, line 77: glad to see these updated; it kept showing up for me as well

client/src/hooks/useDemonstrationOptions.ts
```diff
@@ -0,0 +1,25 @@
+import { useEffect } from "react";
+
+import { useDemonstration } from "./useDemonstration";
+
+export const useDemonstrationOptions = () => {
```
connor-parke-icf, line 5: i like this pattern of having a translation layer between the Demonstrations and the Dropdown to translate the raw data into a more friendly Option. i imagine we could use this in different places, though im not a huge fan of the useDemonstration call, since we are looking to decentralize the gql queries out of the hooks folder and into more usage-specific places. I see this as something to iterate on, but not a blocker.

client/src/mock-data/event/eventMocks.ts
```diff
@@ -38,6 +37,9 @@ const testEvent2: Event = {
   withRole: testRole,
   route: "/demonstrations",
   createdAt: new Date("2025-01-01T11:00:00Z"),
+  eventData: {
```
connor-parke-icf, line 40: looks like this may have been unintended? might be a relic of a previous iteration of events

client/src/mock-data/demonstrationMocks.ts
```diff
@@ -565,7 +263,343 @@ export const demonstrationMocks: MockedResponse[] = [
...
+            extensions: [],
+          },
+        ],
+      },
     },
   },
 ];
+
+// Extract the data from the DEMONSTRATIONS_PAGE_QUERY mock for use in tests
+const demonstrationsPageQueryMock = demonstrationMocks.find(
```
connor-parke-icf, line 594: im a little confused on the usage of this, but i do see you implemented some failure queries for testing which is fantastic.  im looking to revamp a lot of this mock data structuring to be a little easier to use but i dont want to lose any important testing equipment in the process

client/src/pages/DemonstrationDetail/DemonstrationDetailModals.tsx
```diff
@@ -16,60 +19,33 @@ export type DemonstrationModalDetails = {
...
+      <AmendmentDialog mode="add" demonstrationId={demonstration.id} onClose={onCloseEntityModal} />
     )}
 
     {entityCreationModal === "extension" && (
-      <CreateNewModal
-        mode="extension"
-        data={{ demonstration: demonstration.id }}
-        onClose={onCloseEntityModal}
-      />
+      <ExtensionDialog mode="add" demonstrationId={demonstration.id} onClose={onCloseEntityModal} />
```
connor-parke-icf, line 41: love how much cleaner this reads

client/vitest.setup.ts
```diff
@@ -3,6 +3,34 @@ import "@testing-library/jest-dom";
 import React from "react";
 import { vi } from "vitest";
 
+// Mock HTML dialog element for tests
```
connor-parke-icf, line 6: oooo, does this replace all the dialog behaviors across all the tests? if so thats really nice. I can imagine a number of usecases where we will 99% of the time want to skip/modify the rendering of specific components (for example, the MUI date picker is awful to test with, i would love to never have to try to interact with it in tests again lol)

connor-parke-icf (approved):
my main goal here, since there are a lot of changes, is to get this merged in and get main up to date so we dont have to keep revising and running into merge conflicts. That being said, there are a few spots that i worry are unintended changes, or maybe reversions of some behaviors implemented between this ticket's conception and now. I dont _think_ many of my comments are are blocking, but may be worth checking out in case something is off

client/src/mock-data/event/eventMocks.ts
nunesmlucas: It's required in the Event type but optional in LogEventInput, so I am leaving it as an empty object for now then.

PR #177, opened 2025-08-26
DEMOS-580-collapse-migrations

TARehman:
This PR takes the dozen or so migrations we had and collapses them into a single one that reflects the current state of the code as it is on this branch. It also does the following.

 - Fixes a small column ordering item on `demonstration` and `modification`.
 - Correctly makes `effectiveDate` and `expirationDate` optional on `demonstration`. I forgot to do this before when I said I would. 😅

I did this by taking the following steps:

 - Delete all the migration files.
 - Run `prisma migrate reset` to clean my database.
 - Run `primsa migrate dev` to create a new migration.
 - Restore the original initial baseline, and copy the new migration's code into it.
 - Run the Python code to generate historical triggers.
 - Add in manual items that were added in previous migrations.


TARehman:
I'm tagging this with `DO NOT MERGE` not because it doesn't need reviews, but simply because the actual merge process will involve resetting the dev environment, and so it shouldn't just be merged.

server/src/model/modification/modification.prisma
connor-parke-icf, line 1: i assume this is the column ordering you mention in the description, to put the "meta" fields at the end?

server/src/model/demonstration/demonstration.prisma
```diff
@@ -1,17 +1,17 @@
...
-  projectOfficerUserId  String   @map("project_officer_user_id") @db.Uuid
+  effectiveDate         DateTime? @map("effective_date") @db.Date
+  expirationDate        DateTime? @map("expiration_date") @db.Date
+  cmcsDivisionId        String?   @map("cmcs_division_id")
+  signatureLevelId      String?   @map("signature_level_id")
+  demonstrationStatusId String    @map("demonstration_status_id")
+  stateId               String    @map("state_id")
+  projectOfficerUserId  String    @map("project_officer_user_id") @db.Uuid
+  createdAt             DateTime  @default(now()) @map("created_at") @db.Timestamptz()
+  updatedAt             DateTime  @updatedAt @map("updated_at") @db.Timestamptz()
```
connor-parke-icf, line 14: hard to see in the diff, but this just moves the "meta" fields to the end of the table yeah?

server/src/model/demonstration/demonstrationHistory.prisma
```diff
@@ -6,15 +6,15 @@ model DemonstrationHistory {
   bundle_type_id        String
   name                  String
   description           String
-  effectiveDate         DateTime     @map("effective_date") @db.Date
-  expirationDate        DateTime     @map("expiration_date") @db.Date
+  effectiveDate         DateTime?    @map("effective_date") @db.Date
```
connor-parke-icf, line 9: good catch

server/src/model/migrations/20250623155458_init_baseline/migration.sql
connor-parke-icf, line 1: really hard to capture the essence of the change by just looking at the diff, so i cant really tell if everything is accounted for. Nonetheless, the process followed seems solid; basically a delete and recreate. this is where having a more fleshed out testing suite for the server would come in handy. I suggest before the next compression we get a handle on that so we more definitively verify the change. Nonetheless, this all looks good to me.

server/src/model/modification/modification.prisma
TARehman: Yes, that's right. Just a change. The canonical order doesn't even really matter, its more of a nit than anything. Once we have fully deployed it won't be correctable again.

PR #183, opened 2025-08-29
DEMOS-669: Application Workflow Framework

tresdonicf:
For DEMOS-669, this adds the framework for using the application workflow to summon different phases of the application process.

There are a couple of items that are a bit more involved and would likely benefit from a "polish" step for a couple of tricky items that I didn't want to spend too much time on here, namely:

1. The vertical dividers
2. The current phase should be a larger scale than the others

These will just take a bit of time getting the styling right to display these in a way that's consistent and clean. I've asked for a ticket to take care of these items, will attach it here when created

In the screen recording below you'll notice the "Phase Content" being rendered just as text, but these are components that are ready to be used and so we can do our development for particular phases in those already existing components

https://github.com/user-attachments/assets/79e8f358-75aa-4aa6-a348-42b583a938ed


client/src/components/application/phase-selector/PhaseSelector.tsx
```diff
@@ -0,0 +1,108 @@
...
+  "OGC & OMB": "not_started",
+  "Approval Package": "not_started",
+  "Post Approval": "not_started",
+};
+
+const MOCK_PHASE_DATE_LOOKUP: Partial<Record<PhaseName, Date>> = {
+  Concept: new Date(2024, 4, 20),
+  "State Application": new Date(2024, 4, 22),
+  Completeness: new Date(2024, 11, 31),
+};
```
connor-parke-icf, line 67: i imagine this is here until we have a more rigid data structure for phases in the api?

dustbuster (approved):
Looks good! Got a comment. but it's probably not relevant yet.

client/src/components/application/phase-selector/PhaseSelector.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+export const DisplayPhase = ({ selectedPhase }: { selectedPhase: PhaseName }) => {
+  const PhaseComponent = PHASE_COMPONENTS_LOOKUP[selectedPhase];
+
+  return (
+    <div className="border border-dashed w-full h-full min-h-64">
+      <PhaseComponent />
+    </div>
+  );
+};
```
connor-parke-icf, line 108: This doesnt matter like at all, but to me it reads better if all the callee components are above the main caller component (so the caller is at the bottom of the file). If i understand this flow correctly, I would move the DisplayPhase above the PhaseSelector. Of course, its visually lower on the page, so the way you have it makes sense too. I would maybe remove the export on it, unless we want to test it in isolation

client/src/components/application/ApplicationWorkflow.tsx
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+import { PhaseSelector } from "./phase-selector/PhaseSelector";
+import { DemonstrationStatusBadge } from "./DemonstrationStatusBadge";
+
+export type DemonstrationStatus = "under_review" | "approved" | "rejected";
```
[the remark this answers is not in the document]
connor-parke-icf: i believe we're going to pull the demonstration statuses from the server as a const eventually.  I think we determined that theyre constant enough to call static.

client/src/components/application/DemonstrationStatusBadge.tsx
```diff
@@ -0,0 +1,40 @@
+import React from "react";
+import { HomeIcon, ReviewIcon, SaveIcon } from "icons";
+import { tw } from "tags/tw";
+import { DemonstrationStatus } from "./ApplicationWorkflow";
+
+const BASE_STYLES = tw`whitespace-nowrap h-full min-h-3 flex items-center gap-xs ml-auto text-lg`;
+
+export const DemonstrationStatusBadge = ({
```
connor-parke-icf, line 8: this is nice, id like to pull this into the demonstrations table once we combine demonstrations, amendments, and extensions into one thing (can probably be done before then, but might be easier to just wait)

connor-parke-icf (approved):
This is fantastic. Love having a bunch of small files each with clear responsibilities. A couple of suggestions/questions, but very far from anything blocking. great work!

james-becwar-cms (approved):

client/src/components/application/ApplicationWorkflow.tsx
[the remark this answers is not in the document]
tresdonicf: Ahh good to know! Yeah for now I was just going off of the types of badges I had seen in the mockups (though definitely don't think they're complete)

PR #186, opened 2025-09-02
Demos 683 add new amendment or extension from demonstrations tab

connor-parke-icf:
Adds a button on the Demonstration Detail page header which can be used to create a new Amendment or Extension.  Options there only include creating an amendment or an extension. When either are selected, the standard create modal for amendments and extensions are opened, however with the demonstration prefilled. This is the same behavior as the Create New buttons on the Amendments and Extensions tab. When the demonstrationId exists, the field is made to be disabled. 

The getAllDemonstrations method in the useDemonstrations hook was removed as well, replaced by defining the query within the useDemonstrationOptions hook. An additional mock was created to prevent refetching the queries, which would lead to a refetch of the main demonstrations page query. Components should not share the exact queries where possible, as refetching those queries will refresh all the components that use that same query. 

image
image


client/src/components/dialog/AmendmentDialog.test.tsx
```diff
@@ -21,17 +21,14 @@ vi.mock("components/toast", () => ({
...
-      error: null,
-    },
+vi.mock("hooks/useDemonstrationOptions", () => ({
+  useDemonstrationOptions: () => ({
+    demoOptions: [
+      { label: "Test Demo 1", value: "demo-1" },
+      { label: "Test Demo 2", value: "demo-2" },
+    ],
+    loading: false,
+    error: null,
```
connor-parke-icf, line 31: the getAllDemonstrations hook was removed, instead fetching from the useDemonstrationOptions component directly. More on this further down.

client/src/hooks/useDemonstrationOptions.ts
```diff
@@ -1,25 +1,39 @@
...
 
-import { useDemonstration } from "./useDemonstration";
+export const GET_DEMONSTRATION_OPTIONS_QUERY = gql`
+  query GetDemonstrationOptions {
+    demonstrations {
+      id
+      name
+    }
+  }
+`;
```
connor-parke-icf, line 11: GetDemonstrationOptions query moved into the useDemonstrationOptions hook. Eventually, i would like to move this into the Dialog itself, but this is fine for now. Main issue was that opening the modal on the Demonstrations page would cause a refetch of the Demonstrations page. This caused the error where no more mocks existed, as they shared the same mock. Now they are distinct queries that, even though fetch the same data, are treated as separate queries under the hood.

client/src/mock-data/demonstrationMocks.ts
```diff
@@ -192,6 +193,16 @@ export const demonstrationMocks: MockedResponse[] = [
       },
     },
   },
+  {
```
connor-parke-icf, line 196: simple mock created

client/src/pages/DemonstrationDetail/DemonstrationDetailHeader.tsx
```diff
@@ -116,26 +125,44 @@ export const DemonstrationDetailHeader: React.FC<DemonstrationDetailHeaderProps>
...
+                </button>
+                <button
+                  data-testid="button-create-new-extension"
+                  onClick={() => handleSelect("Extension")}
+                  className="w-full text-left px-1 py-[10px] hover:bg-gray-100"
+                >
+                  Extension
+                </button>
+              </div>
+            )}
```
connor-parke-icf, line 161: main change: adds a create button with options for either creating an amendment or extension. Uses same dialog structure as found on other buttons.

PR #188, opened 2025-09-02
DEMOS-555-contacts-assignment-model

TARehman:
Merges in existing changes to the data model with the originally-planned version which would implement contacts as essentially roles, based on long conversations and requirements gathering with David. The layout is getting out of hand but we're reaching a point where that's hard to avoid. I will start working up some tickets based on this but wanted to get it where you in particular could see it @james-becwar-cms.

ETA: I should note that we will enforce that every demonstration or modification has a primary project officer by way of using a deferred constraint trigger. This will allow us to defer evaluation of the trigger to the end of a transaction. This would allow us to insert a new demonstration or modification, and then subsequently insert the primary project officer record into `primary_person_demonstration_role`. Then, at the end of the transaction, it will be confirmed.


TARehman:
I forgot the document_pending_upload table entirely, but I added it in. Right now, everything in this design is about assigning permissions at either the system or the demonstration level. I don't know if we want to retitle that, given that there's a lot of complexity in the idea of being given a role on an amendment but not on the original demonstration. @james-becwar-cms I would probably lean toward leaving it titled as is?

data/docs/DEMOS_Data_Model.mmd
```diff
@@ -103,145 +138,176 @@ erDiagram
...
-    text title
+    uuid id PK
+    text name
     text description
     text s3_path
-    uuid owner_user_id FK "FK: users.id"
-    text document_type_id FK "FK1: phase_document_type.(phase_id, document_type_id)"
-    uuid bundle_id FK, UK "FK: bundle.id<br>UK: (id, bundle_id, phase_id)"
-    text phase_id FK, UK "FK1: phase_document_type.(phase_id, document_type_id)<br>UK: (id, bundle_id, phase_id)"
+    uuid owner_user_id FK "FK026: owner_user_id ∈ users.id"
```
james-becwar-cms, line 205: I think this is right, but it got me thinking we should come up with some guidence when we want to use user vs using person.

james-becwar-cms (approved):

PR #189, opened 2025-09-03
Demos 681 updates for effective and expiration dates

connor-parke-icf:
main change is simple, but a tiny change in each of the mocks and a linter change to allow for null-union types, which better represent the data coming back from gql over undefined/optional types.  

Main change is to make Effective and Expiration dates optional, changing how we are rendering dates, and updating the dialogs such that only the edit dialogs have fields for those. That trickled into the create resolvers, so removed these fields from the input types as well. Updated the mocks and tests to suit, including breaking out the Demo detail header into its own test file. 

Comments throughout the code for more details

### DIFF
--- a/client/src/components/dialog/document/DocumentDialog.test.tsx
+++ b/client/src/components/dialog/document/DocumentDialog.test.tsx
@@ -1,14 +1,17 @@
 import "@testing-library/jest-dom";
 
 import React from "react";
+
 import { ToastProvider } from "components/toast/ToastContext";
 import { describe, expect, it, vi } from "vitest";
+
 import { act, fireEvent, render, screen, waitFor } from "@testing-library/react";
+
 import {
-  RemoveDocumentDialog,
   AddDocumentDialog,
-  EditDocumentDialog,
   DocumentDialogFields,
+  EditDocumentDialog,
+  RemoveDocumentDialog,
 } from "./DocumentDialog";
 
 const mockQuery = vi.fn();
--- a/client/src/components/dialog/document/DocumentDialog.tsx
+++ b/client/src/components/dialog/document/DocumentDialog.tsx
@@ -1,22 +1,24 @@
-import React, { useRef, useState, useCallback, forwardRef } from "react";
+import React, { forwardRef, useCallback, useEffect, useRef, useState } from "react";
+
+import { Button, ErrorButton, SecondaryButton } from "components/button";
+import { BaseDialog } from "components/dialog/BaseDialog";
+import { ErrorIcon, ExitIcon, FileIcon } from "components/icons";
+import { TextInput } from "components/input";
+import { AutoCompleteSelect } from "components/input/select/AutoCompleteSelect";
+import { Option } from "components/input/select/Select";
+import { useToast } from "components/toast";
+import { Document, DocumentType, UpdateDocumentInput, UploadDocumentInput } from "demos-server";
+import { DOCUMENT_TYPES } from "demos-server-constants";
 import { useFileDrop } from "hooks/file/useFileDrop";
 import { ErrorMessage, UploadStatus, useFileUpload } from "hooks/file/useFileUpload";
-import { ErrorButton, Button, SecondaryButton } from "components/button";
-import { AutoCompleteSelect } from "components/input/select/AutoCompleteSelect";
-import { ErrorIcon } from "components/icons";
-import { useToast } from "components/toast";
-import { tw } from "tags/tw";
-import { TextInput } from "components/input";
-import { Document, DocumentType, UploadDocumentInput, UpdateDocumentInput } from "demos-server";
-import { useMutation } from "@apollo/client";
 import {
   DELETE_DOCUMENTS_QUERY,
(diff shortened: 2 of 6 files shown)


client/src/util/formatDate.ts
```diff
@@ -1,14 +1,14 @@
 /* eslint-disable no-nonstandard-date-formatting/no-nonstandard-date-formatting */
 import { format } from "date-fns";
 
-export function formatDate(date: Date): string {
-  return format(date, "MM/dd/yyyy");
+export function formatDate(date?: Date): string {
+  return date ? format(date, "MM/dd/yyyy") : "--/--/----";
```
connor-parke-icf, line 5: main change, made dates render like "--/--/----" if they dont exist. made up placeholders for the timestamp and isoString. but i dont think we have usecases for that yet, so i expect these may change at some point.

server/src/model/demonstration/demonstrationResolvers.ts
```diff
@@ -25,17 +22,12 @@ export const demonstrationResolvers = {
...
-      { input }: { input: CreateDemonstrationInput },
-    ) => {
+    createDemonstration: async (_: undefined, { input }: { input: CreateDemonstrationInput }) => {
       const {
         demonstrationStatusId,
         stateId,
         userIds,
         projectOfficerUserId,
-        effectiveDate,
-        expirationDate,
```
connor-parke-icf, line 38: effective and expiration dates are no longer used for creation, so updated to remove them from the create resolvers. They still exist in the update resolvers.

server/src/model/demonstration/demonstrationSchema.ts
```diff
@@ -24,8 +24,8 @@ export const demonstrationSchema = gql`
     id: ID!
     name: String!
     description: String!
-    effectiveDate: Date!
-    expirationDate: Date!
+    effectiveDate: Date
+    expirationDate: Date
```
connor-parke-icf, line 28: change is consistent between both Demonstrations and Modifications.

server/src/model/demonstration/demonstrationSchema.ts
```diff
@@ -84,8 +79,8 @@ export interface Demonstration {
   id: string;
   name: string;
   description: string;
-  effectiveDate: Date;
-  expirationDate: Date;
+  effectiveDate: Date | null;
+  expirationDate: Date | null;
   cmcsDivision?: CmcsDivision;
   signatureLevel?: SignatureLevel;
```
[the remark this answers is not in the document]
connor-parke-icf: I want to change them, but its outside the scope of the ticket; i will open up a follow up ticket to go through the server code and flag the nullables as such. https://jiraent.cms.gov/browse/DEMOS-721

client/src/util/formatDate.ts
```diff
@@ -1,14 +1,14 @@
 /* eslint-disable no-nonstandard-date-formatting/no-nonstandard-date-formatting */
 import { format } from "date-fns";
 
-export function formatDate(date: Date): string {
-  return format(date, "MM/dd/yyyy");
+export function formatDate(date: Date | null): string {
```
tresdonicf, line 4: I do not think this should take a possible null. It creates a large surface area for silent errors in getting dates back from the server.

client/src/components/application/phase-selector/PhaseDate.tsx
```diff
@@ -32,9 +32,7 @@ export const PhaseDate = ({ phaseStatus, date }: { phaseStatus: PhaseStatus; dat
       <span className={PHASE_TEXT_COLOR_LOOKUP[phaseStatus]}>
         {PHASE_DATE_TEXT_LOOKUP[phaseStatus]}
       </span>
-      <span className={PHASE_DATE_STYLE_LOOKUP[phaseStatus]}>
-        {date ? formatDate(date) : "--/--/----"}
-      </span>
+      <span className={PHASE_DATE_STYLE_LOOKUP[phaseStatus]}>{formatDate(date)}</span>
```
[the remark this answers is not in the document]
tresdonicf: It's not clear we want this behavior in every part of the application - I think we should leave it to the parts of the application we've explored for now rather than making it part of the library function.

client/src/mock-data/documentMocks.ts
```diff
@@ -42,7 +42,7 @@ export const mockDocuments: MockDocument[] = [
     owner: mockUsers[7],
     createdAt: new Date(2025, 0, 4),
   },
-];
+] as const satisfies MockDocument[];
```
tresdonicf, line 45: Really we should avoid `satisfies`. If we're not able to make the type contract work in the type signature, we shouldn't proceed, we should find another type to make that contract work.

tresdonicf (changes requested):
I think we should step back the `formatDate` changes and keep that function a bit more principled.

nunesmlucas (approved):

connor-parke-icf:
ended up reverting the format date changes, actually was very good that i did as one of the behaviors, prefilling the date for a form field, should be prefilled to the empty string instead of "--/--/----", so great suggestion @tresdonicf!

PR #190, opened 2025-09-03
DEMOS-710: Add typescript checking to pipelines

cms-jesse:
The main change for this PR was to add a separate typescript check to the PR checks. The tsc cli is now run after `npm run lint`. A build was also added to the client and server just for additional verification before being merged

The other changes in this PR were to correct the existing typescript issues that had gone unnoticed. One unique issue is related to crossover between the ui and server. I have called it out in the comments


client/types/shims.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module "graphql-tag" {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  export const gql: any;
+}
```
cms-jesse, line 4: This was added (as well as the change in the tsconfig.json) to correct an issue that may not be immediately visible locally. There are files that exist under the server project, but are directly imported into the client code. These files make use of `gql` from `graphql-tag`. If server dependencies have not been installed within the server directory, typescript throws an error for those files. In jenkins, UI and Server projects are compiled separately, so within the UI checks, the server dependencies were never installed. So CI runs show this issue even if its not visible locally since you likely have both projects initialized. 

There are likely other ways to deal with this issue

tresdonicf (approved):
LGTM! Thanks for adding this!

PR #202, opened 2025-09-04
Demos 701 make documents table non queryful 2

connor-parke-icf:

PR #203, opened 2025-09-04
DEMOS-679 Phase 1 - Steps 1 and 2

nunesmlucas:
# DEMOS-679 Phase 1 Implementation

## Overview
Implements Concept Phase functionality allowing users to upload Pre-Submission documents and complete the first phase of the approval workflow.

### Step 1 - Upload
- Upload button opens phase-specific modal titled "Pre-Submission Document"
- Document type defaults to "Pre-Submission" (can change to "General File")
- Documents display in phase with delete functionality

### Step 2 - Verify/Complete  
- Date field auto-populates from document upload timestamp
- Date becomes required when documents are uploaded
- Demonstration Type dropdown available

### Button Logic
- **Finish**: Disabled until documents uploaded AND date populated
- **Skip**: Enabled initially, disabled after any activity (documents or date)

### Validation
- Cross-validation between documents and date fields
- Real-time button state updates
- Warning messages for validation errors

### Components Created
- `ConceptPhase.tsx` (325 lines) - Main phase component
- `ConceptPreSubmissionUploadDialog.tsx` (40 lines) - Upload modal
- Enhanced `DocumentDialog.tsx` with title override support

### Testing
- 39 total tests (26 ConceptPhase + 13 Upload Dialog)
- All acceptance criteria covered

### TODOs
- Navigate to State Application on Skip
- Wire real document deletion 
- Mark phase as "Completed" in UI
- Connect to production GraphQL endpoints

image
image

### DIFF
--- a/server/src/auth/auth.util.ts
+++ b/server/src/auth/auth.util.ts
@@ -6,6 +6,7 @@ import { APIGatewayProxyEventHeaders } from "aws-lambda";
 import { SecretsManagerClient, GetSecretValueCommand } from "@aws-sdk/client-secrets-manager";
 import { getAuthConfig } from "./auth.config.js";
 import { prisma } from "../prismaClient.js";
+import { PERSON_TYPES } from "../constants.js";
 
 const config = getAuthConfig();
 
@@ -121,11 +122,22 @@ async function ensureUserFromClaims(claims: Claims) {
     return existingUser;
   }
 
+  // Derive a safe personTypeId from the role claim (or default)
+  const resolvedPersonTypeId = (() => {
+    // we started requiring a roles without actually adding the role feature.
+    if (!role) return "demos-cms-user";
+    // role may be a CSV or a single value; pick the first matching PERSON_TYPES
+    const parts = role.split(/[,\s]+/).filter(Boolean);
+    for (const part of parts) {
+      const match = (PERSON_TYPES as readonly string[]).find((p) => p === part);
+      if (match) return match;
+    }
+    return "demos-cms-user";
+  })();
+
   const person = await prisma().person.create({
     data: {
-      personType: {
-        connect: { id: role },
-      },
+      personTypeId: resolvedPersonTypeId,
       email: emailForCreate,
       fullName,
       displayName,
--- a/server/src/server.ts
+++ b/server/src/server.ts
@@ -17,16 +17,17 @@ import type {
   APIGatewayProxyEventHeaders,
 } from "aws-lambda";
 
-type JwtClaims = { sub: string; email?: string };
+type JwtClaims = { sub: string; email?: string; role?: string };
 
 function extractAuthorizerClaims(event: APIGatewayProxyEvent): JwtClaims | null {
   // In REST custom authorizer, requestContext.authorizer is a flat map of strings
   const auth = (event.requestContext?.authorizer ?? {}) as Record<string, unknown>;
   const sub = typeof auth.sub === "string" && auth.sub.length > 0 ? auth.sub : null;
   const email = typeof auth.email === "string" ? auth.email : undefined;
+  const role = typeof auth.role === "string" ? auth.role : undefined;
   if (!sub) return null;
 
-  return { sub, email };
+  return { sub, email, role };
 }
 
 function withAuthorizerHeader(


client/src/components/application/phases/ConceptPhase.tsx
```diff
@@ -1,5 +1,239 @@
...
+const GET_CONCEPT_DOCUMENTS = gql`
+  query GetConceptDocuments($demonstrationId: ID!) {
+    conceptDocuments: documents(demonstrationId: $demonstrationId, phase: "concept") {
+      id
+      title
+      uploadedAt
+      sizeMB
+    }
+  }
+`;
```
connor-parke-icf, line 32: all of the data attached to the demonstration, including all of its documents, should already be on the page already from the toplevel page load. So we should probably look to filter those documents rather than refetching them (assuming we're sticking with the current data fetching pattern).  Just made a small PR that moves the document fetching from the documents table to the main page query, for reference: https://github.com/Enterprise-CMCS/demos/pull/202

client/src/components/application/phases/ConceptPhase.tsx
```diff
@@ -1,5 +1,325 @@
...
+    setMockDocuments((prev) => prev.filter((doc) => doc.id !== docId));
+  };
+
+  const clearAllMockDocuments = () => {
+    setMockDocuments([]);
+  };
+
+  return (
+    <div>
+      {process.env.NODE_ENV === "development" && (
```
tresdonicf, line 145: This should use `isLocalDevelopment()`to check for if we're running locally

Additional NBD: it would be nice for readability to declare this component above and call it here (like a sub-component)

client/src/components/dialog/document/ConceptPreSubmissionUploadDialog.tsx
```diff
@@ -0,0 +1,40 @@
+import React from "react";
+
+import { AddDocumentDialog, DocumentDialogFields } from "components/dialog/document/DocumentDialog";
+import { DocumentType } from "demos-server";
+
+const DOCUMENT_TYPE_SUBSET: DocumentType[] = ["Pre-Submission", "General File"];
+
+type Props = {
+  isOpen: boolean;
+  onClose: () => void;
```
tresdonicf, line 10: NBD - I don't think we need to keep passing in `onClose` - that was so that we could use it to do `modalType == "..." && <RenderComponent>` but now since the dialog can close it self that shouldn't be needed. in some places it's still rendered like that so it would be a larger effort to remove it all together, but for new instances we can just pass `onClose = () => {}` if it's required

tresdonicf (commented):
Overall it's looking great! A few items I'd like to see some changes on:

1. using `isLocalDevelopment()` instead of `process.env`
2. bundleId and demonstrationId, maybe I misunderstand but I think we only want to deal in one of those
3. A few named units / sub-components in the <ConceptPhase> such as `<Step1 ...>`, `<FileBadge ...>` would be really nice to anchor the reader

client/src/components/application/phase-selector/PhaseSelector.test.tsx
```diff
@@ -2,10 +2,18 @@ import React from "react";
...
 describe("PhaseSelector", () => {
   it("renders all phase names", () => {
-    render(<PhaseSelector />);
+    render(
+      <ToastProvider>
+        <MockedProvider mocks={[]} addTypename={false}>
+          <PhaseSelector />
+        </MockedProvider>
+      </ToastProvider>
+    );
```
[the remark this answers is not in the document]
nunesmlucas: Created the  TestProvider, but im sure you wont like where its placed lol. Let me know if you have a specific location you prefer.

client/src/components/application/phases/ConceptPhase.tsx
```diff
@@ -1,5 +1,325 @@
...
+    completeConceptPhase(input: $input) {
+      id
+      success
+    }
+  }
+`;
+
+type Props = {
+  bundleId?: string;
+  demonstrationId?: string;
```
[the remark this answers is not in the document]
nunesmlucas: you're right.

client/src/test-utils/TestProvider.tsx
```diff
@@ -0,0 +1,43 @@
...
+ * ```
+ */
+export const TestProvider: React.FC<TestProviderProps> = ({
+  children,
+  mocks = [],
+  addTypename = false,
+}) => {
+  return (
+    <ToastProvider>
+      <MockedProvider mocks={mocks} addTypename={addTypename}>
```
tresdonicf, line 38: We can use `<DemosApolloProvider>` to automatically use the `<MockedProvider>` which could be nice

client/src/components/application/phases/ConceptPhase.tsx
```diff
@@ -1,5 +1,325 @@
...
+          disabled={!isFinishEnabled}
+          size="small"
+        >
+          {finishing ? "Saving…" : "Finish"}
+        </Button>
+      </div>
+    </div>
+  );
+
+  return (
```
tresdonicf, line 300: Thank you for breaking this out, it looks great!

tresdonicf (approved):

PR #212, opened 2025-09-08
Demos 567 implement person user distinction

connor-parke-icf:
This PR covers the breaking apart of Users into a User object and Person object. The Person object will be used for contacts in future implementations, so that not every Contact is a User. The goal with this particular update was to make the database changes without effecting the output of the existing resolvers too much. As a result, the Person object is not currently an object in graphql, but rather the User resolvers fetch the Person data and combine with the user.  

Authentication had to be updated to match, and the seeder was updated (a User requires that a Person exists before it can be created).  A migration was created to cover these changes; it is expected that at some point this migration will be condensed into the main migration for simplification. 

This does not cover all of the changes required for contacts, as such some parts are kind of bandaged together. For instance, at the end of this we will be removing both the demonstration.users and demonstration.projectOfficer; however that is not yet included in this PR.

### DIFF
--- a/data/utilities/scripts/history_trigger_generator.py
+++ b/data/utilities/scripts/history_trigger_generator.py
@@ -11,6 +11,7 @@ TBL_FOLDERS = [
     "_userState",
     "_userStateDemonstration",
     "bundle",
+    "bundlePhaseStatus",
     "demonstration",
     "demonstrationStatus",
     "document",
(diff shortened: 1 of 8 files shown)


server/src/model/person/person.prisma
```diff
@@ -0,0 +1,16 @@
...
+  createdAt    DateTime @default(now()) @map("created_at") @db.Timestamptz()
+  updatedAt    DateTime @updatedAt @map("updated_at") @db.Timestamptz()
+
+  personType PersonType @relation(fields: [personTypeId], references: [id])
+  user       User?
+
+  @@id([id])
+  @@unique([id, personTypeId])
+  @@map("person")
+}
```
connor-parke-icf, line 16: person does not have any graphql schema or resolvers; just exists in prisma for now.

james-becwar-cms (approved):

PR #217, opened 2025-09-09
task/fix-up-missing-unique-key

TARehman:
Fix up migration for a missing unique key.


connor-parke-icf (approved):
discussed; change made it into codebase but left out of migration mistakenly. this corrects that drift.

PR #221, opened 2025-09-09
DEMOS-753-add-views-and-basic-roles-to-migrations

TARehman:
Moves the previous documented groups and views into a proper migration. This also updates the views to be more properly aligned to modern PostgreSQL by not using the deprecated pg_group view and instead just using roles.


connor-parke-icf (approved):
mostly views and setting up permissions boilerplate so they can used later.

PR #244, opened 2025-09-15
Read success field from CreateDemonstrationResult

tresdonicf:
In order to better reflect the result from the `CreateDemonstrationResult`, read the message and surface the error in the toast.


james-becwar-cms (approved):

connor-parke-icf (approved):
I think this is fine; however i think i want to open a discussion during office hours on how we want to handle the returns of our mutations in a general way, whether we should aim to wrap them in a return object, and what that return object should contain, and which mutators to apply this pattern to. I am still a fan of returning the mutated object back to the user, and maybe an error message if the query fails

PR #245, opened 2025-09-15
DEMOS-807: Update authorization lambda to return 401 errors

cms-jesse:
Throwing an error instead of returning a DENY policy will result in a 401 instead of a 403, which currently conflicts with the cloudfront configuration returning misleading responses

### DIFF
--- a/lambda_authorizer/index.mjs
+++ b/lambda_authorizer/index.mjs
@@ -19,7 +19,7 @@ export const handler = async (event) => {
 
   if (!token) {
     console.log("no token");
-    return generatePolicy("unknown", "Deny", event.methodArn, {});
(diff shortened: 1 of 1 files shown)


tresdonicf (approved):
🙌

dustbuster:
> @dustbuster This will add a check for the no roles.

I already have a check for this. that happens in auth.utils.

```Javascript
function verifyRole(role: string): void {
  const validRoles = PERSON_TYPES as readonly string[];
  if (!validRoles.includes(role)) {
    throw new GraphQLError(`Invalid user role: '${role}'`, {
      extensions: {
        code: "UNAUTHORIZED",
        http: { status: 403 },
      },
    });
  }
}

PR #249, opened 2025-09-16
DEMOS-783: Make TEST IDM-only (disable native Cognito) && allow native Cognito only in DEV & Ephemeral

sandeepb910:
I have removed native Cognito in Test to require IDM (SAML) only. 

DEV and ephemeral environments are kept native Cognito and IDM.

Please review and approve.


tresdonicf (commented):
Some of the changes look duplicative from #244 so maybe a `git pull origin main` is nice to simplify the diffs but they look good!

PR #251, opened 2025-09-16
Demos 724 implement demonstration role assignment 2 2

connor-parke-icf:
Finally i think this is good to review. Apologies in advance for how big this PR is. Unfortunately, updating the types and the backend for Users and Person and demonstrationRoleAssignments caused this PR to involve a lot of small changes to many files. 

This change involves the ability to assign People to Demonstrations via a DemonstrationRoleAssignment. There are a couple of quirks to this, like how all demonstrations are required to have exactly one Primary project officer at all times. this is handled in a Deferred Check Constraint, which can be found in a migration. The database schema is probably best explained using the mermaid diagram. 

A source of many of the changes was the distinction between Users and People. I realized that for most of the usages of a person or user, we did not actually care about their "user" aspects (username, cognitoId), and should expect those fields to not be present.  Becauswe of all this, i have updated the frontend to use a dedicated Person object, rather than user. This unfortunately meant that many files were changes with very small adjustments. 

The contacts Tab, table, and modals were updated to take on the DemonstrationRoleAssignment object instead of the temporary, flat "contacts" object.  That object has been removed. The Contacts update and create functionalility has not been implemented, thus the changes made to the modals were for the sake of getting Typescript to be happy. These will need to be adjusted with a follow up. 

On creation of a demonstration, it will need to be assigned a Project Officer from the same state as the demonstration. For demoability i have added a state id to the labels of the ProjectOfficer select of the Create Demonstration modal. This will likely need to be adjusted (perhaps automatically filter the person list by state upon state selection).

Create user mutations were removed from graphql, since we do not actually have a means of creating users at the moment (outside of the Authorization functions, which query prisma directly). As such many of the user queries were replaced with Person queries in the application, tests, and mocks.   Unused queries and hooks were removed.

Tests and mockdata were updated to reflect these changes. I believe at this point, all of our mock data is within the same mock schema. I want to follow up on this to organize the mocks such that they can referenced by some kind of name, rather than the numerical index.

apologies again for the size of this PR; i had initially tried to split this up into a frontend and backend section, but it was impossible to make the requisite server changes without the frontend changes while keeping the app functional during that transition. looking forward to any feedback so dont hold back!

### DIFF
--- a/server/src/model/person/personSchema.ts
+++ b/server/src/model/person/personSchema.ts
@@ -1,6 +1,5 @@
 import { gql } from "graphql-tag";
-import { PersonType, State } from "../../types";
-import { DemonstrationRoleAssignment } from "../demonstrationRoleAssignment/demonstrationRoleAssignmentSchema";
+import { PersonType, State, DemonstrationRoleAssignment } from "../../types";
 
 export const personSchema = gql`
   type Person {


client/src/mock-data/personMocks.ts
```diff
@@ -0,0 +1,76 @@
...
+    personType: "demos-cms-user",
+    displayName: "cara.lee",
+    email: "cara.lee@email.com",
+  },
+  {
+    id: "10",
+    fullName: "David Chen",
+    personType: "demos-cms-user",
+    displayName: "david.chen",
+    email: "david.chen@email.com",
```
dustbuster, line 74: Golden opportunity. 🥇 Also Guids?

https://github.com/dustbuster/starwars_users.json/blob/main/star_wars_users.json

client/src/mock-data/userMocks.ts
```diff
@@ -1,99 +1,40 @@
...
+  { id: "1", username: "john.doe", person: mockPeople[0] },
+  { id: "2", username: "jane.smith", person: mockPeople[1] },
+  { id: "3", username: "jim.smith", person: mockPeople[2] },
+  { id: "4", username: "darth.smith", person: mockPeople[3] },
+  { id: "5", username: "bob.johnson", person: mockPeople[4] },
+  { id: "6", username: "alice.brown", person: mockPeople[5] },
+  { id: "7", username: "carlos.rivera", person: mockPeople[6] },
+  { id: "8", username: "emily.clark", person: mockPeople[7] },
+  { id: "9", username: "cara.lee", person: mockPeople[8] },
+  { id: "10", username: "david.chen", person: mockPeople[9] },
```
dustbuster, line 20: samsies. ^^

https://github.com/dustbuster/starwars_users.json/blob/main/star_wars_users.json

server/src/model/demonstration/demonstrationResolvers.ts
```diff
@@ -47,32 +46,56 @@ export async function createDemonstration(
...
+      if (!person) {
+        throw new Error(`Person with id ${input.projectOfficerUserId} not found`);
+      }
+
+      await tx.demonstrationRoleAssignment.create({
+        data: {
+          demonstrationId: bundle.id,
+          personId: input.projectOfficerUserId,
+          personTypeId: person.personTypeId,
+          roleId: "Project Officer",
```
TARehman, line 66: I'd make this a variable and give it the type at the top, so that we get type enforcement. Same with grant level below?

server/src/model/demonstration/demonstrationResolvers.ts
```diff
@@ -113,10 +135,6 @@ export const demonstrationResolvers = {
       });
     },
 
-    projectOfficer: async (parent: Demonstration) => {
```
TARehman, line 116: Did we want to not return the project officer at all anymore?

server/src/model/demonstration/demonstrationResolvers.ts
TARehman, line 1: I think we need to add the updating of the primary project officer to the update mutator for a demonstration. In turn, this might mean we should consider encapsulating that inside a function - however, since it has to take place in a transaction, and the creation also takes place inside a transaction, I'm not sure how tractable that is.

server/src/model/user/userSchema.ts
```diff
@@ -1,68 +1,32 @@
...
-  type Mutation {
-    createUser(input: CreateUserInput!): User
-    updateUser(id: ID!, input: UpdateUserInput!): User
-    deleteUser(id: ID!): User
+    createdAt: DateTime!
+    updatedAt: DateTime!
   }
 
   type Query {
-    users: [User!]!
```
TARehman, line 47: Do we not need to be able to enumerate users?

server/src/model/primaryDemonstrationRoleAssignment/primaryDemonstrationRoleAssignment.prisma
```diff
@@ -0,0 +1,11 @@
+model PrimaryDemonstrationRoleAssignment {
+  personId        String @map("person_id") @db.Uuid
+  demonstrationId String @map("demonstration_id") @db.Uuid
+  roleId          String @map("role_id")
+
+  demonstrationRoleAssignment DemonstrationRoleAssignment @relation(fields: [personId, demonstrationId, roleId], references: [personId, demonstrationId, roleId])
+
+  @@id([demonstrationId, roleId])
+  @@unique([personId, demonstrationId, roleId])
```
TARehman, line 9: Add the comment here explaining why this should be?

server/src/model/person/personSchema.ts
```diff
@@ -0,0 +1,33 @@
+import { gql } from "graphql-tag";
+import { PersonType, State } from "../../types";
+import { DemonstrationRoleAssignment } from "../demonstrationRoleAssignment/demonstrationRoleAssignmentSchema";
```
TARehman, line 3: Why not import this from types as well?

server/src/model/migrations/20250911195725_demonstration_role_assignment/migration.sql
TARehman, line 1: We should split the migrations into ones starting with `prisma_` which indicates that they're just autogenerated by Prisma, and ones that don't have the prefix, which would contain the custom logics that we wrote.

TARehman (commented):
I left some comments. A few larger notes:
 - I think you should clean up / collapse the migrations into a smaller set of files.
 - I think you didn't include the function to automatically add every CMS user to every state; I can't recall but I thought that was in scope for this ticket? Or is that in 725?
 - We should figure out if we need to be able to update the project officer through the existing updateDemonstration mutator, since we removed that functionality.

server/src/model/demonstration/demonstration.prisma
```diff
@@ -1,28 +1,27 @@
...
+  description           String?
+  effectiveDate         DateTime? @map("effective_date") @db.Date
+  expirationDate        DateTime? @map("expiration_date") @db.Date
+  cmcsDivisionId        String?   @map("cmcs_division_id")
+  signatureLevelId      String?   @map("signature_level_id")
+  statusId              String    @map("status_id")
+  stateId               String    @map("state_id")
+  currentPhaseId        String    @map("current_phase_id")
+  createdAt             DateTime  @default(now()) @map("created_at") @db.Timestamptz()
+  updatedAt             DateTime  @updatedAt @map("updated_at") @db.Timestamptz()
```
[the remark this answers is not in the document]
connor-parke-icf: nice catch, reformatted!

server/src/model/migrations/20250922171655_implement_demonstration_role_assignment/migration.sql
connor-parke-icf: compressed the migrations into one, then split it up into prisma and non-prisma changes

server/src/model/demonstration/demonstrationResolvers.ts
connor-parke-icf: the projectOfficer is now returned via the roles key, and filtering by project officer

server/src/model/primaryDemonstrationRoleAssignment/primaryDemonstrationRoleAssignment.prisma
connor-parke-icf: added!

server/src/model/person/personSchema.ts
connor-parke-icf: updated to pull from types

nunesmlucas (approved):

PR #254, opened 2025-09-16
DEMOS-752-add-bundle-phase-date-resolvers

TARehman:
This is my first attempt at a way to create and update the dates on phases. This PR includes the following.

 - Some minor changes to the README regarding naming mutators. @connor-parke-icf, we should take a pass over the whole thing at some point to reflect some of our other patterns that have started to emerge.
 - Adds a very barebones attempt to throw slightly more useful errors from Prisma that we can use elsehwere.
 - Adds an exported function in `bundleResolvers.ts` that is used to get a bundle; this isn't an actual resolver, but I thought the best place to put the function was in that folder. I don't know if we want to revisit the naming pattern.
 - Takes the existing GQL Bundle type / schema stuff and moves it under `bundleSchema.ts` and out of `documentSchema.ts`. Since it is now used more generally, this seemed like the right place to put it. I also moved some stuff from `documentSchema.ts` into `documentTypeSchema.ts` for the same reasons.
 - Implements resolvers for `bundlePhase` and everything for `bundlePhaseDate`. Explicitly makes the model `bundle -> bundlePhase -> bundlePhaseDate` in terms of the way the foreign keys, etc are implemented.
 - Exports functions from `demonstrationResolvers.ts` and `modificationResolvers.ts` for getting demonstrations, amendments, and extensions. These are used both in those resolvers, and in the new additions to ensure that we return a Bundle when updating a date.


server/src/model/demonstration/demonstrationResolvers.ts
```diff
@@ -14,13 +14,15 @@ const amendmentBundleTypeId: BundleType = BUNDLE_TYPE.AMENDMENT;
...
+}
+
 export const demonstrationResolvers = {
   Query: {
-    demonstration: async (_: undefined, { id }: { id: string }) => {
-      return await prisma().demonstration.findUnique({
-        where: { id: id },
-      });
-    },
+    demonstration: getDemonstration,
```
connor-parke-icf, line 25: interesting, is the ID parameter implied here? i thought the Parent is what gets passed into these subresolvers

server/src/model/demonstration/demonstrationResolvers.ts
```diff
@@ -231,5 +233,13 @@ export const demonstrationResolvers = {
...
       return parent.currentPhaseId;
     },
+
+    phases: async (parent: Demonstration) => {
+      return await prisma().bundlePhase.findMany({
+        where: {
+          bundleId: parent.id,
+        },
+      });
+    },
```
connor-parke-icf, line 243: yeah a findManyPhases method on the Phase resolvers would be nice to use here, but outside the scope. Good tech-debt/cleanup item

server/src/model/demonstration/demonstrationSchema.ts
```diff
@@ -31,6 +31,7 @@ export const demonstrationSchema = gql`
     demonstrationStatus: DemonstrationStatus!
     state: State!
     currentPhase: Phase!
+    phases: [BundlePhase!]!
```
connor-parke-icf, line 34: complete aside, but i still think about how we could enumerate the scalars like SignatureLevel so that by definition they can only contain the identified values. Would make a good spike

connor-parke-icf (approved):
looks great; a couple of nitpicks, but nothing blocking

server/src/model/bundlePhase/bundlePhaseResolvers.ts
```diff
@@ -0,0 +1,21 @@
+import { BundlePhase } from "@prisma/client";
+import { prisma } from "../../prismaClient.js";
+
+export const bundlePhaseResolvers = {
```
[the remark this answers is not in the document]
TARehman: Yes. In general we see this kind of all over. Listing documents attached to a demonstration or to a modification has the same field resolver. We can start to centralize and import them as needed.

server/src/model/bundlePhaseDate/bundlePhaseDateSchema.ts
```diff
@@ -0,0 +1,22 @@
+import { gql } from "graphql-tag";
+import { DateType } from "../../types.js";
+
+export const bundlePhaseDateSchema = gql`
```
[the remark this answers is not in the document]
TARehman: Right. It needs a schema / resolver because of the need to have specific queries for it, but its not a full-fledged resolver.

server/src/model/demonstration/demonstrationResolvers.ts
```diff
@@ -14,13 +14,15 @@ const amendmentBundleTypeId: BundleType = BUNDLE_TYPE.AMENDMENT;
 const extensionBundleTypeId: BundleType = BUNDLE_TYPE.EXTENSION;
 const conceptPhaseId: Phase = "Concept";
 
+export async function getDemonstration(_: undefined, { id }: { id: string }) {
+  return await prisma().demonstration.findUnique({
+    where: { id: id },
+  });
+}
+
```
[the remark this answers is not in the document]
TARehman: Let's defer that decision until we have a use case that helps us determine what the proper approach should be.

server/src/model/demonstration/demonstrationResolvers.ts
TARehman: Every resolver gets four arguments: `parent`, `args`, `context`, and `info`. That's true no matter what kind of resolver it is. We defined the function with just the parent argument so anything else that it happened to receive would be thrown away - not that it would receive anything else, of course.

PR #255, opened 2025-09-17
DEMOS-811: exit cli with non-zero when scripts error

cms-jesse:
Updates the demosctl utility (which is used in the jenkins pipelines) so that the exit codes match the processes the commands run. Previously at the root level, commands were being called, but some exit codes were being ignored. As part of this change, I modified the code to return the exit code and only call process.exit at the top level rather than individually within the functions.

### DIFF
--- a/deployment/demosctl/commands/buildClient.test.ts
+++ b/deployment/demosctl/commands/buildClient.test.ts
@@ -87,10 +87,7 @@ describe("buildClient", () => {
     const rc = runCommand as jest.Mock;
     rc.mockResolvedValue(1);
 
-    //@ts-expect-error ignore invalid mock
-    jest.spyOn(process, "exit").mockImplementation(() => "exit");
-
-    await buildClient(mockStageName, true);
+    const exitCode = await buildClient(mockStageName, true);
     expect(rc).toHaveBeenCalled();
     expect(rc).toHaveBeenCalledWith(
       "deploy-core-no-execute",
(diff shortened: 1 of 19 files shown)


deployment/demosctl/commands/buildClient.test.ts
```diff
@@ -86,17 +86,14 @@ describe("buildClient", () => {
...
+    const exitCode = await buildClient(mockStageName, true);
     expect(rc).toHaveBeenCalled();
     expect(rc).toHaveBeenCalledWith(
       "deploy-core-no-execute",
       "npx",
       expect.arrayContaining([`stage=${mockStageName}`, `demos-${mockStageName}-core`])
     );
 
-    expect(process.exit).toHaveBeenCalled();
+    expect(exitCode).toBe(1);
```
tresdonicf, line 97: Just curious - were these changes things you just wanted to revisit or something a tool brought up to change?

deployment/demosctl/commands/fullDeploy.ts
```diff
@@ -17,16 +17,17 @@ export async function fullDeploy(environment: string) {
...
   const outputData = readOutputs("all-outputs.json");
-  addCognitoRedirect(
+  await addCognitoRedirect(
     getOutputValue(outputData, `demos-${environment}-core`, "cognitoAuthority").split("/").pop()!,
     getOutputValue(outputData, `demos-${environment}-core`, "cognitoClientId"),
     getOutputValue(outputData, `demos-${environment}-ui`, "CloudfrontURL")
   );
 
   console.log(`\n======\ncomplete deploy command succeeded\n======\n`);
+  return 0;
```
tresdonicf, line 32: It would be awesome if JS had some symbols to represent what this is - though it's funny that this is how it's done in C and the like!

tresdonicf: Though I guess with typescript we could declare something like this and use it as a return type and that helps

```
type ExitCode = 0 | 1  // | ... other signals we might want to use
```

PR #257, opened 2025-09-17
DEMOS-677: Handle Workflow Phase Dates on the UI

tresdonicf:
JIRA: DEMOS-677 - Handle start and completion for each phase of the Application Approval Process

**Note, while this ticket is just to handle the start and completion dates for the phases that are currently in development (the first 3). I kinda went above and beyond in doing some of the extra handling here because we need helper methods to deal with the dates more generally to get to this point so I figured I'd build those out before getting deep into the date work**

Adds a library of functions to work with phase dates on the backend as well as as simulation for the phase dates to help ground conversations around this topic and visualize expected behavior.

These will largely be utilized in the `integration` stories for the workflow phases.

I don't think the simulation is exactly where it should be (for instance with the completeness section) but for now I implemented the business rules that were in the ticket, and the infrastructure to modify / add / etc as needed!

https://github.com/user-attachments/assets/307c5e86-cb49-48d2-8e3c-4941d1440c1a


client/src/util/formatDate.test.tsx
```diff
@@ -1,65 +1,73 @@
...
-    expect(formatDate(endOfYear)).toMatch(/^\d{2}\/\d{2}\/\d{4}$/);
-    expect(formatDate(noon)).toMatch(/^\d{2}\/\d{2}\/\d{4}$/);
-    expect(formatDate(midnight)).toMatch(/^\d{2}\/\d{2}\/\d{4}$/);
-    expect(formatDate(singleDigitMonthDay)).toMatch(/^\d{2}\/\d{2}\/\d{4}$/);
+    expect(formatDate(testDate)).toBe("04/15/2023");
+    expect(formatDate(leapYearDate)).toBe("02/29/2024");
+    expect(formatDate(startOfYear)).toBe("01/01/2022");
+    expect(formatDate(endOfYear)).toBe("12/31/2022");
+    expect(formatDate(noon)).toBe("07/04/2023");
+    expect(formatDate(midnight)).toBe("07/04/2023");
```
tresdonicf, line 29: Made these tests a bit more concrete

client/package.json
```diff
@@ -62,7 +64,7 @@
     "test:serial": "vitest --mode test --run --no-file-parallelism",
     "test:debug": "vitest --mode test --run --inspect-brk --no-file-parallelism",
     "tsc": "tsc --noEmit --skipLibCheck",
-    "lint": "eslint src",
+    "lint": "eslint src; npx tsc --noEmit --skipLibCheck",
```
tresdonicf, line 67: `lint` really should run typechecks as well

server/src/model/bundlePhaseDate/bundlePhaseDateSchema.ts
```diff
@@ -9,11 +9,25 @@ export const bundlePhaseDateSchema = gql`
...
+  type SetPhaseDateInput {
+    bundleId: ID!
+    phase: Phase!
+    dateType: DateType!
+    dateValue: DateTime!
+  }
+
   type Mutation {
-    setPhaseDate(bundleId: ID!, phase: Phase!, dateType: DateType!, dateValue: DateTime!): Bundle
+    setPhaseDate(input: SetPhaseDateInput): Bundle
```
tresdonicf, line 20: hopefully this is cool, kinda just moved it to the pattern that we're using elsewhere so we can export a type for it and use it on the client

client/src/components/application/dates/PhaseDateSimulation.tsx
tresdonicf, line 1: The simulation code - not that pretty but not too bad for the complexity. Didn't think it's worth the time-sink to clean this up ATM but it's a nice baseline to start from!

client/src/components/application/dates/phaseDates.ts
```diff
@@ -0,0 +1,116 @@
...
+ * This file uses branded types to prevent mixing up date types in function calls:
+ * https://www.learningtypescript.com/articles/branded-types
+ */
+export type DateUTC = UTCDate & { readonly __utc: never };
+export type DateEST = TZDate & { readonly __est: never };
+export type StartOfDayEST = DateEST & { readonly __start: never };
+export type EndOfDayEST = DateEST & { readonly __end: never };
+
+export const getStartOfDayEST = (year: number, month: number, day: number): StartOfDayEST => {
+  return new TZDate(year, month, day, 0, 0, 0, 0, EST_TIMEZONE) as StartOfDayEST;
```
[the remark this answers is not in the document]
tresdonicf: Totally! I did initially want to allow this to take a `Date` but that was providing all sorts of weirdness since when we call date.getYear and the like it'll be cast into the locale timezone so had to settle for providing all the arguments directly, should be a handy little function though!

PR #259, opened 2025-09-18
DEMOS-762: Set up server logs for tracing of errors.

dustbuster:
This is a logger of our design and replaces console logs and adds prisma and GQL logging too. 
Replaces consle.log

Screenshot 2025-09-23 at 09 37 51


server/src/plugins/logging.plugin.ts
connor-parke-icf, line 1: huge! so this plugs into every request and just logs that the request happened and succeeded or erred? fantastic; would love to be able to audit stuff like this to check against where the frontend cache is refetching when it should be pulling from cache, or double querying or something, or if theres a hidden query failing for some reason that isnt getting flagged

server/src/prismaClient.ts
```diff
@@ -1,9 +1,37 @@
...
+const createExtendedClient = () => {
+  const client = new PrismaClient({
+    log: [
+      { level: "warn", emit: "event" },
+      { level: "error", emit: "event" },
+      { level: "query", emit: "event" },
+    ],
+  }).$extends(prismaRandom());
+
+  // Log slow queries (>500ms) without params to avoid PII
```
connor-parke-icf, line 16: oo, yeah very good point on the PII; i didnt think about that. We should definitely have a dedicated spike once we fully flesh out the logging to triple check we're not logging PII

server/src/server.ts
```diff
@@ -60,8 +62,19 @@ export const graphqlHandler = startServerAndCreateLambdaHandler(
...
+      setRequestContext({ requestId, correlationId });
+      if (claims?.sub) addToRequestContext({ userId: claims.sub });
+
       const gqlCtx = await buildLambdaContext(headersWithClaims);
 
+      // Enrich context with resolved user id
+      if (gqlCtx?.user?.id) addToRequestContext({ userId: gqlCtx.user.id });
+
+      log.debug("lambda.context.built");
+
```
connor-parke-icf, line 77: might be good to create a context generator or const or something that we use in both server entrypoints to avoid drift between our local and remote envs.

server/src/logger.ts
connor-parke-icf, line 1: eventually, i think we should aim to replace this with a logging library, so we can do more advanced stuff like daily logs, log rotations, etc. Not needed now of course, but little features like that may be useful (slack integration? _shivers_)

connor-parke-icf (approved):
very welcome change! A couple of comments but nothing blocking. Im curious though, does this log on the GraphQL requests or the Prisma functions or both? looks like a little of both, which is nice especially if it can give us a timeline of the entry/exitpoints of each layer

server/src/logger.ts
dustbuster: True that home boy!

PR #262, opened 2025-09-19
Update client package.json and Jenkins Pipeline

tresdonicf:
Updated this so that the right exit codes are returned to Jenkins - also did the following:

1. Alphabetized Scripts
2. Added `:ci` to the end of scripts used in ci to get some signal that changing these scripts will have downstream effects
3. Moved `tsc` before `eslint` as we care a bit more about the typechecking errors than the eslint errors


pipelines/Jenkinsfile.client
```diff
@@ -21,9 +21,9 @@ pipeline {
         dirCon('client', 'node') {
             sh '''
             npm ci
-            npm run lint
-            npx tsc --noEmit --skipLibCheck
```
connor-parke-icf, line 25: nice; so now the jenkins runs only commands that are defined in the package.json? i.e. the jenkins processes should be easily reproducible locally, minus the sonarqube stuff?

connor-parke-icf (approved):
dont have much to comment on; im not really familiar with all the jenkins/pipeline stuff, but i welcome the change in the package.json to make it easier to do a quick check of compilation

cms-jesse (approved):

PR #266, opened 2025-09-20
DEMOS-776 Add pg_cron to devcontainer

james-becwar-cms:
Adds pg_cron ( https://github.com/citusdata/pg_cron ) to the postgres devcontainer.  Also fixed a bug where the pg data folder wasn't set properly and added a root .env to the .gitignore just in case.

### DIFF
--- a/.devcontainer/docker-compose.yml
+++ b/.devcontainer/docker-compose.yml
@@ -15,11 +15,13 @@ services:
       - "4000:4000"
 
   db:
-    image: postgres:latest
+    build:
+      context: ./postgres
+      dockerfile: Dockerfile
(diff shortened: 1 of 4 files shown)


connor-parke-icf (approved):
Nothing sticks out to me as blocking! though, i wonder in general whether we want cron logic to live in the database and not as a standalone lambda or part of the server.  Not sure if theres a meaningful difference, or if one is more maintainable over the others. My first thought would be to have the cron handled in a lambda, like the file processor, separate from the other components of the stack. That being said, having the cron directly attached to the DB i imagine simplifies some aspects.

PR #276, opened 2025-09-23
DEMOS 684 Add Federal Comment Phase UI

JeffreyM-GA:
image

- Adds UI for federal comment phase
- Moves upload section to shared component
- Migration to add federal comment phase to DB
- Small seeder change to insert one FC phase demonstration to DB


client/src/components/application/phase-selector/PhaseSelector.tsx
```diff
@@ -72,20 +95,18 @@ const MOCK_PHASE_STATUS_LOOKUP: PhaseStatusLookup = {
...
-const MOCK_PHASE_DATE_LOOKUP: Partial<Record<PhaseName, Date>> = {
-  Concept: new Date(2024, 4, 20),
-  "State Application": new Date(2024, 4, 22),
-  Completeness: new Date(2024, 11, 31),
-};
-
 export type PhaseStatusLookup = Record<PhaseName, PhaseStatus>;
 interface PhaseSelectorProps {
-  initialPhase?: PhaseName;
+  demonstration: ApplicationWorkflowDemonstration;
```
tresdonicf, line 100: Thanks for getting this in place!

client/src/components/application/phase-selector/PhaseSelector.tsx
```diff
@@ -72,20 +95,18 @@ const MOCK_PHASE_STATUS_LOOKUP: PhaseStatusLookup = {
...
   phaseStatusLookup?: PhaseStatusLookup;
 }
 
 export const PhaseSelector = (props: PhaseSelectorProps) => {
-  const [selectedPhase, setSelectedPhase] = useState<PhaseName>(props.initialPhase ?? "Concept");
+  console.log(props.demonstration);
```
tresdonicf, line 105: Looks like a hanging console.log perhaps we can remove this

server/src/model/migrations/20250923131101_insert_phase_four/migration.sql
tresdonicf, line 1: I think this migration might be a bit too narrow - could we insert all of the phase names here?

server/src/seeder.ts
tresdonicf, line 1: If we're doing the insert with a migration do we need updates to the seeder script?

client/src/components/dialog/document/FederalCommentUploadDialog.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+const DOCUMENT_TYPE_SUBSET: DocumentType[] = ["General File"];
+
+const REFETCH_QUERIES = ["GetConceptDocuments", "GetDemonstrationDocuments"];
+
+type Props = {
+  isOpen: boolean;
+  onClose: () => void;
+  bundleId: string;
+  refetchQueries?: string[];
```
tresdonicf, line 14: Let's not allow the user to pass in `refetchQueries`, we'll just control the list of queries that need to be refetched

client/src/components/dialog/document/FederalCommentUploadDialog.tsx
```diff
@@ -0,0 +1,42 @@
...
+  bundleId,
+  refetchQueries = REFETCH_QUERIES,
+}) => {
+  const getInitialDocument = (): DocumentDialogFields => ({
+    id: bundleId,
+    title: "",
+    description: "",
+    documentType: "General File",
+    file: null,
+  });
```
tresdonicf, line 29: We shouldn't need an initial document if we're using the `<AddDocumentDialog>`, only for the edit

client/src/components/application/phases/ConceptPhase.tsx
```diff
@@ -310,7 +245,19 @@ export const ConceptPhase: React.FC<Props> = ({
       <section className={STYLES.pane}>
         <div className={STYLES.grid}>
           <span aria-hidden className={STYLES.divider} />
-          <UploadSection />
+          <ApplicationUploadSection
```
tresdonicf, line 248: Cool to break this out!

client/src/components/application/phases/sections/sectionStyles.ts
tresdonicf, line 1: NBD: I get the feeling we might  want to just declare these as different `STYLES` objects where they're used so it's clearer where they're used

client/src/components/application/phases/FederalCommentPhase.tsx
```diff
@@ -1,5 +1,129 @@
...
+  ApplicationUploadSection,
+  STYLES,
+} from "components/application/phases/sections";
+import { ExitIcon, WarningIcon } from "components/icons";
+import { formatDate } from "util/formatDate";
+import { DocumentTableDocument } from "components/table/tables/DocumentTable";
+import { FederalCommentUploadDialog } from "components/dialog/document/FederalCommentUploadDialog";
+
+interface FederalCommentPhaseProps {
+  demonstrationId?: string;
```
tresdonicf, line 14: I could be wrong I don't think we want the demonstrationId to be optional

tresdonicf (commented):
Great job on this, it's looking good! I had a few comments, many of them style / NBD and a couple that are maybe more salient regarding:

1. Migrations, maybe we can insert all the values in the migration (maybe they're already in there too, this sort of complexity is why I think just sharing a string[] is the simplest way forward)
2. I think a few of the props should be required, removed, etc (more details in comments)

Overall really nice work on this!

client/src/components/dialog/document/FederalCommentUploadDialog.tsx
JeffreyM-GA: This sets the default value for documentType, which was required. Is there a more straightforward way to do it?

client/src/components/application/phases/sections/sectionStyles.ts
JeffreyM-GA: I do actually agree with this, the pipeline can be really picky about duplicated code though. Not a fan of making compromises so the build will pass, I can try to keep them separate and see if it complains.

tresdonicf: I see, yeah if it's a problem due to the way the scanner works then we can  keep it and make a ticket to look into it later!

server/src/seeder.ts
JeffreyM-GA: so, the updates to the seeder script were only for creating a test case in QA, since there's no actual way to advance a demonstration through the phases right now. I asked Tom and he preferred this to manually changing the values in the DB. Not really sure how important it is to have a valid test case just yet, but the ACs said something like "Must default to the correct phase" so I put one in

server/src/model/migrations/20250923131101_insert_phase_four/migration.sql
JeffreyM-GA: I had meant to check with Tom to see if there was any reason they had all not just been shoved in initially, let me reach out to him and see if we can just add them all now

PR #278, opened 2025-09-23
Demos 739 phase 3 - completeness!

dustbuster:
Phase 3, as wired up as i can make it without real dates or docs!

## Completeness Video
https://github.com/user-attachments/assets/f8277bcc-34a4-4444-8c5e-9865f5783a42

Screenshot 2025-09-25 at 16 54 36

Screenshot 2025-09-25 at 16 54 45


client/src/components/application/phases/CompletenessPhase.tsx
```diff
@@ -1,5 +1,325 @@
...
+  const [collapsed, setCollapsed] = useState(false);
+  const [isNoticeDismissed, setNoticeDismissed] = useState(false);
+  const [mockDocuments, setMockDocuments] = useState<DocumentTableDocument[]>([]);
+  const [stateDeemedComplete, setStateDeemedComplete] = useState<string>("");
+  const [federalStartDate, setFederalStartDate] = useState<string>("");
+  const [federalEndDate, setFederalEndDate] = useState<string>("");
+
+  const completenessDocs = mockDocuments.filter(
+    (doc) => doc.documentType === "Application Completeness Letter"
+  );
```
dustbuster, line 62: Easy wire in once we get the real doc in here.

client/src/components/application/phases/CompletenessPhase.tsx
```diff
@@ -1,5 +1,325 @@
...
+      {/* Since this is pretty handy (pre-api hookup), let's make it easy to separate */}
+      {isLocalDevelopment() && (
+        <CompletenessTestingPanel
+          onAddMockDoc={addMockDoc}
+          completenessDocCount={completenessDocs.length}
+          noticeDueDate={noticeDueDate}
+          onNoticeDueDateChange={setNoticeDueDate}
+          noticeDaysValue={noticeDaysValue} // keep for display-only if you show it
+          onResetNotice={() => setNoticeDismissed(false)}
+        />
```
dustbuster, line 321: basically this lil control panels tells us what we need to hydrate!

client/src/components/notice/Notice.tsx
tresdonicf, line 1: Awesome stuff!

client/src/components/dialog/DeclareIncompleteDialog.tsx
tresdonicf, line 1: I kind of have a hunch that this should be bundled with the component it's used since it's not general enough to be re-used across different parts of the application

Looks great though!

client/src/components/dialog/document/CompletenessUploadDialog.tsx
```diff
@@ -0,0 +1,45 @@
...
+const DOCUMENT_TYPE_SUBSET: DocumentType[] = ["Application Completeness Letter", "General File"];
+
+const REFETCH_QUERIES = ["GetCompletenessDocuments", "GetDemonstrationDocuments"];
+
+// This needs to be a toast. But more permanent. - not sure if this is the same.
+type Props = {
+  isOpen: boolean;
+  onClose: () => void;
+  bundleId?: string; // optional starter prop; can be wired later
+  refetchQueries?: string[];
```
tresdonicf, line 16: Let's maintain the list of queries to refetch rather than have the user pass them in

client/src/components/dialog/document/CompletenessUploadDialog.tsx
tresdonicf, line 1: Same thing with this guy, starting to notice we might want to put some of these just right next to where they're used since it will only likely be used in one place

client/src/components/application/phase-selector/PhaseDate.tsx
```diff
@@ -26,13 +28,24 @@ const PHASE_TEXT_COLOR_LOOKUP: Record<PhaseStatus, string> = {
   completed: tw`text-success-darkest`,
 };
 
-export const PhaseDate = ({ phaseStatus, date }: { phaseStatus: PhaseStatus; date?: Date }) => {
+interface PhaseDateProps {
+  phaseStatus: PhaseStatus;
+  date?: Date;
+  isPastDue?: boolean;
```
tresdonicf, line 34: For the purposes of this component, which is just for rendering the phase dates I think this should just be a new `PhaseStatus`  like `past-due` (or any name like that)

Then that status is calculated before passing it into this component

tresdonicf: This means we can also re-use the existing lookups rather than handle it as a separate edge case which it isn't quite

client/src/components/application/phase-selector/PhaseBox.tsx
```diff
@@ -30,6 +30,7 @@ interface PhaseBoxProps {
   phaseNumber: number;
   phaseStatus: PhaseStatus;
   displayDate?: Date;
+  isPastDue?: boolean;
   isSelectedPhase: boolean;
```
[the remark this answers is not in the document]
tresdonicf: I think the way forward here is adding this to our `PhaseStatus` type. We could do it in the server and add a status for past-due or we could just add it on the frontend here:

`type PhaseStatus = ServerPhaseStatus | "past-due"`. If I remember right I think that there might be a need to represent past due on the server anyway for some type of requirement thing but I could be wrong, might be worth a conversation with the team

client/src/components/application/phases/CompletenessPhase.tsx
```diff
@@ -1,5 +1,325 @@
...
+  const dd = String(d.getDate()).padStart(2, "0");
+  return `${yyyy}-${mm}-${dd}`;
+};
+// parser to keep the dates a string
+const parseDateInput = (input?: string) => {
+  if (!input) return undefined;
+  const [year, month, day] = input.split("-").map((value) => Number(value));
+  if (!year || !month || !day) return undefined;
+  return new Date(year, month - 1, day);
+};
```
tresdonicf: Probably can use stuff from date-fns for these

client/src/components/application/phase-selector/PhaseSelector.tsx
tresdonicf, line 28: I think this is the same as the PhaseNames in the server now, is that right? Oh except maybe `None` - if so perhaps we can use that now that it's ready

client/src/components/application/phase-selector/PhaseSelector.tsx
```diff
@@ -69,63 +71,152 @@ interface PhaseSelectorProps {
...
+  }, [demonstration.id]);
 
   const DisplayPhase = ({ selectedPhase }: { selectedPhase: PhaseName }) => {
-    const PhaseComponent = PHASE_COMPONENTS_LOOKUP[selectedPhase];
+    const PhaseComponent = phaseComponentsLookup[selectedPhase as PhaseSelectorPhase];
 
     return (
       <div className="w-full h-full min-h-64">
-        <PhaseComponent />
+        {PhaseComponent ? <PhaseComponent /> : null}
```
tresdonicf, line 193: PhaseComponent should always exist here as we've made the lookup and are using the values from it.

tresdonicf (dismissed):
Overall this is looking good! The one change that I think fragments the system a bit too  much is the handling of the `pastDue` status separately from the other `PhaseStatus`es. I think some of the tooling for `PhaseMeta` and stuff are built out of that but we really don't need anything like that as the whole `phases` model in the dataflow is pretty unstructured so we can put any dates we need into there.

So if we need to capture the due date for a phase we should make sure it exists in `DATE_TYPES` ("Completeness Review Due Date" is already there) and interact with that date if it exists, and setting the `PhaseStatus` based on that

client/src/components/application/phases/CompletenessDocumentUploadDialog.tsx
dustbuster: Done.

dustbuster:
okay. This has some useful stuff that might come in handy for the wiring. Will close and revisit if needed.

PR #283, opened 2025-09-24
DEMOS-746: Landing Page Visual Audit

tresdonicf:


client/src/pages/index.ts
```diff
@@ -1 +1 @@
-export { LandingPage }  from "./LandingPage";
```
dustbuster, line 1: 🤯 DOES NOT COMPUTE, DOES NOT COMPUTE!

client/src/components/footer/Footer.tsx
```diff
@@ -2,7 +2,7 @@ import React from "react";
 
 export const Footer: React.FC = () => {
   return (
-    <footer className="bg-white text-black fixed bottom-0 w-full border-t">
+    <footer className="bg-white text-black fixed bottom-0 w-full border-t z-12">
```
connor-parke-icf, line 5: curious if we want to start a convention for our z-indexes to treat almost like hotel room numbers, something like: 
000 - main content
100 - sidenav
200 - footer
300 - header
400 - modals
etc

client/src/components/header/DefaultHeaderLower.tsx
```diff
@@ -53,7 +53,7 @@ export const DefaultHeaderLower: React.FC = () => {
   };
 
   return (
-    <div className="w-full bg-brand text-white px-4 py-1 flex items-center justify-between">
```
connor-parke-icf, line 56: im not familiar enough with tailwind; i remember running into some issues back when first building out the header that out px-1,2,3,4 wasnt sufficient in describing the spacing needed for a number of places. Maybe its worth seeing if those are set right or are due for adjustment? Generally, i try to avoid numbers bigger than like 8 or so in css; it tends to point me towards using a different scale/system like rem, percents, calc stuff. Not sure if that instinct carries over to tailwind though; if this works it works!

client/src/components/header/QuickLinks.tsx
```diff
@@ -2,26 +2,25 @@ import { HelpIcon, NotifyIcon, SettingsIcon } from "components/icons";
...
+      <a href="#" className={linkStyles}>
+        <NotifyIcon className={iconStyles} />
+        <span>Notifications</span>
+      </a>
+
+      <a href="#" className={linkStyles}>
+        <HelpIcon className={iconStyles} />
+        <span>Help</span>
+      </a>
+    </div>
```
connor-parke-icf, line 24: This works! im not entirely sure why i initially made those as list items. I think it may have had to do with screenreaders or something, but im also guessing whatever principles i was thinking about are outdated or maybe never existed lol. Happy to see this simplified

client/src/components/table/tables/DemonstrationTable.tsx
```diff
@@ -10,7 +10,7 @@ import {
   Demonstration,
   DemonstrationAmendment,
   DemonstrationExtension,
-} from "pages/Demonstrations";
+} from "pages/DemonstrationsPage";
```
connor-parke-icf, line 13: happy to see this renamed, "Demonstrations" was too broad. Been thinking about how we're naming the other page as well. Floating the idea of: DemonstrationListingPage and DemonstrationDetailPage (or maybe just DemonstrationPage). the awkwardness comes from having a DemonstrationDetailTab within the DemonstrationDetail page, so having the page itself mean "demonstration" that tab mean "details" and the other page mean "demonstration list" makes sense in my mind.

client/src/components/table/Table.tsx
```diff
@@ -128,86 +195,21 @@ export function Table<T>({
...
-          {columnFilter && columnFilter(table)}
-        </div>
-        <div className="mr-1">
-          {/* Action Buttons Section */}
-          {actionButtons && actionButtons(table)}
-        </div>
+      <div className="flex items-center justify-between mb-[24px]">
+        <TableSearch table={table} keywordSearch={keywordSearch} columnFilter={columnFilter} />
+        <div className="mr-1">{actionButtons && actionButtons(table)}</div>
       </div>
```
connor-parke-icf, line 201: very very happy to see this stuff revised. This component has been bothering me for a while. its very awkward to work around and within, so happy to see this cleaned up

client/src/layout/SideNav.test.tsx
```diff
@@ -1,30 +1,37 @@
 import React from "react";
 import { render, screen, fireEvent } from "@testing-library/react";
 import { MemoryRouter } from "react-router-dom";
-import { vi, describe, it, expect } from "vitest";
+import { describe, it, expect } from "vitest";
 import { SideNav } from "./SideNav";
 
+const renderSideNav = () => {
+  return render(<SideNav />, { wrapper: MemoryRouter });
```
connor-parke-icf, line 8: dont think ive seen this wrapper convention before; i assume its the equivalent of 
render(<MemoryRouter><SideNav /></MemoryRouter>)?
if so thats huge, that big pyramid in the main router might can get squashed down

client/src/router/DemosRouter.test.tsx
```diff
@@ -80,23 +79,23 @@ vi.mock("layout/PrimaryLayout", () => ({
...
 }));
-vi.mock("pages/Demonstrations", () => ({
-  Demonstrations: () => <div>Demonstrations</div>,
+vi.mock("pages/DemonstrationsPage", () => ({
+  DemonstrationsPage: () => <div>DemonstrationsPage</div>,
   DEMONSTRATIONS_PAGE_QUERY: {},
 }));
 
 describe("DemosRouter", () => {
-  it("renders the LandingPage at root path", () => {
```
connor-parke-icf, line 89: Im a fan of this, describing the page on whats on it and what it does rather than how a user flows through the application

connor-parke-icf (approved):
nice work! a couple conversation points and general nitpick comments, but nothing blocking. Love a good cleanup!

client/src/components/footer/Footer.tsx
tresdonicf: I wonder too! I don't know if it should crop up too much but if anything it would be nice to codify it outside of this in maybe it's own `z-index.css` file where we can define these there and just reuse them here like `z-modal`, `z-sidenav`, etc.

client/src/layout/SideNav.test.tsx
tresdonicf: Honestly I had to look this one up as well, it seems like it's just for testing so this could have been also `return render(<MemoryRouter><SideNav /></MemoryRouter>)` and it seems like it would be equivalent? Probably just to be able to hot-swap the wrapper components a bit easier?  https://testing-library.com/docs/react-testing-library/api/#wrapper

PR #284, opened 2025-09-25
Demos 419 add contacts for a demonstration

connor-parke-icf:
Though the conversation around this is still somewhat in flux, this is implemented under the assumption that assigning a new demonstrationRoleAssignment to a user that is flagged as primary for a role that has an existing primary entry will not remove the old entry from the demonstration entirely, and instead will set it to be the non-primary entry. If we end up wanting to remove the old entry entirely, it would be a purely additive change so i think we're good to review this now but ill hold off on merging it until we're settled. 

Adds two resolvers, setDemonstrationRole and unsetDemonstration role.  Set demonstrationRole's purpose is to create a role entry which matches the input variables, including the isPrimary field. If another primary exists, it will be demoted to make space for the new entry. One can set a primary role to be non-primary if they choose, however attempting to do so to a primary project officer (when not setting a new primary project officer to replace it) will cause an error.  This will be used for both Creating and Editing a contact. 

UnsetDemonstrationRole is simpler, in that it will remove the role entry from the database, first removing its primary assignment if it exists. This will be used for deleting a contact.


client/src/components/table/tables/ContactsTable.tsx
```diff
@@ -192,15 +189,18 @@ export const ContactsTable: React.FC<ContactsTableProps> = ({ roles = [], onUpda
...
       )}
 
       {/* Remove Contact Dialog */}
       <RemoveContactDialog
         isOpen={dialogMode === "remove"}
         onClose={handleCloseDialog}
-        contactIds={selectedContactIds}
+        contacts={selectedContacts.map((contact) => ({
+          ...contact,
+          demonstration: { id: demonstrationId },
```
connor-parke-icf, line 202: not a fan of having the demonstrationId be separate here then conjoined in. Ultimately, i think i want the "contacts" to include the demonstrationId within it. that way all of the submit handlers dont need to care about what page theyre on, they can just deal with the contact objects themselves. Of course, that leads to a Graphql query like:
demonstration -> roles -> demonstration,
which is perfectly acceptable, but i dont think our mock data is equipped for it (issues with circular dependencies). It is starting to make me rethink our mocked data (for the 100th time), so if we're looking to make the jump away from the dev:mocks, i could see this issue arguing in favor of that; mock data shouldnt get in the way of core code decisions.

client/src/components/table/tables/ContactsTableDemo.tsx
```diff
@@ -2,32 +2,32 @@ import React, { useState } from "react";
...
-      )
-    );
+  //   // Update the contact in state
+  //   setRoles((prevContacts) =>
+  //     prevContacts.map((contact) =>
+  //       contact.person === role.person && contact.isPrimary === role.isPrimary
+  //         ? { ...contact, contactType: contactType as Role }
+  //         : contact
+  //     )
+  //   );
```
connor-parke-icf, line 24: wasnt really sure how to handle this; it is a symptom of pulling the onSubmit handler into the modal rather than the ContactsTable. Conventional react seems to point to keeping the handler as passed in props, but graphql seems to point to keeping the callee component responsible for handling its data.  Given that this is the only place we mutate contacts, and the edit contacts modal's main job is to edit contacts, im opting towards getting that logic out of the parent component.

client/src/components/table/tables/ContactsTableDemo.tsx
```diff
@@ -2,32 +2,32 @@ import React, { useState } from "react";
...
 
-    console.log(`Updated contact of ${role.person.fullName} to type: ${contactType}`);
-  };
+  //   console.log(`Updated contact of ${role.person.fullName} to type: ${contactType}`);
+  // };
 
   const resetContacts = () => {
-    setRoles(demonstrationRoleAssignmentMocks);
+    setRoles(mockDemonstrationRoleAssignments);
```
connor-parke-icf, line 30: flipped the naming convention of this to match the other ones. Getting tedious to manage these.

server/src/model/demonstration/demonstrationResolvers.ts
```diff
@@ -176,15 +176,11 @@ export const demonstrationResolvers = {
...
 
     roles: async (parent: Demonstration) => {
-      const roleAssignments = await prisma().demonstrationRoleAssignment.findMany({
+      return await prisma().demonstrationRoleAssignment.findMany({
         where: { demonstrationId: parent.id },
-        include: { primaryDemonstrationRoleAssignment: true },
       });
-      return roleAssignments.map((assignment) => ({
-        ...assignment,
-        isPrimary: !!assignment.primaryDemonstrationRoleAssignment,
```
connor-parke-icf, line 185: moved the isPrimary mapping to the demonstrationRoleAssignment resolvers

server/src/model/demonstrationRoleAssignment/demonstrationRoleAssignmentResolvers.ts
```diff
@@ -1,8 +1,130 @@
...
+const DEMONSTRATION_GRANT_LEVEL = "Demonstration";
+
+export async function unsetDemonstrationRoles(
+  parent: undefined,
+  { input }: { input: UnsetDemonstrationRoleInput[] }
+) {
+  return await prisma().$transaction(async (tx) => {
+    const deletedRoles: DemonstrationRoleAssignment[] = [];
+
+    for (const roleInput of input) {
```
connor-parke-icf, line 18: made this delete many roles to reflect the frontend's usage

server/src/model/graphql.ts
```diff
@@ -47,35 +45,6 @@ import { demonstrationRoleAssigmentResolvers } from "./demonstrationRoleAssignme
 
 const scalarTypes = [JSONObjectDefinition, DateTimeTypeDefinition, DateTypeDefinition];
 
-const mockDemonstrationSchemaExtension = gql`
```
connor-parke-icf, line 50: no longer need to mock the old contacts serverside

connor-parke-icf (commented):
update: got the frontend part of this working to a functional degree.  There is a good amount of things i want to revisit but largely this is the quick-and-dirty version while we await a more solid UI design for handling contacts.

client/src/components/dialog/RemoveContactDialog.tsx
```diff
@@ -5,30 +5,67 @@ import { ErrorButton } from "components/button/ErrorButton";
...
     try {
       setIsDeleting(true);
 
-      // TODO: Add actual API call here when available
-      console.log("Deleting contacts:", contactIds);
+      const mutationInput = contacts.map((contact) => ({
+        personId: contact.person.id,
+        demonstrationId: contact.demonstration.id,
+        roleId: contact.role,
+      }));
```
tresdonicf, line 59: Great to see! This is a place I find that providing a type annotation for the input is really helpful, this way if something changes in the expected input on the server we'll get an update that it's not all in sync through `gql linter --> exported type --> type usage here`

client/src/components/dialog/RemoveContactDialog.tsx
```diff
@@ -5,30 +5,67 @@ import { ErrorButton } from "components/button/ErrorButton";
...
+
+      await unsetDemonstrationRolesTrigger({ variables: { input: mutationInput } });
 
-      const isMultipleContacts = contactIds.length > 1;
+      const isMultipleContacts = contacts.length > 1;
       showSuccess(`Your contact${isMultipleContacts ? "s have" : " has"} been removed.`);
+      console.log(
+        "Deleting contacts:",
+        contacts.map((contact) => contact.person.id)
+      );
```
tresdonicf, line 68: Not sure if this is meant to stay in or just a remnant

client/src/components/table/tables/ContactsTableDemo.tsx
tresdonicf, line 1: Is this still in use? it appears to be a demo version of the contacts table but perhaps we can remove it if the real version is in place?

tresdonicf (approved):
A few comments but overall LGTM!

client/src/components/table/tables/ContactsTableDemo.tsx
connor-parke-icf: I dont think this is really in use to be honest. but it was very recently added and the current design isnt the "final" design, so im thinking ill leave it in until we have the true, full replacement.

PR #294, opened 2025-09-29
DEMOS-782: Refactor Demonstration Dialogs

tresdonicf:
In preparation for other changes related to DEMOS-782, here's a refactoring PR, primarily moves the Create / Edit demonstration modals to their own files along with their queries

### DIFF
--- a/server/src/model/demonstration/demonstrationSchema.ts
+++ b/server/src/model/demonstration/demonstrationSchema.ts
@@ -50,11 +50,13 @@ export const demonstrationSchema = gql`
   input UpdateDemonstrationInput {
     name: String
     description: String
-    stateId: ID
     effectiveDate: Date
     expirationDate: Date
     cmcsDivision: CmcsDivision
     signatureLevel: SignatureLevel
+    status: BundleStatus
+    currentPhase: Phase
+    stateId: ID
   }
 
   type CreateDemonstrationResponse {
@@ -108,9 +110,11 @@ export interface CreateDemonstrationInput {
 export interface UpdateDemonstrationInput {
   name?: string;
   description?: string;
-  stateId?: string;
   effectiveDate?: Date;
   expirationDate?: Date;
   cmcsDivision?: CmcsDivision;
   signatureLevel?: SignatureLevel;
+  status?: BundleStatus;
+  currentPhase?: Phase;
+  stateId?: string;
 }


client/src/queries/demonstrationQueries.ts
```diff
@@ -28,51 +28,3 @@ export const GET_ALL_DEMONSTRATIONS_QUERY = gql`
...
-      users {
-        id
-        fullName
-      }
-      projectOfficer {
-        fullName
-      }
-    }
-  }
-`;
```
tresdonicf, line 53: Moved these to the components where they're used 🥳

connor-parke-icf (commented):
im a fan of this change! Love the idea of just droppin in the modal component, giving it a demonstrationId and letting it figure out the rest. Keeping the query behavior in the caller components rather than the callee makes sense; the callee is shared by the two callers that have separate submit behaviors and data needs. 

In the back of my head, i keep having thoughts about a global context for modal management which may further simplify things; assuming that modals are somewhat singleton.  We do have a few instances where we have our exit warning modal pop up in addition to the current modal, so it may not be appropriate. That all being said, i think its something worth looking into. 

Nonetheless, nice work! not much to comment on

connor-parke-icf (approved):

tresdonicf:
> im a fan of this change! Love the idea of just droppin in the modal component, giving it a demonstrationId and letting it figure out the rest. Keeping the query behavior in the caller components rather than the callee makes sense; the callee is shared by the two callers that have separate submit behaviors and data needs.
> 
> In the back of my head, i keep having thoughts about a global context for modal management which may further simplify things; assuming that modals are somewhat singleton. We do have a few instances where we have our exit warning modal pop up in addition to the current modal, so it may not be appropriate. That all being said, i think its something worth looking into.
> 
> Nonetheless, nice work! not much to comment on

Thanks! Appreciate you continuing to bring up the idea of bringing these queries a bit closer to the components, it's definitely a solid change! I totally agree about the modal context thoughts - it would be really interesting to prototype some type of `<ModalManager>` context that allows us to summon any modal throughout the application. Especially now that we're using proper `<dialog>` elements that can close themselves this is the next logical step! Still thinking about this but currently I imagine it working kinda like toasts:

```
const { createDemonstrationModal, editDemonstrationModal} = useModal();
....
createDemonstrationModal.show()
editDemonstrationModal.show(demonstrationId)
```

Then these could be wrapped in `() => { createDemonstrationModal.show() }` for things like lazy execution in callbacks, that way we don't need every component that shows a modal to manage the state itself!

PR #295, opened 2025-09-29
DEMOS-844-phase-specific-documents

TARehman:
Closes DEMOS-844.

This PR adds the concept of a phase to the documents, which is necessary to properly associate a document to a phase when it is added there. It makes the following changes.

 - More broadly, this adds a way to quickly add custom scalar validation in Apollo, so we can check that input are valid against our defined types rather than having the database catch these (though we still have that as a backup). This is accomplished with a generic function I put into `resolverFunctions.ts`. I used this to add custom scalar validation for sets of strings to:
   - BundleStatus
   - CmcsDivision
   - DateType
   - DocumentType
   - Phase
   - PhaseStatus
   - SignatureLevel
 - Conforms the phase / document association to the existing data model by adding a `phase_document_type` association table and populating it with known data. Updates the Prisma files throughout to match this specification.
 - Updates the seeder to reflect the new data model.
 - Reorganizes `graphql.ts` to be in alphabetical order and use a fairly consistent approach to how imports / exports are done.
 - Updates the modification resolvers to return phases like demonstration does (I am not sure why this wasn't already being done).

I tested this by running locally and using the API via Apollo to create and update records that use the new feature.

There is an open issue on the front-end side, but I am out of my depth in handling it.

```
npm run build

> demos-client@0.1.0 build
> npm run tsc && vite build


> demos-client@0.1.0 tsc
> tsc --noEmit --skipLibCheck

src/components/dialog/document/DocumentDialog.tsx:456:11 - error TS2741: Property 'phase' is missing in type '{ bundleId: string; title: string; description: string; documentType: "Pre-Submission" | "Application Completeness Letter" | "Approval Letter" | "Final BN Worksheet" | "Final Budget Neutrality Formulation Workbook" | ... 6 more ... | "State Application"; }' but required in type 'UploadDocumentInput'.

456     const uploadDocumentInput: UploadDocumentInput = {
              ~~~~~~~~~~~~~~~~~~~

  ../server/src/model/document/documentSchema.ts:73:3
    73   phase: Phase;
         ~~~~~
    'phase' is declared here.


Found 1 error in src/components/dialog/document/DocumentDialog.tsx:456
```

Basically, you now need to give a phase to the back-end when you create a document. We likely should not expose this to the user, and should instead derive it from where the modal is called from (e.g. if you are calling it from the Concept phase, just set the phase argument to Concept). However, I'm having difficulty getting the front-end upload to work in `dev` or locally at all; I think something may not be working correctly. As it's not currently appearing to work, and I'm not sure what might be the cause, I wanted to get this open for discussion; however, if it makes sense to fold a fix into this PR I am happy to do so if someone points me in the right direction / helps me figure out what I should be doing there.

ETA: I've just made the build succeed on this by simply passing in "None" for every upload, but this is obviously wrong and needs to be corrected.


server/src/model/modification/modificationResolvers.ts
```diff
@@ -171,6 +171,14 @@ async function getCurrentPhase(parent: Modification) {
...
 }
 
+async function getPhases(parent: Modification) {
+  return await prisma().bundlePhase.findMany({
+    where: {
+      bundleId: parent.id,
+    },
+  });
+}
+
```
connor-parke-icf, line 181: in DocumentResolvers, we are returning the phase as an ID, but here it looks like we're returning an object. Worried that we're overloading the word "phase" if it can be both a thing and a type, unless im reading this wrong.

connor-parke-icf: update, i think i understand; this is the discussion we were having earlier. The only thing document needs is the reference to the phase, which could be "None". I think maybe changing the name of the phase id/type in the documentschema would be a good idea, to signify that this isnt a phase itself, but rather the phaseId.

server/src/resolverFunctions.ts
```diff
@@ -0,0 +1,44 @@
+import { GraphQLScalarType, Kind, StringValueNode } from "graphql";
+
+export function generateCustomSetScalar(
+  acceptableValues: readonly string[],
+  name: string,
+  description: string
+): GraphQLScalarType {
+  const scalarDescription =
+    description + " Expected values are:\n- " + acceptableValues.join("\n- ");
```
connor-parke-icf, line 9: ahh this answers my question further above. Nice! built-in tool-tip!

client/src/components/dialog/document/DocumentDialog.tsx
```diff
@@ -458,6 +458,7 @@ export const AddDocumentDialog: React.FC<{
       title: dialogFields.title,
       description: dialogFields.description,
       documentType: dialogFields.documentType,
+      phase: "None",
```
connor-parke-icf, line 461: good temp change until we get the document-phase relationship solidified ui-side

connor-parke-icf (approved):
server stuff looks good to me; most of my comments i answered myself by reading on. Im assuming all of the changes are already documented in the mermaid model.

PR #297, opened 2025-09-30
DEMOS-784: fixed Add amendments, Add extensions.

dustbuster:
I'll come back to this. 
WIth another ticket. 

Amendments and Extensions!
Refactored Amendment and Extensions to A) work. But B) use the same theory as CreateDemo. 

I have NOT made an `EditExtension` Or `EditAmendment` but i did inject as much logic as i could into the normal `AmendmentDialog`/`ExtensionDialog` So those are big TODO left over. I felt that was kinda out of scope. 

Not a fan of Create or update so i was trying to keep those separate. 

Screenshot 2025-09-30 at 17 02 50

Screenshot 2025-09-30 at 17 02 55

But using @tresdonicf's pattern i was able to implement simplied creates.


client/src/queries/amendmentQueries.ts
connor-parke-icf, line 1: i would aim to put these queries in the components that are using them; at the moment i think we have defined queries that are not used anywhere in the site, which i think has led us to maintaining/testing fetch operations that are not used.  Keeping the queries next to their usage couples the data fetching with the requirements of the component. For instance, the editAmendment modal doesnt need to know what demonstration its attached to, since its not an editable field anyway; so we shouldnt load it.

connor-parke-icf (commented):
a few architecture-related comments. Main idea is that we should aim to keep the queries next to their usages; the EditAmendment modal needs data for prefilling its form fields, so it should query for that data itself. Nothing else is handling the creation/editing of Amendments, so the mutations should live in the AmendmentModal components themselves. That way, if we need to remove or add something to the edit window, the query that needs to be adjusted is already right there, with a matching type.

The recently updated Demonstration Modals are a really good example to illustrate my comments i think better than i can describe. https://github.com/Enterprise-CMCS/demos/pull/294

client/src/components/dialog/AmendmentDialog.test.tsx
```diff
@@ -1,97 +1,94 @@
...
-  });
-
-  it("shows warning when demonstration is not selected", async () => {
-    render(getAmendmentDialog("add"));
+const AMENDMENT_ID = "amendment-123";
+
+const amendmentMock = {
+  request: {
+    query: AMENDMENT_DIALOG_QUERY,
+    variables: { id: AMENDMENT_ID },
```
dustbuster, line 14: Dumbed this down to be view only.

client/src/components/dialog/CreateAmendmentDialog.tsx
```diff
@@ -0,0 +1,94 @@
+import React from "react";
+import { useToast } from "components/toast";
+import { useMutation } from "@apollo/client";
+import { gql } from "@apollo/client";
+import { CreateAmendmentInput, CreateAmendmentPayload } from "demos-server";
+import { BaseModificationDialog, BaseModificationDialogProps } from "./BaseModificationDialog";
```
dustbuster, line 6: Followed same pattern as CreateDemonstrationDialog.

client/src/components/header/DefaultHeaderLower.tsx
```diff
@@ -110,10 +110,10 @@ export const DefaultHeaderLower: React.FC = () => {
       )}
 
       {modalType === "amendment" && (
-        <AmendmentDialog mode="add" onClose={() => setModalType(null)} />
+        <CreateAmendmentDialog mode="add" onClose={() => setModalType(null)} />
       )}
       {modalType === "extension" && (
-        <ExtensionDialog mode="add" onClose={() => setModalType(null)} />
+        <CreateExtensionDialog mode="add" onClose={() => setModalType(null)} />
```
dustbuster, line 116: Bread and butter!

client/src/components/dialog/ExtensionDialog.tsx
```diff
@@ -26,27 +26,43 @@ export const ExtensionDialog: React.FC<Props> = ({
...
-      // Cast to the proper type since we know the structure from BaseModificationDialog
-      await addExtension.trigger(extensionData as unknown as CreateExtensionInput);
-    } else {
-      // TODO: Implement extension update logic when available
-      console.log("Extension update not yet implemented for ID:", extensionId);
+      const { demonstrationId, name, description } = extensionData as {
+        demonstrationId?: string;
+        name?: string;
+        description?: unknown;
+      };
```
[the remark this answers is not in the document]
dustbuster: I busted out creates for both. But i left the Edit part out. Whatever is in the formal dialog is "To be chopped up" view works in them. But That's all.

server/src/model/modification/modificationResolvers.ts
```diff
@@ -153,25 +167,25 @@ export async function updateExtension(
   return updateModification(parent, args, context, info, extensionBundleTypeId);
 }
 
-async function getParentDemonstration(parent: Modification) {
```
connor-parke-icf, line 156: unsure of this change; the parent here i believe should refer to the Modification object generated by prisma instead of the types we have defined in our schema files. this may work, but it obfuscates the mechanism that is doing the fetching from the database, rather than the return types we use in the rest of the application.

server/src/model/modification/modificationSchema.ts
```diff
@@ -65,15 +65,27 @@ export const modificationSchema = gql`
...
   type Mutation {
-    createAmendment(input: CreateAmendmentInput!): Amendment
+    createAmendment(input: CreateAmendmentInput!): CreateAmendmentResponse!
     updateAmendment(id: ID!, input: UpdateAmendmentInput!): Amendment
     deleteAmendment(id: ID!): Amendment
-    createExtension(input: CreateExtensionInput!): Extension
+    createExtension(input: CreateExtensionInput!): CreateExtensionResponse!
     updateExtension(id: ID!, input: UpdateExtensionInput!): Extension
     deleteExtension(id: ID!): Extension
   }
```
connor-parke-icf, line 86: mentioned it a few times, but really trying to lean away from this pattern where we do not allow the client to query the fields of the created object.

server/src/model/modification/modificationSchema.ts
```diff
@@ -143,3 +155,13 @@ export interface UpdateExtensionInput {
   status?: BundleStatus;
   currentPhaseName?: PhaseName;
 }
+
+export interface CreateAmendmentResponse {
+  success: boolean;
+  message?: string | null;
```
tresdonicf, line 161: FWIW we don't really need the `?` and the `| null`

server/src/model/modification/modificationSchema.ts
```diff
@@ -65,15 +65,27 @@ export const modificationSchema = gql`
...
 
+  type CreateAmendmentResponse {
+    success: Boolean!
+    message: String
+  }
+
+  type CreateExtensionResponse {
+    success: Boolean!
+    message: String
+  }
```
tresdonicf, line 76: I do like these simplified responses as what we're primarily interested in is just whether the operation succeeded (we generally just close the modal and refetch the queries so we hardly need any data back) and this means we can just add fields to this response type as they're needed.

That being said, it would be a good idea to move this into throwing an error proper rather than returning success. I generally like this level of control especially if GQL is going to return everything as a `200` anyway (removing the most canonical signal to rely on) but that's just some food for thought. Might be worth making sure we're actually throwing errors here as well

server/src/model/modification/modificationSchema.ts
tresdonicf: I see this as leaning into the ability to provide just the fields we care about for a specific operation. Most of our operations right now just close a modal and refetch a query so we're discarding the whole response once we get it back. This pattern should hold throughout the rest of the application as well, if we're re-fetching the queries when we do the mutation we shouldn't need to write client code to include the new record in a less-than-proper fashion.

Additionally, this supports the case that we want to return a field that doesn't happen to be in the model but is relevant in the context of the operation. Let's say we want to for some reason return `lastUserToLogin` with this response. If we are returning amendment then we have to do some odd typing to get the query types to match up and it becomes harder to reason about. This less-opinionated version means we just need to add `lastUserToLogin` to `CreateExtensionResponse` and account for it in the resolver which is a bit more straightforward of a change.

That being said, I'm looking to move away from the `success` field in the responses

client/src/pages/DemonstrationDetail/DemonstrationDetailModals.tsx
```diff
@@ -60,5 +67,23 @@ export const DemonstrationDetailModals: React.FC<DemonstrationDetailModalsProps>
...
+      />
+    )}
+
+    {activeExtensionId && (
+      <ExtensionDialog
+        isOpen={true}
+        onClose={onCloseExtensionDialog ?? onCloseEntityModal}
+        mode="view"
+        extensionId={activeExtensionId}
+      />
```
tresdonicf, line 86: Alternatively, `<ExtensionDialog isOpen={Boolean(activeExtensionId)} ... />` 🤯

client/src/components/dialog/ExtensionDialog.tsx
```diff
@@ -1,65 +1,362 @@
...
 
 import { BaseModificationDialog, BaseModificationDialogProps } from "./BaseModificationDialog";
 
-// Pick the props we need from BaseModificationDialogProps and rename entityId to extensionId for clarity
-type Props = Pick<
-  BaseModificationDialogProps,
-  "isOpen" | "onClose" | "mode" | "demonstrationId" | "data"
-> & {
-  extensionId?: string;
+export const EXTENSION_DIALOG_QUERY = gql`
```
tresdonicf, line 16: NBD: a name like `GET_EXISTING_EXTENSION` might tell the story of this a bit better

client/src/components/header/DefaultHeaderLower.tsx
tresdonicf: Awesome to break these out! One thing that would be kinda cool is if it wasn't expecting the `mode` parameter here since we've already provided it with the idea of it being a CreateExtensionDialog in general

tresdonicf (commented):
Overall looking good! I will take another closer look tomorrow, the main comments I have are nits - and I was also wondering about the case of the "View modification" - I wasn't aware of us having requirements for building that out yet so I'm not sure what to make of it

client/src/components/header/DefaultHeaderLower.tsx
dustbuster: yeah i knew you'd say this. I can do this, but i'm way outta scope on this already. so I'm trying to keep with "good enough" for now level of "all the things"

client/src/components/dialog/CreateAmendmentDialog.tsx
```diff
@@ -0,0 +1,90 @@
...
+  amendmentId?: string;
+};
+
+const ERROR_MESSAGE = "Failed to create amendment. Please try again.";
+
+export const CREATE_AMENDMENT_MUTATION = gql`
+  mutation CreateAmendment($input: CreateAmendmentInput!) {
+    createAmendment(input: $input) {
+      success
+      message
```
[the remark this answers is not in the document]
dustbuster: Dude. I just spent a day getting this sorted with success/message and getting rid of the object per @tresdonicf's suggeston.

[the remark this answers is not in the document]
dustbuster: I'm not going back. This is a bug ticket to make these work. And it's fixed. We're at good enough for now mode.

client/src/components/dialog/CreateExtensionDialog.test.tsx
```diff
@@ -0,0 +1,100 @@
...
+const baseMocks: MockedResponse[] = [
+  {
+    request: {
+      query: GET_DEMONSTRATION_OPTIONS_QUERY,
+    },
+    result: {
+      data: {
+        demonstrations: [
+          {
+            __typename: "Demonstration",
```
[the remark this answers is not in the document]
dustbuster: We can come back to cache. It was a Bug fix ticket. And i did a major rewrite. We can come back to some of this stuff it's own ticket.

client/src/components/dialog/ExtensionDialog.tsx
dustbuster: I think `SHOW_EXTENSION` since `show(Model $id)` is what i used previously in APIs for getting 1 thing. 
I don't think we have a good standard.

client/src/components/dialog/ExtensionDialog.tsx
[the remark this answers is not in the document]
dustbuster: I did not write this. We can bust it out into it's own ticket. I'm already super out of scope.

client/src/components/header/DefaultHeaderLower.tsx
dustbuster: Donzo!

server/src/model/modification/modificationSchema.ts
james-becwar-cms, line 76: We need to follow the graphql standards and return client requested fields or a standard graphql error object.
Please see: https://graphql.org/learn/response/ under errors for the pattern we should follow.

james-becwar-cms (changes requested):
Please remove the response types and return standard graphql errors on failure.

PR #299, opened 2025-09-30
DEMOS-793-change-data-type-dates

TARehman:
Small PR that changes the type of the `effectiveDate` and `expirationDate` from `DATE` to `TIMESTAMPTZ`. I also updated the seeder to make marginally better fake data. I checked that the API continues to work as expected. The change to package-lock.json is to add `date-fns-tz` to the server code.

I'm not 100% clear on whether this has a major impact on the front-end. As far as I can tell, the front-end is already using test data, etc that are timestamps, since in JS the Date object has time on it already. I think this is largely self-contained as a result.

Originally, I thought this might have required front-end changes as well, but I've confirmed this is limited to just the DB and the small changes to the GQL types. We'll have a different ticket that integrates the front-end and the back-end, and then another ticket to add a validation / custom scalar for start and end dates to check that what comes in makes sense.


connor-parke-icf (approved):
nothing to really comment on, seems clean to me!

PR #302, opened 2025-09-30
DEMOS-810 - Demonstration Details page Summary tab end Edit Demonstration modal

nunesmlucas:
# Fix EditDemonstrationDialog Field Persistence and Date Formatting Issues

## Summary
This PR resolves critical issues with the EditDemonstrationDialog where field updates were not persisting correctly and date fields were displaying inconsistent values due to timezone handling problems.

## Issues Fixed
- **Field Persistence**: CMCS Division, Signature Level, and date fields were not saving when editing demonstrations
- **Date Display Inconsistency**: SummaryDetailsTable showed dates offset by one day compared to the EditDemonstrationDialog due to timezone conversion issues
- **Project Officer Updates**: Project Officer changes were not being handled through the separate role assignment system
- **Cache Synchronization**: Updated demonstration data was not refreshing across different components after edits

### Enhanced EditDemonstrationDialog Functionality
- Added missing `cmcsDivision` and `signatureLevel` fields to the `UPDATE_DEMONSTRATION_MUTATION` response
- Implemented proper date field handling in the `getUpdateDemonstrationInput` function
- Added separate `SET_DEMONSTRATION_ROLE_MUTATION` to handle Project Officer updates through the role assignment system
- Integrated Apollo cache refetching to ensure UI consistency after updates

### Fixed Date Formatting Consistency
- Resolved timezone issues by implementing consistent date parsing across all components
- Added `safeDateFormat` utility function that handles both Date objects and ISO strings without timezone conversion
- Updated `SummaryDetailsTable`, `DemonstrationDetailHeader`, and `EditDemonstrationDialog` to use unified date formatting approach
- Ensured date display consistency by using string-based date parsing (`split("T")[0]`) to avoid timezone offset problems

### Improved GraphQL Query Integration
- Added missing `cmcsDivision` and `signatureLevel` fields to `DEMONSTRATION_DETAIL_QUERY`
- Implemented proper query refetching to synchronize data across components after updates
- Enhanced error handling and user feedback for failed updates

## Technical Details
- **Date Handling**: Switched from timezone-sensitive `new Date()` parsing to string-based date extraction to prevent off-by-one day issues
- **Role Management**: Integrated with existing role assignment mutations for Project Officer updates
- **Cache Management**: Added `refetchQueries` to mutations to ensure immediate UI updates
- **Type Safety**: Maintained TypeScript compatibility throughout all changes

## Testing
- Verified that all demonstration fields (name, description, dates, CMCS Division, Signature Level, Project Officer) now save and persist correctly
- Confirmed date consistency between edit dialog and display components
- Tested real database integration with proper field population and persistence
- Validated that changes are immediately visible after saving without requiring page refresh

## Files Modified
- `EditDemonstrationDialog.tsx` - Enhanced with complete field saving functionality
- `DemonstrationDetail.tsx` - Added missing GraphQL query fields
- `SummaryDetailsTable.tsx` - Fixed date formatting with timezone-safe approach

image
image

### DIFF
--- a/client/src/components/dialog/demonstration/DemonstrationDialog.test.tsx
+++ b/client/src/components/dialog/demonstration/DemonstrationDialog.test.tsx
@@ -1,9 +1,29 @@
 import React from "react";
-import { render, screen, fireEvent, waitFor } from "@testing-library/react";
-import { describe, it, expect, vi } from "vitest";
-import { CreateDemonstrationDialog, EditDemonstrationDialog } from "./";
+
+import { GET_USER_SELECT_OPTIONS_QUERY } from "components/input/select/SelectUsers";
 import { TestProvider } from "test-utils/TestProvider";
-import { GET_DEMONSTRATION_BY_ID_QUERY } from "./EditDemonstrationDialog";
+import {
+  describe,
+  expect,
+  it,
+  vi,
+} from "vitest";
+
+import {
+  fireEvent,
+  render,
+  screen,
+  waitFor,
+} from "@testing-library/react";
+
+import {
+  CreateDemonstrationDialog,
+  EditDemonstrationDialog,
+} from "./";
+import {
+  GET_DEMONSTRATION_BY_ID_QUERY,
+  UPDATE_DEMONSTRATION_MUTATION,
+} from "./EditDemonstrationDialog";
 
 const DEFAULT_DEMONSTRATION = {
   name: "",
@@ -25,9 +45,26 @@ const SUBMIT_BUTTON_TEST_ID = "button-submit-demonstration-dialog";
 const CANCEL_BUTTON_TEST_ID = "button-cancel-demonstration-dialog";
 
 describe("CreateDemonstrationDialog", () => {
+  const GET_USER_SELECT_OPTIONS_MOCK = {
+    request: {
+      query: GET_USER_SELECT_OPTIONS_QUERY,
+    },
+    result: {
+      data: {
+        people: [
+          {
+            id: "test-officer-id",
+            fullName: "Test Officer",
+            personType: "demos-cms-user",
+          },
+        ],
+      },
+    },
+  };
+
   const getCreateDemonstrationDialog = () => {
     return (
-      <TestProvider>
+      <TestProvider mocks={[GET_USER_SELECT_OPTIONS_MOCK]}>
         <CreateDemonstrationDialog {...DEFAULT_PROPS} />
       </TestProvider>
     );
@@ -61,7 +98,25 @@ describe("CreateDemonstrationDialog", () => {
 });
 
 describe("EditDemonstrationDialog", () => {
-  const TEST_DEMO_ID = "test-demo-id";
+  const TEST_DEMO_ID = "1";
+
+  const GET_USER_SELECT_OPTIONS_MOCK = {
+    request: {
+      query: GET_USER_SELECT_OPTIONS_QUERY,
+    },
+    result: {
+      data: {
+        people: [
+          {
+            id: "test-officer-id",
+            fullName: "Test Officer",
+            personType: "demos-cms-user",
+          },
+        ],
+      },
+    },
+  };
+
   const GET_DEMONSTRATION_BY_ID_MOCK = {
     request: {
       query: GET_DEMONSTRATION_BY_ID_QUERY,
@@ -73,16 +128,54 @@ describe("EditDemonstrationDialog", () => {
           id: TEST_DEMO_ID,
           name: "Test Demonstration",
           description: "Test demonstration description",
-          cmcsDivision: "Test Division",
-          signatureLevel: "Test Level",
+          cmcsDivision: "Division of System Reform Demonstrations",
+          signatureLevel: "OA",
           state: {
             id: "test-state-id",
           },
-          projectOfficer: {
-            id: "test-officer-id",
+          roles: [
+            {
+              isPrimary: true,
+              role: "Project Officer",
+              person: {
+                id: "test-officer-id",
+              },
+            },
+          ],
+          effectiveDate: "2023-01-01T00:00:00.000Z",
+          expirationDate: "2024-01-01T00:00:00.000Z",
+        },
+      },
+    },
+  };
+
+  // Mock for the update mutation - this needs to match what the form actually submits
+  const UPDATE_DEMONSTRATION_MOCK = {
+    request: {
+      query: UPDATE_DEMONSTRATION_MUTATION,
+      variables: {
+        id: "1",
+        input: {
+          name: "Test Demonstration 123",
+          description: "A test demonstration.",
+          stateId: "AL",
+          cmcsDivision: "Division of System Reform Demonstrations",
+          signatureLevel: "OA",
+        },
+      },
+    },
+    result: {
+      data: {
+        updateDemonstration: {
+          id: "1",
+          name: "Test Demonstration 123",
+          description: "A test demonstration.",
+          effectiveDate: "2023-01-01T00:00:00.000Z",
+          expirationDate: "2024-01-01T00:00:00.000Z",
+          state: {
+            id: "AL",
+            name: "Alabama",
           },
-          effectiveDate: "2023-01-01",
-          expirationDate: "2024-01-01",
         },
       },
     },
@@ -90,7 +183,13 @@ describe("EditDemonstrationDialog", () => {
 
   const getEditDemonstrationDialog = () => {
     return (
-      <TestProvider mocks={[GET_DEMONSTRATION_BY_ID_MOCK]}>
+      <TestProvider
+        mocks={[
+          GET_DEMONSTRATION_BY_ID_MOCK,
+          GET_USER_SELECT_OPTIONS_MOCK,
+          UPDATE_DEMONSTRATION_MOCK,
+        ]}
+      >
         <EditDemonstrationDialog {...DEFAULT_PROPS} demonstrationId={TEST_DEMO_ID} />
       </TestProvider>
     );
@@ -116,7 +215,36 @@ describe("EditDemonstrationDialog", () => {
   });
 
   it("disables Submit button if required fields are empty", async () => {
-    render(getEditDemonstrationDialog());
+    // Create a mock with incomplete data (missing state and project officer)
+    const incompleteDataMock = {
+      request: {
+        query: GET_DEMONSTRATION_BY_ID_QUERY,
+        variables: { id: TEST_DEMO_ID },
+      },
+      result: {
+        data: {
+          demonstration: {
+            id: TEST_DEMO_ID,
+            name: "", // Empty name
+            description: "Test demonstration description",
+            cmcsDivision: "Division of System Reform Demonstrations",
+            signatureLevel: "OA",
+            state: {
+              id: "", // Empty state ID
+            },
+            roles: [], // No project officer assigned
+            effectiveDate: "2023-01-01T00:00:00.000Z",
+            expirationDate: "2024-01-01T00:00:00.000Z",
+          },
+        },
+      },
+    };
+
+    render(
+      <TestProvider mocks={[incompleteDataMock, GET_USER_SELECT_OPTIONS_MOCK]}>
+        <EditDemonstrationDialog {...DEFAULT_PROPS} demonstrationId={TEST_DEMO_ID} />
+      </TestProvider>
+    );
 
     // Wait for loading to complete
     await waitFor(() => {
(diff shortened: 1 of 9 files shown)


client/src/components/dialog/demonstration/EditDemonstrationDialog.tsx
tresdonicf, line 1: For my current ticket [DEMOS-782](https://jiraent.cms.gov/browse/DEMOS-782) I'm working on this as well - I approached it a little differently, I'm happy to review this and make some edits perhaps if you're good with that - just wanted to give a heads up!

client/src/components/dialog/demonstration/EditDemonstrationDialog.tsx
```diff
@@ -52,8 +77,8 @@ const getUpdateDemonstrationInput = (
   stateId: demonstration.stateId,
   sdgDivision: demonstration.sdgDivision,
   signatureLevel: demonstration.signatureLevel,
-  // effectiveDate: demonstration.effectiveDate,
-  // expirationDate: demonstration.expirationDate,
+  effectiveDate: demonstration.effectiveDate ? new Date(demonstration.effectiveDate) : undefined,
+  expirationDate: demonstration.expirationDate ? new Date(demonstration.expirationDate) : undefined,
```
tresdonicf, line 81: Think we'll want to use getStartOfDayEst() and getEndOfDayEst() for these (I can make that edit)

client/src/components/dialog/demonstration/DemonstrationDialog.test.tsx
```diff
@@ -73,24 +128,68 @@ describe("EditDemonstrationDialog", () => {
...
-            id: "test-officer-id",
+          roles: [
+            {
+              isPrimary: true,
+              role: "Project Officer",
+              person: {
+                id: "test-officer-id",
+              },
+            },
+          ],
```
connor-parke-icf, line 144: i dont remember where exactly we landed on whether we wanted to remove editing the project officer from the edit modal, or if all forms of editing roles should be done in the (to come) contacts modal.  I think this is fine though, until thats all cleared up

client/src/components/dialog/demonstration/DemonstrationDialog.test.tsx
connor-parke-icf, line 127: as a complete aside, not sure how much this matters if at all, but we might can type this demonstration object to be based on the return value of the query in the component, just to make sure the mocked data is as representative as possible of the expected value.

client/src/mock-data/amendmentMocks.ts
tresdonicf, line 1: Not sure of the need to change the names of these mocks if it's going to cause lots of changes in the tests

client/src/components/table/tables/SummaryDetailsTable.tsx
```diff
@@ -4,18 +4,44 @@ import { SecondaryButton } from "components/button";
...
+
+    if (date instanceof Date) {
+      return formatDate(date);
+    }
+
+    return "--/--/----";
+  } catch {
+    return "--/--/----";
+  }
+};
```
tresdonicf, line 34: This seems to be defined twice. it should probably be defined once in `formatDate.ts`

client/src/components/dialog/demonstration/EditDemonstrationDialog.tsx
connor-parke-icf, line 51: not really related to this ticket, but do we need to retrieve the state's name if we have the ID here?

client/src/components/dialog/demonstration/EditDemonstrationDialog.tsx
```diff
@@ -67,18 +92,67 @@ export const EditDemonstrationDialog: React.FC<{
     variables: { id: demonstrationId },
   });
   const [updateDemonstrationTrigger] = useMutation(UPDATE_DEMONSTRATION_MUTATION);
+  const [setDemonstrationRoleTrigger] = useMutation(SET_DEMONSTRATION_ROLE_MUTATION);
```
connor-parke-icf, line 95: if we do want to allow for setting the project officer from the edit Demonstration modal, i think we might want to transactional-ize this, such that if it fails to set the project officer for some reason, it also rolls back the edit operation in general. That probably points to updating the updateDemonstration resolver such that includes the projectOfficerId in the input. Think there is a wider discussion here worth exploring in office hours, but it might depend on the contacts flow (whether we want to allow users to change the PO from this modal or force them to do so via the Assign Contacts modals)

Also thinking about whether we want to reinclude projectOfficer directly on the demonstration GQL object as a shortcut for "roles[<where is PO and Primary>].person".  might clean up the frontend so we dont need to include so much role info for such a common usecase.

client/src/components/dialog/demonstration/EditDemonstrationDialog.tsx
```diff
@@ -67,18 +92,67 @@ export const EditDemonstrationDialog: React.FC<{
...
+          refetchQueries: [
+            {
+              query: GET_DEMONSTRATION_BY_ID_QUERY,
+              variables: { id: demonstrationId },
+            },
+            {
+              query: DEMONSTRATION_DETAIL_QUERY,
+              variables: { id: demonstrationId },
+            },
+          ],
```
connor-parke-icf, line 148: really hoping with caching this refetch queries pattern can get abstracted away such that when the cache is invalidated (via mutation) apollo knows to refetch the queries that touched the objects automatically.  adding refetch queries on all of our queries might not scale very well.

client/src/components/dialog/demonstration/EditDemonstrationDialog.tsx
```diff
@@ -90,13 +164,33 @@ export const EditDemonstrationDialog: React.FC<{
...
+            role.role === "Project Officer" && role.isPrimary === true
+        );
+        return primaryProjectOfficer?.person?.id || "";
+      })(),
+      effectiveDate: data.demonstration.effectiveDate
+        ? data.demonstration.effectiveDate.split("T")[0]
+        : "",
+      expirationDate: data.demonstration.expirationDate
+        ? data.demonstration.expirationDate.split("T")[0]
+        : "",
```
connor-parke-icf, line 183: might be a date-fns function we could use to handle this instead of string manipulation; not that it should really make a difference in this case.

client/src/components/dialog/demonstration/EditDemonstrationDialog.tsx
```diff
@@ -90,13 +164,33 @@ export const EditDemonstrationDialog: React.FC<{
     return <div>Error loading demonstration data.</div>;
   }
 
+  const transformedDemonstration = data?.demonstration
```
connor-parke-icf, line 167: maybe rename to something like "flattenedDemonstration", to specify the kind of transformation (to prepare for the form fields)

client/src/pages/DemonstrationDetail/DemonstrationDetail.tsx
```diff
@@ -1,22 +1,44 @@
...
+import {
+  Document,
+  Person,
+} from "demos-server";
 import { usePageHeader } from "hooks/usePageHeader";
-import { Tab, Tabs } from "layout/Tabs";
+import {
+  Tab,
+  Tabs,
+} from "layout/Tabs";
```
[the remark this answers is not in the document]
dustbuster: I vote 1 liner on this too. But this is more a nit.

client/src/components/table/tables/SummaryDetailsTable.test.tsx
```diff
@@ -1,16 +1,39 @@
...
+  name: mockDemo.name,
+  description: mockDemo.description,
+  sdgDivision: mockDemo.sdgDivision,
+  signatureLevel: mockDemo.signatureLevel,
+  effectiveDate: new Date(mockDemo.effectiveDate),
+  expirationDate: new Date(mockDemo.expirationDate),
+  status: mockDemo.status,
+  state: mockDemo.state,
+  roles: mockDemo.roles,
+});
```
connor-parke-icf, line 35: not sure im following, is the purpose here just to convert the effective and expiration dates to Date objects?

client/src/components/table/tables/SummaryDetailsTable.tsx
```diff
@@ -4,18 +4,44 @@ import { SecondaryButton } from "components/button";
...
+
```
connor-parke-icf, line 35: we have a date utils file we can put this into i think; there are a few places we're doing similar like the demonstration details header.  That being said, i think we might need to figure out exactly how we're expecting to get dates from the backend, whether theyre date objects already or where we might want to do conversions.

client/src/mock-data/amendmentMocks.ts
```diff
@@ -8,37 +8,37 @@ export type MockAmendment = Pick<Amendment, "id" | "name" | "effectiveDate"> & {
 export const mockAmendments = [
   {
     id: "1",
-    name: "Amendment 1 - Montana Medicaid Waiver",
+    name: "Amendment 1 - Test Demonstration 1",
```
connor-parke-icf, line 11: i dont mind the renaming of these mocked values, im more curious why. I can totally see it making the data a lot easier to reference and work with, but i think it led to a lot of changes in other tests which mightve been unrelated to the main story (demo table, modification tables)

client/src/mock-data/demonstrationMocks.ts
```diff
@@ -1,31 +1,38 @@
...
+  mockStates,
+} from "./stateMocks";
 
 export type MockDemonstration = Pick<
   Demonstration,
-  "id" | "name" | "description" | "effectiveDate" | "expirationDate"
+  "id" | "name" | "description" | "sdgDivision" | "signatureLevel"
 > & {
+  effectiveDate: string | null;
+  expirationDate: string | null;
```
connor-parke-icf, line 35: wherever we can, i think we should rely on the Pick<> syntax for scalar fields, so that they can reflect any changes to the server definition. If effectiveDate and expirationDate are nullable (sounds like they are), we might should update that in the demonstrationSchema so its propagated everywhere, and the type of effective/expirationDate is automatically string | null just by saying Pick<Demonstration, "effectiveDate" | "expirationDate">

client/src/mock-data/demonstrationMocks.ts
```diff
@@ -38,57 +45,134 @@ export type MockDemonstration = Pick<
...
   {
     id: "1",
-    name: "Montana Medicaid Waiver",
-    description: "A demonstration project in Montana.",
-    effectiveDate: new Date(2025, 0, 1),
-    expirationDate: new Date(2025, 11, 1),
+    name: "Test Demonstration 1",
+    description: "A test demonstration.",
+    effectiveDate: "2025-01-01T00:00:00.000Z",
+    expirationDate: "2025-12-01T00:00:00.000Z",
```
connor-parke-icf, line 51: if we are going with timestamp strings instead of date strings (and the strings i imagine better represent the GQL query results), we might want to set this test data's expiration date to be the last second of the day, rather than midnight to align with the time range convention we're leaning towards. Not sure if that matters at all, since this is mock data.

client/src/mock-data/demonstrationMocks.ts
```diff
@@ -38,57 +45,134 @@ export type MockDemonstration = Pick<
...
+        effectiveDate: new Date(2025, 1, 1),
+        status: "Approved",
+      },
+      {
+        id: "3",
+        name: "Extension 3 - Test Demonstration 1",
+        effectiveDate: null,
+        status: "Approved",
+      },
+    ],
```
connor-parke-icf, line 95: curious why we're redefining mock amendments when we can pull from the list that exists in the amendmentMocks.  The primary concern with splitting it up that way was that we didnt have to define a new set of data for each object, we can just stitch together the related objects how we needed.  Was it a circular dependency issue?

client/src/mock-data/demonstrationMocks.ts
```diff
@@ -162,29 +264,126 @@ export const demonstrationMocks: MockedResponse[] = [
...
+        id: "2",
+        input: {
+          name: "Test Demonstration 2",
+          description: "Another test demonstration.",
+          stateId: "AK",
+          sdgDivision: "Division of Eligibility and Coverage Demonstrations",
+          signatureLevel: "OCD",
+        },
+      },
+    },
```
connor-parke-icf, line 343: there is a way to parameterize mock variables such that they can take in a range of different variable sets and output a result based on that; we might can compress these mocks into one of those.

client/src/pages/DemonstrationDetail/DemonstrationDetailHeader.tsx
```diff
@@ -1,11 +1,36 @@
...
+      const datePart = isoString.split("T")[0];
+      const [year, month, day] = datePart.split("-");
+      return `${month}/${day}/${year}`;
+    }
+
+    return "--/--/----";
+  } catch {
+    return "--/--/----";
+  }
+};
```
connor-parke-icf, line 33: we can centralize this logic i think

client/src/mock-data/demonstrationMocks.ts
nunesmlucas: Left this since we didn't update the demonstrationSchema and I've already made a chunk of mocks changes.

PR #326, opened 2025-10-06
DEMOS-904 converted some selects to use react-select

connor-parke-icf:
this is a preliminary pass at seeing what can be done with the react-select component as the core of our selects across the application. As i was going through this, i found myself making some decisions that _might_ be shortsighted, so hoping to get some feedback on this before i go much deeper

a few things to note: 
- tried making a select which can handle both single-selection and multiselection but kept running into issues with typing the other params. For instance, if its a multiselect, the component should expect a value thats an array, not a single value.  This is actually not that big of a deal, since we are only using multiselects in our table component, so opting to having a separate multiselect component for those, and single selects for the rest. 
- we are able to combine the "searchable" and "non-searchable" selects pretty easily. 
- we had at least one select that wasnt actually used anywhere other than the library and tests, so removed it


client/src/components/application/phases/ConceptPhase.tsx
```diff
@@ -210,12 +210,12 @@ export const ConceptPhase: React.FC<Props> = ({
...
         </div>
 
-        <AutoCompleteSelect
+        <Select
           id="demo-type"
           label="Demonstration Type(s) Requested"
           options={DEMONSTRATION_TYPE_OPTIONS}
           value={demoType}
-          onSelect={(v) => setDemoType(String(v))}
+          onChange={(v) => setDemoType(String(v))}
```
connor-parke-icf, line 218: switched onChange to unify the language between multi and single selects, and to align with the props in react-select.

client/src/components/dialog/demonstration/DemonstrationDialog.tsx
```diff
@@ -241,12 +240,12 @@ export const DemonstrationDialog: React.FC<{
...
-            initialValue={activeDemonstration.sdgDivision}
-            onSelect={(sdgDivision) => setActiveDemonstration((prev) => ({ ...prev, sdgDivision }))}
+            value={activeDemonstration.sdgDivision}
+            onChange={(sdgDivision) => setActiveDemonstration((prev) => ({ ...prev, sdgDivision }))}
           />
           <SelectSignatureLevel
-            initialValue={activeDemonstration.signatureLevel}
-            onSelect={(signatureLevel) =>
+            value={activeDemonstration.signatureLevel}
+            onChange={(signatureLevel) =>
```
connor-parke-icf, line 248: throughout it seemed that the "initialValue" could just be the state-managed value.  Discussion on managed-unmanaged select further below, in the changes to the inputs

client/src/components/dialog/document/DocumentDialog.tsx
```diff
@@ -424,12 +424,12 @@ const DocumentDialog: React.FC<DocumentDialogProps> = ({
...
 
-      <DocumentTypeInput
+      <SelectDocumentTypes
         value={activeDocument.documentType}
-        onSelect={(val) => {
+        onChange={(val) => {
           setActiveDocument((prev) => ({ ...prev, documentType: val as DocumentType }));
         }}
-        documentTypeSubset={documentTypeSubset}
+        options={documentTypeSubset}
```
connor-parke-icf, line 432: this is an odd case i think I want to talk about. If we're supplying our own options, it may be better to simply use a base Select supplied with options rather than a documentTypeselect here.

client/src/components/input/select/Select.tsx
```diff
@@ -1,34 +1,46 @@
 import { ChevronDownIcon } from "components/icons";
 import React from "react";
 import { tw } from "tags/tw";
+import ReactSelect from "react-select";
+
 const ICON_CLASSES = tw`text-text-placeholder w-2 h-1`;
 
-export interface Option {
+export interface Option<T = string> {
```
connor-parke-icf, line 8: added a generic type here so that we can use the enumerations we have on Server. So for instance, the DocumentType select will take on options of type 
{
    label: string,
    value: DocumentType
}

client/src/components/input/select/Select.tsx
```diff
@@ -1,34 +1,46 @@
...
-  value,
-}) => {
+}: SelectProps<T>) => {
+  // Convert value to Option format for ReactSelect
+  const selectedOption = value ? options.find((option) => option.value === value) : null;
+
+  // Handle ReactSelect onChange and convert back to value type
+  const handleChange = (selectedOption: Option<T> | null) => {
+    onChange(selectedOption ? selectedOption.value : ("" as T));
+  };
```
connor-parke-icf, line 42: it seems react-select is built to work with the option object as its atom, rather than a string value which is used for vanilla-html selects. Because of this we basically had two options, 1) wrap our handlers with mappings between option objects and the string values (which i went with) or 2) return and pass in the option objects themselves. Went with the former, since it required fewer changes and seemed more appropriate of our use-case where most of our selects are based on single string options.

client/src/components/input/select/Select.tsx
```diff
@@ -1,34 +1,46 @@
...
 }
 
-export interface SelectProps {
-  options: Option[];
-  placeholder?: string;
-  onSelect: (value: string) => void;
+export interface SelectProps<T = string> {
+  options: Option<T>[];
+  onChange: (value: T) => void;
+  value?: T;
```
connor-parke-icf, line 16: went back and forth on this for a while, and ultimately didnt land anywhere in particular. But not sure if we want to leave this selects able to be unmanaged or force them to be used in a managed-state pattern. Found that i think all except a couple of our use-cases have the parent component manage the value, so it may be a good idea to force a value to be supplied, even if that value is null. Of course, that means in all of our tests for these selects we have to supply some kinda value for it to track. Which isnt a big deal, but is a decision with a few implications that i want to gauge opinion on.

client/src/components/input/select/SelectDocumentTypes.tsx
```diff
@@ -0,0 +1,25 @@
...
+      placeholder={placeholder}
+      id={id}
+      label={label}
+      options={options.map((documentType) => ({
+        value: documentType,
+        label: documentType,
+      }))}
+      {...rest}
+    />
+  );
```
connor-parke-icf, line 24: DocumentTypes is the odd case among these, since there are places where we want to provide our own options to override the default ones. At that point, it may be better to simply use the base Select to handle that, but that would require making an array of options to pass in; not just an array of values (at least, how its set up now). I might look to move some of that logic around. Its not a problem for the static constants since their mapping of value to option is simple, but for People and States, its more complicated, so not sure if we can have a default to rely on here.

PR #329, opened 2025-10-06
DEMOS-782: Edit Demonstration

tresdonicf:
This was largely fixed in DEMOS-810 by @nunesmlucas - I mostly just made some refactoring changes and paid special attention to the date formatting / parsing

Changes:

- Simplified top level component using functions and local "hooks"
- Broke tests for DemonstrationDialog out into separate files

https://github.com/user-attachments/assets/879d2b01-5041-4f10-8e8c-a3cf8b67cd0d

### DIFF
--- a/client/src/components/dialog/demonstration/DemonstrationDialog.test.tsx
+++ b/client/src/components/dialog/demonstration/DemonstrationDialog.test.tsx
@@ -4,13 +4,9 @@ import { GET_USER_SELECT_OPTIONS_QUERY } from "components/input/select/SelectUse
 import { TestProvider } from "test-utils/TestProvider";
 import { describe, expect, it, vi } from "vitest";
 
-import { fireEvent, render, screen, waitFor } from "@testing-library/react";
+import { render, screen } from "@testing-library/react";
 
-import { CreateDemonstrationDialog, EditDemonstrationDialog } from "./";
-import {
-  GET_DEMONSTRATION_BY_ID_QUERY,
-  UPDATE_DEMONSTRATION_MUTATION,
-} from "./EditDemonstrationDialog";
+import { DemonstrationDialog } from "./DemonstrationDialog";
 
 const DEFAULT_DEMONSTRATION = {
   name: "",
@@ -26,12 +22,13 @@ const DEFAULT_PROPS = {
   onClose: vi.fn(),
   onSubmit: vi.fn(),
   initialDemonstration: DEFAULT_DEMONSTRATION,
+  mode: "create" as const,
 };
 
 const SUBMIT_BUTTON_TEST_ID = "button-submit-demonstration-dialog";
 const CANCEL_BUTTON_TEST_ID = "button-cancel-demonstration-dialog";
 
-describe("CreateDemonstrationDialog", () => {
+describe("DemonstrationDialog", () => {
   const GET_USER_SELECT_OPTIONS_MOCK = {
     request: {
       query: GET_USER_SELECT_OPTIONS_QUERY,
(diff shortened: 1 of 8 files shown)


client/src/components/dialog/demonstration/EditDemonstrationDialog.tsx
```diff
@@ -72,125 +71,137 @@ export const SET_DEMONSTRATION_ROLE_MUTATION = gql`
...
-          {
-            query: GET_DEMONSTRATION_BY_ID_QUERY,
-            variables: { id: demonstrationId },
-          },
-          {
-            query: DEMONSTRATION_DETAIL_QUERY,
-            variables: { id: demonstrationId },
-          },
-        ],
-      });
```
tresdonicf, line 116: Moved some stuff like this into local `hooks` in the file to clean up the story of this file a bit

client/src/components/dialog/demonstration/DemonstrationDialog.test.tsx
tresdonicf, line 1: Broke the functionality tests out into `CreateDemonstrationDialog` and `EditDemonstrationDialog` test files

client/src/components/dialog/demonstration/EditDemonstrationDialog.tsx
```diff
@@ -72,125 +71,141 @@ export const SET_DEMONSTRATION_ROLE_MUTATION = gql`
...
+    const updateInput = getUpdateDemonstrationInput(demonstrationDialogFields);
+
+    await updateDemonstrationTrigger({
+      variables: {
+        id: demonstrationId,
+        input: updateInput,
+      },
+      refetchQueries: [
+        {
+          query: DEMONSTRATION_DETAIL_QUERY,
```
connor-parke-icf, line 127: because we arent updating creating any new objects, we shouldnt have to run any explicit refetches, so long as we return from the mutation the fields that were changed. it should return the demonstration object, find the entry in the cache corresponding to the demonstration id, update those values with the new ones returned, and update all of the queries that were dependent on those values in the cache (refetching them if anything is missing, but i dont think anything is in this case), and rerender the appropriate displays. 

I would see if everything frontend is updated automagically without the explicit refetch; i think it should in this case.

client/src/components/dialog/demonstration/EditDemonstrationDialog.tsx
```diff
@@ -72,125 +71,141 @@ export const SET_DEMONSTRATION_ROLE_MUTATION = gql`
...
+          roleId: "Project Officer",
+          isPrimary: true,
+        },
+      },
+      refetchQueries: [
+        {
+          query: DEMONSTRATION_DETAIL_QUERY,
+          variables: { id: demonstrationId },
+        },
+      ],
```
connor-parke-icf, line 161: here im not actually sure the cache behavior; i believe this mutation returns the role assignment object and not the demonstration, so it may require some kind of refetch/cache updating. I havnt looked into the behavior of how a list of items attached to another object gets updated.  It may be that if we return the demonstrationId on these roles that its smart enough to realize that the list of roles on the demonstration needs to be updated, but im not sure.

connor-parke-icf (approved):
nice! couple of comments, but nothing really blocking. Main thing im thinking about is whether we actually need to explicitly refetch queries after the mutation or if it will get taken care of automagically. doesnt hurt anything if we do explicitly refetch though

client/src/components/dialog/demonstration/EditDemonstrationDialog.tsx
```diff
@@ -72,125 +71,141 @@ export const SET_DEMONSTRATION_ROLE_MUTATION = gql`
...
+    : "",
+  expirationDate: demonstration.expirationDate
+    ? formatDateForServer(demonstration.expirationDate)
+    : "",
 });
 
+const getProjectOfficerId = (demonstration: Demonstration): string => {
+  const primaryProjectOfficer = demonstration.roles.find(
+    (role) => role.role === "Project Officer" && role.isPrimary === true
+  );
```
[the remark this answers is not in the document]
tresdonicf: That would certainly be helpful! I totally agree it might end up being common, luckily it's simple to write so however you think the server should go about that we can work around it on the FE

client/src/components/dialog/demonstration/DemonstrationDialog.tsx
```diff
@@ -183,16 +183,18 @@ export const DemonstrationDialog: React.FC<{
...
       <form id="demonstration-form" className="flex flex-col gap-[24px]" onSubmit={handleSubmit}>
         <div className="grid grid-cols-3 gap-[24px]">
-          <SelectUSAStates
-            label="State/Territory"
-            currentState={activeDemonstration.stateId}
-            value={activeDemonstration.stateId}
-            isRequired
-            onStateChange={(stateId) => setActiveDemonstration((prev) => ({ ...prev, stateId }))}
-          />
+          <div data-testid="select-state">
```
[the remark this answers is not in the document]
tresdonicf: Fixed!

client/src/components/dialog/demonstration/EditDemonstrationDialog.tsx
connor-parke-icf: i guess the larger point im making here is that the explicit refetch might not be doing anything; we are already returning the ID of the created object, so apollo will automatically refetch what it needs anyway unless we explicitly stop it from doing that

PR #330, opened 2025-10-07
Demos 787 amendment or extension records not linked to demonstration

connor-parke-icf:


client/src/components/dialog/AmendmentDialog.tsx
```diff
@@ -22,31 +38,51 @@ export const AmendmentDialog: React.FC<Props> = ({
...
+          id: cache.identify({
+            __typename: "Demonstration",
+            id: amendment.demonstration.id,
+          }),
+          fields: {
+            amendments(existingAmendments = []) {
+              return [...existingAmendments, amendment];
+            },
+          },
+        });
```
connor-parke-icf, line 60: i prefer this style of cache invalidation to refetching queries individually, as we dont need to know what queries need refetching. the queries that pulled the data initially decide whether they need to refetch or not.  Adding more queries to the page, or breaking up the main-toplevel query on this page to smaller chunks will have no effect on this mutation. If we broke the toplevel query into one for mutations/amendments and one for the rest of the details, we would require no edits to this mutation for it to still refetch the appropriate query. 

The procedure is fairly straightforward:
* the mutation returns an amendment with an id, and its demonstration containing an id
* get the demonstration ID, and use it to find the demonstration in the cache
* add the newly created amendment to the demonstration's list of amendments

this triggers a potential refetch of the demonstration detail query, which requires data on the amendment that does not exist on the cache, so it refetches, updating everything.

dustbuster: Can abstract the caching away into it's own component?
Therefore making it reusable? This does not have to be a right now thing. But once in React do as the reactors do and make it a function!

Quick proto from AI:
```
export const useCacheUpdater = () => {
  const client = useApolloClient();

  const appendToList = <T>(
    parent: CacheParent,
    fieldName: string,
    newItem: T
  ) => {
    client.cache.modify({
      id: client.cache.identify(parent),
      fields: {
        [fieldName](existingItems: T[] = []) {
          return [...existingItems, newItem];
        },
      },
    });
  };

  const removeFromList = <T extends { id: string | number }>(
    ... more stuff ...

  const replaceInList = <T extends { id: string | number }>(
    ... more stuff ...
    });
  };

  return { appendToList, removeFromList, replaceInList };
};
```

tresdonicf: There might be a misunderstanding of how this analysis relates to our use case and current application state. If you take a look at the [apollo docs on mutations](https://www.apollographql.com/docs/react/data/mutations) you can see that the recommended approach is to combine them in production code, with `refecthQueries` being used for making sure your data is always up to date (data integrity) and the cache invalidation being a strategy to help gains in performance (local rendering only, works offline). There isn't really an "either / or" discussion to be had here but the idea being to start with the simple, reliable approach and moving to the optimizations when / if it's needed. 

So for where we're currently at we should be implementing `refetchQueries` as we care most about data correctness and not about performance. Additionally, we don't have to know which queries need to be updated, as there are shorthands for `active` and `all` that could be used here if we want. Personally I would avoid them and just think about the components that need updating and grab the query names out of there but that's not really something we need to design around with the data volume of our application at this point.

The other benefit to the `refetchQueries` approach is that it's easy to reason about: if a component we expect to be updated is not being updated when we call a mutation we know we just need to add that query to the list and bam we're done 😎 

This isn't to say we shouldn't implement the cache modifications but it should likely be after a `refetchQueries`  declaration and we might find that we don't need it much at all. From reading around it also seems that there are some ways to mismanage this cache that are hard to interpret and lead to some strange bugs so I'd prefer we didn't introduce this pattern until the app design / models are in a bit less flux.

Image

tresdonicf (approved):
Thanks for making the changes re: refetchQueries!

PR #334, opened 2025-10-07
DEMOS-846-date-validator

TARehman:
This implements validation of various date fields on input. It also happens to fix the DateTime scalar / resolver so that incoming strings are properly cast to actual JS Date objects, which wasn't being done before.

 - Adds `date-fns` so that it can be used for various date checking.
 - Adds a `CHECK` constraint to the database on `demonstration` and `modification` which prevents the `effectiveDate` from being after the `expirationDate`.
 - Updates the `handlePrismaError` function to return more information and to parse when a `CHECK` constraint is violated (it turns out these are considered `Prisma.PrismaClientUnknownRequestError` objects).
 - Updates the unit tests for the `handlePrismaError` function accordingly.
 - Wraps update statements for `demonstration` and `modification` into try/catch blocks and routes them through `handlePrismaError`.
 - Adds `validateDateInput` and the accompanying support functions to implement checks for dates that are not `effectiveDate` and `expirationDate`.
 - Adds tests for the new validation suite.
 - Adds a `customScalarResolvers` object to `graphql.ts` and properly assigns the `graphql-scalars` resolver for DateTime to the DateTime scalar.
 - Updates the seeder to correctly put dates in order to avoid making the `CHECK` constraint angry.

I've validated that this works by doing the following.

 - Writing unit tests for the changes that have been made.
 - Exercising the API to verify that the results are as expected. Example images below.

image

image

image

image


server/src/model/bundleDate/getTargetDateValue.ts
```diff
@@ -0,0 +1,22 @@
...
+        dateTypeId: dateType,
+      },
+    },
+  });
+  if (result === null) {
+    throw new Error(
+      `The date ${dateType} for bundle ${bundleId} was requested as part of a validation, but is undefined.`
+    );
+  }
+  return result.dateValue;
```
connor-parke-icf, line 21: same with this; not sure if this needs to be explicitly tested for and handled in its own unit test if the only place we use this is in the validateInputDate.  I think i would rather have this logic as part of the validateInputDate file so everything date-validation is in the same place

server/src/model/bundleDate/validateInputDate.ts
```diff
@@ -0,0 +1,71 @@
...
+  checkInputDateMeetsOffset,
+} from "./checkInputDateFunctions.js";
+
+export async function validateInputDate(input: SetBundleDateInput): Promise<void> {
+  const inputDate = {
+    dateType: input.dateType,
+    dateValue: input.dateValue,
+  };
+
+  switch (input.dateType) {
```
connor-parke-icf, line 14: This is a pretty big switch-case structure. Im wondering if there is a way to handle this with some kind of mapping of dateType to its corresponding dateType. If this works it works, but i feel like we have enough dates that come in pairs that we could find a less verbose way of handling.

server/src/model/graphql.ts
```diff
@@ -56,7 +56,16 @@ import { stateResolvers } from "./state/stateResolvers.js";
...
+import {
+  JSONObjectDefinition,
+  DateTimeTypeDefinition,
+  DateTypeDefinition,
+  DateTimeResolver,
+} from "graphql-scalars";
+
+const customScalarResolvers = {
+  DateTime: DateTimeResolver,
+};
```
connor-parke-icf, line 68: it looks like our custom resolver is just a map of the graphql-scalars one, is this intentional? if so, why not just pass in the graphql-scalars resolvers as they are?

server/src/model/bundleDate/validateInputDate.ts
TARehman: I mean, they all come in pairs by definition (all checks do). 😄 But we could have a data structure where we list the functions that need to be run for each pair. But how is that anything except a big switch statement?

server/src/model/graphql.ts
TARehman: You have to associate the scalar name with the resolver. So the DateTimeResolver is literally the one from `graphql-scalars`, we need to associate it with DateTime (like with our other custom scalars), so the object is needed.

PR #390, opened 2025-10-20
Demos-805: Add type linting to fix async issues in client and server

james-becwar-cms:
Client was having issues where some times the tests would randomly fail on CI.  I'm 99% sure this was due to a missing await or other race condition.  I enabled type linting on the client and found a bunch of issues with async and await.  Fixed them, and did the same for the Server.

Testing I did:
Lint passes in client and server
test pass in client and server
ran client and server local and clicked around.  Everything seems to be working.


client/eslint.config.js
```diff
@@ -9,7 +9,15 @@ import { noNonstandardDateFormatting } from "./eslint-rules/no-nonstandard-date-
...
-  ...tseslint.configs.recommended,
+  ...tseslint.configs.recommendedTypeChecked,
+  {
+    languageOptions: {
+      parserOptions: {
+        projectService: true,
+        allowDefaultProject: ["*.js", "*.mjs", "eslint.config.js"],
+      },
+    },
+  },
```
tresdonicf: I thought after we talked about this at office hours we determined these rules don't seem to be that helpful. I would certainly second that after seeing some of the changes that have been made after the introduction of these rules.

We should probably stick to `eslint` for our style-checking static analysis and `tsc` for our typechecking static analysis as it is now

tresdonicf (changes requested):
I'm not seeing many changes we might need on the `client/` side. The `await` in the signout function seems to be a positive change but otherwise it seems like we're sacrificing code quality to appease some overly obtuse / strict linting rules.

I think some of the instances of adding / removing `await / async` are fine but overall requesting that we remove the eslint typechecking rules and continue to use `tsc` for our typechecking. This would also mean removing the scattered `void` statements and passing in the functions directly as "first-class" citizens when it makes sense

client/package.json
```diff
@@ -62,7 +62,7 @@
     "dev:mocks": "VITE_USE_MOCKS=true npm run dev",
     "format": "prettier --write src",
     "format:check": "prettier --check src",
-    "lint": "eslint src; npm run tsc",
```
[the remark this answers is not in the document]
dustbuster: yeah please do not change this. SUPER HELPFUL script!

server/src/logger.ts
```diff
@@ -46,11 +46,16 @@ function formatValue(value: unknown): string {
...
+  if (typeof value === "object" && value !== null) {
+    try {
+      return JSON.stringify(value);
+    } catch {
+      return "[object]";
+    }
   }
+  if (typeof value === "function") return "[function]";
+  if (typeof value === "symbol") return value.toString();
+  return "[unknown]";
```
dustbuster: wut

connor-parke-icf (commented):
So some of these linter opinions i like, but some of them i disagree with. I like the promise resolution stuff, where we remove asyncs where not needed and explicitly await or resolve stuff where needed.  Im not a fan of wrapping functions as inputs into functions and the use of the void type; i feel like it kinda obscures the intention of the code a little bit. im also not a huge fan of the casting to strings in some places; i fear it could cover up some legitimate flaws in logic that should be explicitly handled.

Is there a way we could expand the current linter/tslint to include specific configurations to target only the async stuff rather than introducing a new library of linting?

tresdonicf:
> So some of these linter opinions i like, but some of them i disagree with. I like the promise resolution stuff, where we remove asyncs where not needed and explicitly await or resolve stuff where needed. Im not a fan of wrapping functions as inputs into functions and the use of the void type; i feel like it kinda obscures the intention of the code a little bit. im also not a huge fan of the casting to strings in some places; i fear it could cover up some legitimate flaws in logic that should be explicitly handled.
> 
> Is there a way we could expand the current linter/tslint to include specific configurations to target only the async stuff rather than introducing a new library of linting?

We should be able to select the rules we want rather than include all the ones that are in the list here and it seems the team is open to this based on our discussions. Generally I feel the same about the observations you mentioned here and I agree with things like "don't add async if not needed" but that's pretty tough to tell just in static analysis alone (we can't know just from the tokens if the functions called are / should be async) so it might not be the best tool for this sort of thing

PR #1055, opened 2026-04-08
DEMOS-1741: adding infrustructure for row-level permissions

connor-parke-icf:
This is the general framework setup for the row-based permission system which contains all the baseline interface that will be utilized for describing the permissions of each model in follow up PRs.


server/src/constants.ts
```diff
@@ -458,3 +458,12 @@ export const BUDGET_NEUTRALITY_VALIDATION_STATUSES = [
   "Pending",
   "In Progress",
 ];
+
+export const PERMISSIONS = [
```
connor-parke-icf, line 464: starting permissions which will be expanded and used in future PRs

server/src/prismaClient.ts
```diff
@@ -73,13 +76,39 @@ const createExtendedClient = () => {
...
+
+          const result = await context.findMany(args);
+          if (result.length < 1) {
+            return null;
+          }
+          if (result.length > 1) {
+            throw new Error(`Expected at most one record, but found ${result.length}`);
+          }
+          return result[0];
+        },
```
connor-parke-icf, line 106: because of the way prisma's findUnique works, it prevents the addition of filters that do not correspond to an exact unique key. This is annoying, since the permission filters generally route through a connection to the demonstration and the roles on the demonstration. So this method was made which essentially acts as a findUnique that can be used with constraints in general, returning null if not found and erroring if many found.

connor-parke-icf:
EDIT: updated with recent changes to auth.util, created simple migration for 3 common roles, updated seeder, initialized new users with corresponding system role

server/src/prismaClient.ts
TARehman: Honestly, I think you should put this explanation as a comment in the code itself as its important context to have as to why we're doing this at all.

TARehman (approved):
Looks good and consistent with what we've discussed before. There's more cleanup to be done in these files I think, but this is a good start.

server/src/prismaClient.ts
connor-parke-icf: not a bad idea; will do

PR #1062, opened 2026-04-09
DEMOS-1795 Phase 3 Declare Incomplete clears dates

nunesmlucas:
On Phase 3 (Completeness), clicking **Declare Incomplete** didn't clear the date pickers.

The server cleared the dates in the DB, but the client was auto-filling them again from the uploaded document's upload time.

Client-only change in `CompletenessPhase.tsx`: when the phase is `"Incomplete"`, skip the auto-fill and clear any date the user had typed in. No server changes, no new functions.

https://github.com/user-attachments/assets/8de16c29-45ac-439a-9190-eb27c98d8293

### DIFF
--- a/client/src/components/application/phases/CompletenessPhase.tsx
+++ b/client/src/components/application/phases/CompletenessPhase.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useMemo, useState } from "react";
+import React, { useMemo, useState } from "react";
 
 import { Button, SecondaryButton } from "components/button";
 import { ExportIcon } from "components/icons";
@@ -12,6 +12,7 @@ import {
   ApplicationDateInput,
   LocalDate,
   PhaseName,
+  PhaseStatus,
 } from "demos-server";
 import { useDialog } from "components/dialog/DialogContext";
 import { useSetApplicationDates } from "components/application/date/dateQueries";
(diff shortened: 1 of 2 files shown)


client/src/components/application/phases/CompletenessPhase.tsx
```diff
@@ -186,9 +191,16 @@ export const CompletenessPhase = ({
   const [userSelectedStateDeemedCompleteDate, setUserSelectedStateDeemedCompleteDate] =
     useState("");
 
+  useEffect(() => {
+    if (completenessIncomplete) {
+      setUserSelectedStateDeemedCompleteDate("");
+    }
+  }, [completenessIncomplete]);
```
tresdonicf: No need for a `useEffect` here. Since `completenessIncomplete` is a prop we can just calculate directly off of it if we need (and no need to have a prop changing in the dependency array as prop changes will re-render the component anyway)

Since this line is the same as the above (useState sets `userSelectedStateDeemedCompleteDate = ""`) I don't think it has any effect

client/src/components/application/phases/CompletenessPhase.tsx
```diff
@@ -139,6 +139,7 @@ export const getApplicationCompletenessFromApplication = (
       applicationIntakeComplete={applicationIntakePhase?.phaseStatus === "Completed"}
       completenessReviewDate={findDate("Completeness Review Due Date")}
       completenessComplete={completenessPhase?.phaseStatus === "Completed"}
+      completenessIncomplete={completenessPhase?.phaseStatus === "Incomplete"}
```
tresdonicf: When I was looking at this I thought it was weird at first that we didn't just carry `phaseStatus` into the component and use it however it makes sense. 

Not a blocker for now but that may be a worthwhile change, that way we can just check for any phase status we need now and in the future.

This approach becomes a bit odd because we could (in testing) pass both `completenessComplete = true` and `completenessIncomplete = true` which is an invalid application state

tresdonicf (approved):
Nothing to block on here, a couple of changes that might be helpful:

1. Use `PhaseStatus` directly instead of the computations for the phase being complete or incomplete
2. Remove the `useEffect` around the `setUserSelectedStateDeemedCompleteDate` (can just move it out of the hook as a prop-change always re-renders a component)

PR #1064, opened 2026-04-10
Demos 1571 - Deliverables default sort order - added fetch all by demo query

dustbuster:
I think we in good shape. 
No extension requested yet. 

And i wired in the query. 

Also, refactored my types to match like demo table 

Since i added the deliverables query in. this knocks out.
[DEMOS-1887](https://jiraent.cms.gov/browse/DEMOS-1887)
[DEMOS-1888](https://jiraent.cms.gov/browse/DEMOS-1888)
As well as [DEMOS-1571](https://jiraent.cms.gov/browse/DEMOS-1571)

Screenshot 2026-04-10 at 11 03 26


client/src/components/table/tables/DeliverableTable.tsx
```diff
@@ -12,18 +12,9 @@ import { DeleteIcon } from "components/icons/Action/DeleteIcon";
...
-  deliverableType: string;
-  cmsOwner: string;
-  dueDate: string;
-  submissionDate?: string;
-  status: string;
-  state: { id: string };
-};
+export type DeliverableTableRow = GenericDeliverableTableRow;
 
 export type DeliverableTableViewMode = Exclude<PersonType, "non-user-contact">;
```
dustbuster, line 19: @connor-parke-icf added a USER object that does not have this node in it. Which simplifies some things. SO update this to use that.

client/src/components/user/UserContext.tsx
```diff
@@ -62,3 +62,7 @@ export function getCurrentUser() {
 
   return ctx;
 }
+
+export function getCurrentUserOrUndefined() {
+  return useContext(Ctx);
+}
```
dustbuster: Since i need a user in a test. and it's outside of Provider. It's giving me grief. This solves it. Open to better solutions.

client/src/util/sortDeliverables.test.ts
tresdonicf, line 1: Would be nice to see a few more test cases in here for testing some of the other interactions of this sorting algorithm as I know it's a bit involved.

client/src/mock-data/DeliverablesMockTres.ts
```diff
@@ -0,0 +1,6 @@
+export {
+  MOCK_DELIVERABLES,
+  getDeliverablesForDemonstration,
+  MOCK_DELIVERABLE_1,
+  deliverableMocks,
+} from "./deliverableMocks";
```
nunesmlucas: What's this for? Nothing imports it, and the name `Tres` for Tresdon? lol. Should this be removed? maybe im missing something?

client/src/pages/DeliverablesPage.tsx
```diff
@@ -2,59 +2,79 @@ import { DeliverableTable } from "components/table/tables/DeliverableTable";
...
-  const [error] = React.useState<Error | undefined>(undefined);
-  const [data] = React.useState<DeliverablesPageQueryResult>({
-    deliverables: MOCK_DELIVERABLES,
-    currentUserId: "dustyrhodes",
-  });
-
-  // For when we get the backend working. Based on demo table
-  // const { data, loading, error } = useQuery<DeliverablesPageQueryResult>(
-  //   DELIVERABLES_PAGE_QUERY
-  // );
```
tresdonicf, line 53: NB: I see the comment here but looking closer at this structure I'm starting to think we want this logic in the `DeliverableTable` itself. Not a blocker, just thinking it through

client/src/pages/DeliverablesPage.tsx
```diff
@@ -2,59 +2,79 @@ import { DeliverableTable } from "components/table/tables/DeliverableTable";
...
-  deliverableType: string;
-  cmsOwner: string;
-  dueDate: string;
-  submissionDate?: string;
-  status: string;
-  extensionRequested?: boolean;
-  resubmissionCount?: number;
-  state: {
-    id: string;
+export type GenericDeliverableTableRow = Omit<
```
tresdonicf: NB: What's `Generic` about this? Since this type refers to the type expected by the table we probably want to export it from there and just use it in this case (instead of the current way of aliasing `DeliverableTableRow` to `GenericDeliverableTableRow`, probably can drop the "Generic" all together)

client/src/pages/DeliverablesPage.test.tsx
```diff
@@ -7,6 +7,14 @@ import { DeliverablesPage } from "./DeliverablesPage";
...
 import { mockUsers } from "mock-data/userMocks";
 
+const mockUseQuery = vi.fn();
+
+vi.mock("@apollo/client", () => ({
+  gql: (literals: TemplateStringsArray, ...placeholders: string[]) =>
+    literals.reduce((acc, curr, idx) => acc + curr + (placeholders[idx] || ""), ""),
+  useQuery: (...args: unknown[]) => mockUseQuery(...args),
+}));
+
```
nunesmlucas: NIT: Can we use `MockedProvider` with the `deliverableMocks` array here instead of mocking the whole @apollo/client module? since you already added a `DELIVERABLES_PAGE_QUERY` response to `deliverableMocks` in this PR. Having two different ways to mock Apollo might get confusing

client/src/pages/DeliverablesPage.tsx
```diff
@@ -2,59 +2,79 @@ import { DeliverableTable } from "components/table/tables/DeliverableTable";
...
 };
 
 type DeliverablesPageQueryResult = {
-  deliverables: Deliverable[];
-  currentUserId: string;
+  deliverables: GenericDeliverableTableRow[];
 };
 type DeliverableTableViewMode = Exclude<PersonType, "non-user-contact">;
 
+export const DELIVERABLES_PAGE_QUERY = gql`
```
tresdonicf, line 30: NB: This is quite a tall query, do we need all the fields here? For this reason alone the hooks pattern would be a nice QOL improvement

client/src/pages/DeliverablesPage.test.tsx
```diff
@@ -22,8 +30,18 @@ describe("DeliverablesPage tab persistence", () => {
...
+      },
+      loading: false,
+      error: undefined,
+    });
     mockGetCurrentUser.mockReturnValue({
-      currentUser: mockUsers[0],
+      currentUser: {
+        ...mockUsers[0],
+        id: CURRENT_USER_ID,
+      },
```
tresdonicf: This isn't quite what we usually do here - we want to leverage the apollo `<MockedProvider [...]>` either through that wrapper directly or `<TestProvider mocks=[...]>`.

The latter will actually provide all the mocks from `src/mock-data/index.ts` by default so you can declare entities there to use in tests if you want. The big difference is that it needs to match a specific query name (not my favorite thing)

client/src/pages/DemonstrationDetail/deliverables/DeliverablesTab.tsx
```diff
@@ -14,9 +16,11 @@ export const DeliverablesTab = ({
   deliverables,
 }: {
   parentDemonstration: AddDeliverableSlotDemonstration;
-  deliverables: Deliverable[];
+  deliverables: GenericDeliverableTableRow[];
 }) => {
   const { showAddDeliverableSlotDialog } = useDialog();
+  const rawPersonType = getCurrentUserOrUndefined()?.currentUser?.person.personType;
+  const viewMode = (rawPersonType ?? "demos-cms-user") as DeliverableTableViewMode;
```
tresdonicf: Defaulting to `demos-cms-user` if this value isn't found seems a bit dangerous, am I missing something here?

client/src/components/user/UserContext.tsx
tresdonicf: I think that this might be fixed with `<TestProvider>` via the fact it uses `userMocks` which returns a response for the `getCurrentUser` query. You might need to add `maxUsageCount: Number.POSITIVE_INFINITY` if you get a "No more matching queries for ..." type message

client/src/components/table/tables/DeliverableTable.tsx
```diff
@@ -63,10 +35,14 @@ export const DeliverableTable: React.FC<{
...
-    status: formatDeliverableStatus(deliverable),
-  }));
+  const formattedDeliverables = React.useMemo(
+    () =>
+      sortDeliverablesByDefault(deliverables).map((deliverable) => ({
+        ...deliverable,
+        status: formatDeliverableStatus(deliverable),
+      })),
+    [deliverables]
+  );
```
tresdonicf: No need to specify a `useMemo` here. Also since the only thing in the dependency array is `[deliverables]` I don't think this will have any effect (a component is always re-rendered on prop changes)

client/src/components/user/UserContext.tsx
dustbuster: NUKED THIS.

client/src/mock-data/DeliverablesMockTres.ts
dustbuster: This is an accidental commiting!

client/src/components/table/tables/DeliverableTable.tsx
```diff
@@ -12,18 +12,9 @@ import { DeleteIcon } from "components/icons/Action/DeleteIcon";
...
-  deliverableName: string;
-  demonstrationName: string;
-  deliverableType: string;
-  cmsOwner: string;
-  dueDate: string;
-  submissionDate?: string;
-  status: string;
-  state: { id: string };
-};
+export type DeliverableTableRow = GenericDeliverableTableRow;
```
tresdonicf: I think the model should probably be defined here since it's the main component that cares about this data shape

client/src/mock-data/DeliverablesMockTres.ts
dustbuster: NUKED

client/src/pages/DeliverablesPage.test.tsx
dustbuster: yea man. I did not even realize this was here.

tresdonicf (commented):
I think I got to everything, in the future it would be appreciated to keep the PRs separate for different functionality as it makes it a bit tougher to review the unrelated changes together. Mostly looking good, a couple of things that I'm wondering about before approving:

1. The `viewMode` defaulting to `demos-cms-user` if it can't be found - this sounds like it might leak some information compared to defaulting to a lower permission level. Maybe we use `demos-state-user` here instead or just throw an error because this field really should exist on the user object I imagine
2. Try out `<TestProvider>` for the mocking - should help with getting a fake user in tests as well! A few more tests for the default sort order would be well-found here as well

client/src/pages/DemonstrationDetail/deliverables/DeliverablesTab.tsx
tresdonicf: This is outdated now, just uses the value directly, nice job!

client/src/pages/DeliverablesPage.test.tsx
dustbuster: Totally sorted.

server/src/seeder.ts
```diff
@@ -135,6 +138,109 @@ async function seedTagsAndStatuses() {
...
+      },
+    },
+    select: {
+      id: true,
+      sdgDivisionId: true,
+    },
+  });
+  if (!approvedDemonstration) {
+    throw new Error("No demonstrations with demonstration types found for approved-demo seeding");
+  }
```
dustbuster, line 156: NOTE: this is not a complete demo, but its enough for deliverables.

client/src/components/table/columns/DemonstrationColumns.tsx
```diff
@@ -3,7 +3,7 @@ import { highlightCell } from "../KeywordSearch";
 import { SecondaryButton } from "components/button";
 import { ChevronDownIcon, ChevronRightIcon } from "components/icons";
 import React from "react";
-import { GenericDemonstrationTableRow } from "../tables/DemonstrationTable";
```
dustbuster, line 6: Since I got Generic From demos, then let's just ditch this generic naming structure alltogether!

client/src/pages/DeliverablesPage.test.tsx
```diff
@@ -19,17 +20,38 @@ describe("DeliverablesPage tab persistence", () => {
...
+      request: {
+        ...deliverableMocks[0]!.request,
+        query: DELIVERABLES_PAGE_QUERY,
+      },
+    },
+  ];
+
+  const renderDeliverablesPage = async () => {
+    render(
+      <TestProvider mocks={deliverablesPageMocks} addTypename={false}>
```
tresdonicf, line 35: NB: Could add `userMocks` to this array for getCurrentUser

client/src/pages/DemonstrationDetail/DemonstrationTab.test.tsx
tresdonicf, line 1: NB: If the `DemonstrationTab` itself wasn't touched I'm a bit confused about this changing

tresdonicf (approved):
Nothing to block on here! The only thing I think you might wanna take a look at is mocking the user context directly. We can keep this the same as the rest of the apollo client mocks (using `MockedProvider` or `TestProvider`) and we just need to make sure to add `userMocks` to the mock array in order to support this, could save some complexity but it's not a huge deal, nice job with this!

client/src/pages/DemonstrationDetail/DemonstrationTab.test.tsx
dustbuster: Good question. Resetting gives me this, so let me dig in a bit. 

```
 FAIL  src/pages/DemonstrationDetail/DemonstrationTab.test.tsx > DemonstrationTab > Deliverables tab conditional rendering > renders Deliverables tab when status is "Approved"
 FAIL  src/pages/DemonstrationDetail/DemonstrationTab.test.tsx > DemonstrationTab > Deliverables tab conditional rendering > defaults to Deliverables tab when status is "Approved"
Error: getCurrentUser must be used within <UserProvider>
 ❯ getCurrentUser src/components/user/UserContext.tsx:60:11
     58|
     59|   if (!ctx) {
     60|     throw new Error("getCurrentUser must be used within <UserProvider>");
       |           ^
     61|   }
     62|
```

dustbuster: DeliverablesTab now calls getCurrentUser() to set viewMode (DeliverablesTab.tsx).

In DemonstrationTab.test, the "Approved" cases render DeliverablesTab.

Without a UserProvider or a mock for getCurrentUser, those tests throw: getCurrentUser must be used within <UserProvider>.

So the test was updated to mock components/user/UserContext and provide mockGetCurrentUser in beforeEach.

Also, the Deliverables tab label assertion was loosened from a hardcoded count to `/Deliverables \(\d+\)/` because count depends on mock data and had become brittle.

client/src/util/sortDeliverables.test.ts
dustbuster: this is done

PR #1067, opened 2026-04-10
DEMOS-1771: Close ConfirmApprovalDialog Upon Clicking "Approve" Button

tresdonicf:
Fix here was to utilize the `<BaseDialog>` component that handles this behavior already.

Good news is that the new `AGENTS.md` file would help prevent things like this in the future by making it clear to re-use this component

https://github.com/user-attachments/assets/e9f4d8cd-9e26-4a05-9f86-891c237f7143

### DIFF
--- a/client/src/components/dialog/DialogContext.tsx
+++ b/client/src/components/dialog/DialogContext.tsx
@@ -225,11 +225,7 @@ export const useDialog = () => {
     applicationType: WorkflowApplicationType
   ) => {
     context.showDialog(
-      <ConfirmApproveDialog
-        onClose={context.hideDialog}
-        onConfirm={onConfirm}
-        applicationType={applicationType}
-      />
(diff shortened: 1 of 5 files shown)


client/src/components/dialog/DialogContext.tsx
```diff
@@ -225,11 +225,7 @@ export const useDialog = () => {
     applicationType: WorkflowApplicationType
   ) => {
     context.showDialog(
-      <ConfirmApproveDialog
-        onClose={context.hideDialog}
-        onConfirm={onConfirm}
-        applicationType={applicationType}
-      />
+      <ConfirmApproveDialog onConfirm={onConfirm} applicationType={applicationType} />
```
connor-parke-icf, line 228: nice! i like the omission of onClose and opting to let useDialog provide closeDialog in the dialog itself

connor-parke-icf (approved):
no notes! great that we can offload a bunch of this to the baseDialog.

PR #1070, opened 2026-04-10
Demos 1741 refactor auth utils

connor-parke-icf:
The PR i have for getting the architecture of the row-based authorization started is currently getting blocked because of code coverage as there was some changes made in the authorization functions (to make context.user non nullable, and to inject the user permissions there).  I took the opportunity to try and simplify some of the code here so that the parts could be more effectively covered by tests.  

There were a lot of helper functions and utilities here that seemed to be made ambivalent to whether they were called from the context of server.ts (our lambda handler) and local-server.ts (our local server).  As such, they needed a lot of handling for the different auth data each env would provide. One of the main changes here was to move the env-specific functionalility/mappings from the shared resource and into the individual server files. 

Some scenarios were being supported that didnt have a use case; for instance, if the server lambda was triggered without authorization context in its triggering event, it would fallback on the local-server cookie-based auth. That fallback has been removed. 

Functionalility was broken up into smaller packages: userContext handles finding/creating the user, decodeToken handles decoding the auth token found in the authorization cookie, a dedicated Claim validator function was created to centralize relevant checks and improve type safety. 

etc


server/src/auth/auth.config.ts
connor-parke-icf, line 1: this config was only used to decode the token, so moved it to that dedicated file

server/src/auth/auth.util.ts
```diff
@@ -1,357 +1,48 @@
...
-export function verifyRole(role: string): void {
-  const validRoles = (PERSON_TYPES as readonly string[]).filter(r => r !== "non-user-contact");
-  if (!validRoles.includes(role)) {
-    throw new GraphQLError(`Invalid user role: '${role}'`, {
-      extensions: {
-        code: "UNAUTHORIZED",
-        http: { status: 403 },
-      },
-    });
+export function validateClaims(
```
connor-parke-icf, line 18: validates the AuthClaims, centralizing all the relevant checks to a function with type assertion.

server/src/auth/decodeToken.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export function decodeToken(token: string): Promise<JwtPayload> {
+  const verifyOpts: VerifyOptions = {
+    audience: JWKS_AUDIENCE,
+    issuer: JWKS_ISSUER,
+    algorithms: ["RS256"],
+  };
+
+  return new Promise((resolve, reject) => {
+    function getKey(header: JwtHeader, cb: (err: Error | null, key?: string) => void) {
```
connor-parke-icf, line 30: this function used to decode the token then map it to a common Claims shape. i moved that mapping out of here, and into the local-server file, since its only applicable there.

server/src/auth/userContext.ts
```diff
@@ -0,0 +1,52 @@
+import { prisma } from "../prismaClient";
+import { UserType } from "../types";
+import { AuthorizationClaims } from "./auth.util";
+
+export type ContextUser = {
+  id: string;
+  cognitoSubject: string;
+  personType: UserType;
+};
```
connor-parke-icf, line 9: this used to be "id" "sub" "role", matching the claims; but since this is used moreso in the application layer, i modified it to reflect the field in the db and gql. 

This is where "permissions": Permission[] will go

server/src/model/event/eventResolvers.ts
```diff
@@ -2,14 +2,14 @@ import { prisma } from "../../prismaClient.js";
...
-import { GraphQLContext, getCurrentUserId } from "../../auth/auth.util.js";
+import { GraphQLContext } from "../../auth/auth.util.js";
 
 export async function logEvent(
   parent: unknown,
   { input }: { input: LogEventInput },
   context: GraphQLContext
 ) {
-  const userId = await getCurrentUserId(context);
+  const userId = context.user.id;
```
connor-parke-icf, line 12: because context.user is now non nullable, this could be updated to just grab it from the context; no helper needed

server/src/constants.ts
```diff
@@ -103,6 +103,8 @@ export const PERSON_TYPES = [
   "non-user-contact",
 ] as const;
 
+export const USER_TYPES = ["demos-admin", "demos-cms-user", "demos-state-user"] as const;
```
connor-parke-icf, line 106: found myself making Exclude<PersonType, "non-user-contact"> too many times, so just went ahead and made a const/type for it. May be able to write this to derive from PERSON_TYPES but eh...

server/src/local-server.ts
```diff
@@ -19,19 +27,42 @@ const server = new ApolloServer<GraphQLContext>({
...
   resolvers,
   introspection: process.env.ALLOW_INTROSPECTION === "true",
-  ...protection,  
+  ...protection,
   plugins: [...protection.plugins, authGatePlugin, gatedLandingPagePlugin(), loggingPlugin],
-  validationRules: [...protection.validationRules]
+  validationRules: [...protection.validationRules],
 });
 
+export function extractClaimsFromDecodedToken(decodedToken: JwtPayload): AuthorizationClaims {
```
connor-parke-icf, line 35: each server file now is responsible for decoding the tokens in the format they expect instead of relying on a common help. this works because the local-server gets its auth data from a cookie, which has slightly differently formatted data than the lambda server.  once decoded into a common format it goes to the common pathway

server/src/server.ts
```diff
@@ -1,83 +1,76 @@
...
-    sub,
-    email,
-    role,
-    givenName,
-    familyName,
-    identities,
-    "cognito:username": cognitoUsername,
-  } as JwtClaims & Record<string, unknown>;
-}
+  const secretsManager = new SecretsManagerClient({ region: process.env.AWS_REGION });
```
connor-parke-icf, line 26: could probably move the secrets manager stuff to its own file; but its only applicable to the lambda server. could be convinced either way

server/package.json
```diff
@@ -18,6 +18,7 @@
     "@escape.tech/graphql-armor": "^3.2.0",
     "@prisma/adapter-pg": "^7.6.0",
     "@prisma/client": "^7.6.0",
+    "cookie": "^1.1.1",
```
connor-parke-icf, line 21: using a library for handling cookies rather than handling string manipulation

tresdonicf (approved):
Definitely seems a lot simpler and over half the lines removed, nice job!

server/src/auth/auth.util.ts
```diff
@@ -1,357 +1,48 @@
...
+import { ContextUser, findOrCreateContextUserFromClaims } from "./userContext.js";
+import { UserType } from "../types.js";
 
 export interface GraphQLContext {
-  user: null | {
-    id: string;
-    sub: string;
-    role: string;
-  };
+  user: ContextUser;
```
dustbuster, line 6: Niiiice!

server/src/auth/auth.util.ts
```diff
@@ -1,357 +1,48 @@
...
-};
-
-const jwks = jwksClient({
-  jwksUri: config.jwksUri,
-  cache: true,
-  cacheMaxEntries: CACHE_MAX_ENTRIES,
-  cacheMaxAge: CACHE_MAX_AGE,
-  rateLimit: true,
-  jwksRequestsPerMinute: 10,
-});
```
dustbuster, line 51: You get a type and you get a type, look under your chairs! WE ALL GET TYPES!

connor-parke-icf:
EDIT: went through and tested on ephemeral env to validate everything, and found that we needed some adjustments to make sure that the EUA was actually getting passed to the server. Updated the authorization lambda to handle this with guidance from Jesse.  Now for username it uses the EUA and defaults to user's email should it not exist (in the case of logging in directly via cognito. Tested all 8 scenarios: 
* fresh user on lambda with IDM
* existing user on lambda with IDM
* fresh user on lambda with cognito
* existing user on lambda with cognito
* fresh user on local with IDM
* existing user on local with IDM
* fresh user on local with Cognito
* existing user on local with Cognito

server/src/model/event/eventResolvers.ts
dustbuster: If it's hydrated than easy validation, if it splodes then we get immediate feedback. so Word!

server/src/constants.ts
dustbuster: yeah. I can use this on state user deliverable differences. It's dumb cuz by defautl a user will never be the other type.

server/src/local-server.ts
dustbuster: Geez! Annoying much?

server/src/local-server.ts
```diff
@@ -19,19 +27,40 @@ const server = new ApolloServer<GraphQLContext>({
   typeDefs,
   resolvers,
   introspection: process.env.ALLOW_INTROSPECTION === "true",
-  ...protection,
```
dustbuster, line 22: BLAST! I left trim white spaces ON for this project!

server/src/local-server.ts
```diff
@@ -19,19 +27,40 @@ const server = new ApolloServer<GraphQLContext>({
...
 
+export function extractClaimsFromDecodedToken(decodedToken: JwtPayload): AuthorizationClaims {
+  const externalUserId = decodedToken.identities?.[0]?.userId ?? decodedToken.email;
+
+  const claims = {
+    email: decodedToken.email,
+    sub: decodedToken.sub,
+    role: decodedToken["custom:roles"],
+    givenName: decodedToken.given_name,
+    familyName: decodedToken.family_name,
```
dustbuster, line 43: Like i get why they'd use given And family. But Geez it's always like a 2 second brain block as i figure out which one is which.

server/src/local-server.ts
```diff
@@ -19,19 +27,40 @@ const server = new ApolloServer<GraphQLContext>({
...
+      if (!token) {
+        throw new Error("Missing Authorization header with Bearer token");
+      }
 
+      const decodedToken = await decodeToken(token);
+      const claims = extractClaimsFromDecodedToken(decodedToken);
+      const ctx = await buildContextFromClaims(claims);
+
+      const requestId = (req.headers["x-request-id"] as string | undefined) || randomUUID();
       const additionalContext = {
```
dustbuster, line 64: Still can't just be simple can it?  Always some cookies `x-request-id` making things weird.

server/src/seeder.ts
```diff
@@ -804,7 +805,7 @@ async function seedDatabase() {
     // ~60% of events have a applicationId, rest are null
```
dustbuster, line 805: Image

server/src/seeder.ts
```diff
@@ -804,7 +805,7 @@ async function seedDatabase() {
...
     const attachApplication = Math.random() < 0.6;
     const maybeApplication = attachApplication ? (pick(applicationsForEvents)?.id ?? null) : null;
-    const user = pick(usersForEvents) ?? null;
```
dustbuster, line 807: mmmm hmmm. Which we'd had the forethought to let things fail earlier. might have saved us some bug hunting.

server/src/seeder.ts
```diff
@@ -856,20 +857,9 @@ async function seedDatabase() {
...
-        user: {
-          id: user.id,
-          sub: user.cognitoSubject,
-          role: user.personTypeId,
-        },
-      };
-    }
+    const context = {
+      user,
+    } as GraphQLContext;
```
dustbuster, line 862: 🔥

server/src/server.ts
```diff
@@ -1,83 +1,75 @@
...
+  GraphQLContext,
+  buildContextFromClaims,
+  validateClaims,
+} from "./auth/auth.util.js";
 import { log, reqIdChild, als, store } from "./log.js";
+import type { APIGatewayProxyEvent } from "aws-lambda";
+import { GetSecretValueCommand, SecretsManagerClient } from "@aws-sdk/client-secrets-manager";
 
-import type { APIGatewayProxyEvent, APIGatewayProxyEventHeaders } from "aws-lambda";
+log.info({ type: "graphql.startup.loaded" });
```
dustbuster, line 19: Initially, Jesse requested the ability to see what's going on in cloudwatch logs. SO if you can see any other opportunities for smart logging, then by all means DO ETT

server/src/server.ts
dustbuster: Its really not that much. Already in a lambda only context. Leave it. Easier to read.

server/src/server.ts
```diff
@@ -1,83 +1,75 @@
...
-export function withAuthorizerHeader(
-  headers: APIGatewayProxyEventHeaders,
-  claims: JwtClaims | null
-): APIGatewayProxyEventHeaders {
-  return claims
-    ? { ...headers, "x-authorizer-claims": JSON.stringify(claims) }
-    : headers;
+  if (!response.SecretString) throw new Error("The SecretString value is undefined!");
+  const s = JSON.parse(response.SecretString);
+  return `postgresql://${s.username}:${s.password}@${s.host}:${s.port}/${s.dbname}?schema=demos_app`;
```
dustbuster, line 31: okay, i can see it in another file now. LOL
See what i did here: 

`lambdas/UIPath/db.ts`
And 
`lambdas/budgetNeutrality/db.ts`

Here. I stared to set up a convention, but I'll leave this up to you, not a blocker.

server/src/server.ts
```diff
@@ -1,83 +1,75 @@
...
 }
 
 const setDatabaseUrl = async () => {
-  const url = await getDatabaseUrl()
-  log.debug({type: "graphql.db.creds_retrieved"});
+  const url = await getDatabaseUrl();
+  log.debug({ type: "graphql.db.creds_retrieved" });
```
dustbuster, line 36: Space is for cool people.

server/package.json
dustbuster: 🍪 🧑‍🍳

dustbuster: FYI super popular. (thus hoping that translates to being well supported)

Image

dustbuster (approved):
I do like the idea of having db.ts files to manage the database connections. that being said. We can come back to that. Not a deal break. Server it kinda it's own thing.

dustbuster:
>One of the main changes here was to move the env-specific functionalility/mappings from the shared resource and into the individual server files.


Sonarcube made me do it!

server/src/server.ts
connor-parke-icf: yeah, i think itd be good to have a common set of SecretsManager/db stuff. Debated moving to a helper; totally forgot that the lambdas do a lot of similar stuff otherwise i wouldve debated harder lol.  Might make a good backlog item to centralize helpers for DB secrets/connections.

PR #1071, opened 2026-04-10
DEMOS-1878: Show "(Unapproved)" text next to demonstration types in the AddDeliverableSlotDialog

tresdonicf:
Adds a `(Unapproved)` text next to the demonstration type tags in the `AddDeliverableSlotDialog`

https://github.com/user-attachments/assets/c26b6c68-120f-48f7-b3da-f87281399eaf


connor-parke-icf (approved):
seems simple enough:
* make the select take an array of Tags with id and approval status instead of just string
* render the unapproved as such
* bubble up the typing change through the calling components until you hit the mapping in DemonstrationTab

LGTM, seems very similar to the other approved/unapproved changes. Its shame that each of these places use a different input method (single select, multiselect, apply tags dialog widget), otherwise they could all be one component

PR #1073, opened 2026-04-11
Chore add esbuild

james-becwar-cms:


.devcontainer/localstack/setup_localstack.sh
dustbuster, line 8: Um. i don't think we should add this here. 
if you try and build something without this installed, you do get feedback that esbuild is not installed. 

This prolly needs to be step 1 in probably a `localstack.md`
I ran it just to see what would happen if already installed. 

```bash
➜  client git:(DEMOS-1571) ✗ npm install -g esbuild

changed 2 packages in 1s
npm notice
npm notice New minor version of npm available! 11.6.2 -> 11.12.1
npm notice Changelog: https://github.com/npm/cli/releases/tag/v11.12.1
npm notice To update run: npm install -g npm@11.12.1
npm notice
```

I supposed there's not a lot of issues trying to install it again. but i do not like apps installing things outside of their webroot / Purview.

dustbuster:
Can you add a description @james-becwar-cms?

PR #1074, opened 2026-04-13
DEMOS-1735-create-deliverable-mutator

TARehman:
This PR adds the mutator to create new deliverables along with the validation and checking of the input payload. It also tweaks some small server settings to eliminate some local warnings, and makes a few small changes elsewhere.

- Added createDeliverable functions, resolvers, and associated tests. This includes parsing and validating the inputs.
- Set the default value for some fields in Prisma to be UUID. This doesn't change the schema at all as its entirely internal to Prisma, but it makes generating new records easier.
- Change the getCurrentUserId function to be sync rather than async. It may be removed or refactored later but for now it can just be sync. This required changing it in a few other places as well.
- I changed the `handlePrismaError` function to return a list of `originalMessages` rather than a single `originalMessage`. That is because later I started returning more than one error from the API for the `createDeliverable` validator and I wanted to be consistent in naming.
- Added checking for end of day dates for due dates to the date validator.


server/src/model/deliverable/deliverableSchema.ts
```diff
@@ -1,17 +1,21 @@
...
+  NonEmptyString,
+  TagName,
   User,
 } from "../../types";
 
 export const deliverableSchema = gql`
   type Deliverable {
     id: ID!
     deliverableType: DeliverableType!
+    name: NonEmptyString!
```
dustbuster, line 18: i mean do we really need the name? lol

TARehman: I was terminally database-brained on that table, literally only included the fields that were most relevant to other tables. 🤣

server/src/model/event/eventResolvers.ts
```diff
@@ -9,7 +9,7 @@ export async function logEvent(
   { input }: { input: LogEventInput },
   context: GraphQLContext
 ) {
-  const userId = await getCurrentUserId(context);
+  const userId = getCurrentUserId(context);
```
dustbuster: Best method ever

dustbuster (approved):
Don't really see any issues where. Does `index()` and `show()` work?
(show all or most and show one)

If those fetches work that's what will unblock most of my integration tickets.

server/src/auth/auth.util.ts
```diff
@@ -331,7 +331,7 @@ export const getCurrentUserRoleId = async (context: GraphQLContext): Promise<str
   return user.personTypeId;
 };
 
-export const getCurrentUserId = async (context: GraphQLContext): Promise<string> => {
+export const getCurrentUserId = (context: GraphQLContext): string => {
```
connor-parke-icf: no biggie, but these were removed in my recent PR. user will be non-nullable; if theres a context, it has a user!

server/src/model/deliverable/deliverable.prisma
```diff
@@ -1,5 +1,5 @@
 model Deliverable {
-  id                    String   @db.Uuid
+  id                    String   @db.Uuid @default(uuid())
```
connor-parke-icf, line 2: nice catch

server/src/model/deliverable/createDeliverable.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    await insertDeliverableAction(
+      {
+        deliverableId: newDeliverable.id,
+        actionType: "Created Deliverable Slot",
+        actionTime: actionTime,
+        oldStatus: "Upcoming",
+        newStatus: "Upcoming",
+        oldDueDate: parsedInput.dueDate.easternTZDate,
+        newDueDate: parsedInput.dueDate.easternTZDate,
```
connor-parke-icf, line 39: looks like a similar pattern to our applicationDates

server/src/model/deliverable/deliverableTypes.ts
```diff
@@ -0,0 +1,11 @@
+import { EasternTZDate } from "../../dateUtilities";
+import { DeliverableType, NonEmptyString, TagName } from "../../types";
+
+export type ParsedCreateDeliverableInput = {
```
connor-parke-icf, line 4: think id rather this type be attached to parseCreateDeliverableInput rather than in this generic Types file, but no biggie

server/src/model/deliverable/deliverableTypes.ts
```diff
@@ -0,0 +1,11 @@
...
+  name: NonEmptyString;
+  deliverableType: DeliverableType;
+  demonstrationId: string;
+  cmsOwnerUserId: string;
+  dueDate: EasternTZDate;
+  demonstrationTypes?: TagName[];
```
connor-parke-icf, line 10: for the create, should we explicitly make the frontend pass in an empty array or be able to simply omit? i think i would prefer requiring an empty array but im not sure why.

server/src/model/deliverable/parseCreateDeliverableInput.ts
```diff
@@ -0,0 +1,19 @@
...
+  const parsedDueDate = parseDateTimeOrLocalDateToEasternTZDate(input.dueDate, "End of Day");
+  checkInputDateIsEndOfDay("dueDate", parsedDueDate);
+  return {
+    name: input.name,
+    deliverableType: input.deliverableType,
+    demonstrationId: input.demonstrationId,
+    cmsOwnerUserId: input.cmsOwnerUserId,
+    dueDate: parsedDueDate,
+    demonstrationTypes: input.demonstrationTypes,
+  };
```
connor-parke-icf, line 18: NB, i like the spread operator here since it more clearly convays that only one value is changing, but its doesnt matter. i can see merit to explicitly confining the type to these fields regardless of changes to input. 
```
{
    ...input,
    dueDate: parsedDueDate
}
```

server/src/model/deliverable/validateDeliverableInputs.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  const errors = [];
+
+  if (demonstrationStatus !== "Approved") {
+    errors.push(
+      `Demonstration ${demonstration.id} is not in Approved status; cannot create deliverable.`
+    );
+  }
+
+  const permittedOwnerPersonTypes: PersonType[] = ["demos-admin", "demos-cms-user"];
```
connor-parke-icf, line 40: discussed this; this is not a permission but rather a fact of the database constraining so its fine to leave like this

server/src/model/deliverableAction/deliverableActionTypes.ts
```diff
@@ -0,0 +1,13 @@
+import { DeliverableActionType, DeliverableStatus } from "../../types";
+
+export type DeliverableAction = {
```
connor-parke-icf, line 3: do we want a GQL type for this if we are going to be returning this in some form to the frontend for display? if not, i imagine its included in thew other upcoming "action" tickets we discussed

server/src/model/deliverableDemonstrationType/queries/insertDeliverableDemonstrationTypes.ts
```diff
@@ -0,0 +1,22 @@
...
+import { prisma, PrismaTransactionClient } from "../../../prismaClient.js";
+import { SetDeliverableDemonstrationTypesInput } from "../../../types";
+import { getDeliverable } from "../../deliverable";
+
+export async function insertDeliverableDemonstrationTypes(
+  input: SetDeliverableDemonstrationTypesInput,
+  tx?: PrismaTransactionClient
+): Promise<PrismaDeliverable> {
+  const prismaClient = tx ?? prisma();
+  if (input.demonstrationTypes.length > 0) {
```
connor-parke-icf, line 11: i get why this is done this way, since we are deleting all then creating. But instead of passing in the input, maybe itd be better for this to just take an array of demonstrationTypes so its less contextual of the input and is more purely an insert operation.

server/src/model/deliverableDemonstrationType/setDeliverableDemonstrationTypes.ts
```diff
@@ -0,0 +1,18 @@
...
+  input: SetDeliverableDemonstrationTypesInput,
+  tx?: PrismaTransactionClient
+): Promise<PrismaDeliverable> {
+  const prismaClient = tx ?? prisma();
+  await deleteAllDeliverableDemonstrationTypes(
+    { deliverableId: input.deliverableId },
+    prismaClient
+  );
+  await insertDeliverableDemonstrationTypes(input, prismaClient);
+  return getDeliverable({ id: input.deliverableId }, prismaClient);
```
connor-parke-icf, line 17: discussed: if the constraint of "deliverable of type x needs at least one demonstration type" is going to be deferred, it should still be fine to delete and insert so long as it is in a transaction

server/src/model/event/eventResolvers.ts
connor-parke-icf: heh... sooo now the user object will always be on the context, so this function will likely no longer be needed and we can just do context.user.id

server/tsconfig.json
```diff
@@ -2,10 +2,10 @@
   "compilerOptions": {
     "outDir": "dist",
     "rootDir": "src",
-    "lib": ["es2020"],
+    "lib": ["es2021"],
     "target": "es2021",
     "module": "esnext",
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
```
connor-parke-icf, line 8: this feels out of place, but i remember some kinda discussion about it so just highlighting it

server/src/model/deliverable/createDeliverable.test.ts
```diff
@@ -0,0 +1,163 @@
...
+  // Mock transaction
+  const mockTransaction: any = "Test!";
+  const mockPrismaClient = {
+    $transaction: vi.fn(),
+  };
+
+  beforeEach(() => {
+    vi.resetAllMocks();
+    vi.useFakeTimers();
+    vi.setSystemTime(mockCurrentDate);
```
connor-parke-icf, line 84: 👀 good to know this exists....... does this do timezones??

server/src/model/deliverable/createDeliverable.test.ts
```diff
@@ -0,0 +1,163 @@
...
+      mockTransaction
+    );
+  });
+
+  it("should call the insert function with a transaction", async () => {
+    await createDeliverable(testInput, testContext);
+    expect(insertDeliverable).toHaveBeenCalledExactlyOnceWith(mockParsedInput, mockTransaction);
+  });
+
+  it("should use an empty list if no demonstration types are passed", async () => {
```
connor-parke-icf, line 113: feel like we should explicily require the demonstration types array for creation

server/src/model/deliverable/deliverableResolvers.test.ts
```diff
@@ -95,7 +108,7 @@ describe("deliverableResolvers", () => {
 
   describe("resolveDeliverable", () => {
     it("should throw if given something not supported", async () => {
-      await expect(() =>
+      await expect(
```
connor-parke-icf, line 111: huh; was this a change with the latest vitest or just something that was unneeded but still valid before?

server/src/model/deliverable/validateDeliverableInputs.test.ts
```diff
@@ -0,0 +1,293 @@
...
+      personTypeId: "demos-admin" satisfies PersonType,
+    };
+    const mockCmsUser = {
+      ...mockUser,
+      personTypeId: "demos-cms-user" satisfies PersonType,
+    };
+    const mockStateUser = {
+      ...mockUser,
+      personTypeId: "demos-state-user" satisfies PersonType,
+    };
```
connor-parke-icf, line 82: NB: maybe a generator function would be better, but not enough for it to save much ink

server/src/model/deliverableDemonstrationType/queries/insertDeliverableDemonstrationTypes.test.ts
```diff
@@ -0,0 +1,114 @@
...
+    expect(
+      transactionMocks.deliverableDemonstrationType.createMany
+    ).toHaveBeenCalledExactlyOnceWith(expectedRegularCall);
+    expect(getDeliverable).toHaveBeenCalledExactlyOnceWith(
+      { id: regularTestInput.deliverableId },
+      mockTransaction
+    );
+  });
+
+  it("should skip inserting if no values are given", async () => {
```
connor-parke-icf, line 105: feels a little odd to call the insert method with nothing to insert, but theres no way to validate nonempty array in typescript so i think this is fine.

connor-parke-icf (approved):
lots of comments/nitpicks but nothing blocking! big block of code, nice work

server/src/model/deliverable/deliverableTypes.ts
TARehman: I could move them to the generic parseDeliverableInput file that I'm now working on for updateDeliverable.

server/src/model/deliverable/deliverableTypes.ts
TARehman: I don't think it matters really, but I think these being optional is going to be important because if you don't want to make any changes you either have to pass the same list that currently exists, or nothing at all so I know to skip it. That doesn't matter for creation but will later for updates.

server/src/model/deliverable/parseCreateDeliverableInput.ts
TARehman: I use the spread for update because it handles undefined fields better. I might change this one when I do that.

server/src/model/deliverableAction/deliverableActionTypes.ts
TARehman: I think I'll handle it in the action queries when we get there.

server/src/model/deliverableDemonstrationType/setDeliverableDemonstrationTypes.ts
TARehman: Well, within reason. We'll need to think about it still.

server/src/model/deliverableDemonstrationType/queries/insertDeliverableDemonstrationTypes.ts
TARehman: Then you'd need to also accept the demonstrationId and the deliverableId because you can't insert without them. At that point, you're literally just taking the same arguments as input, but without them being in input.

server/tsconfig.json
TARehman: James and I discussed it, this resolves inconsistencies with the way the config was set up. Originally I was using AggregateError which necessitated going to a higher lib version, but then it turned out that throwing errors that way wasn't really supported by GQL anyway, so it became moot and I just left in the other stuff.

server/src/model/deliverable/createDeliverable.test.ts
TARehman: I don't think so. JS dates are not timezone aware anyway.

server/src/model/deliverable/createDeliverable.test.ts
TARehman: This becomes more relevant once we're dealing with updates.

server/src/model/deliverable/deliverableResolvers.test.ts
TARehman: It was unneeded before, the way I was doing it was redundant and CoPilot caught it while I was having it explain why `resolves.not.toThrow` was illogical.

PR #1075, opened 2026-04-13
Demos-1653  Adds BN folder to shared_lib and excel handling functions

james-becwar-cms:
Adds the utility functions for reading the BN workbook.  Validations will be in the next PR


shared_library/src/BN/index.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export async function parseBNFileFromPath(path: string): Promise<ExcelData> {
+  const data = await fs.readFile(path);
+  const blob = new Blob([data], {
+    type: "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
+  });
+  return await parseBNFile(blob);
+}
+
+export function excelColumnRow(
```
connor-parke-icf, line 31: might should specify the return type of this function as a CellValue, assuming thats what it should be.

shared_library/src/BN/index.test.ts
```diff
@@ -0,0 +1,97 @@
...
+  });
+
+  it("throws when file path does not exist", async () => {
+    const missingPath = path.join(currentDir, "..", "test", "fixtures", "missing.xlsx");
+    await expect(parseBNFileFromPath(missingPath)).rejects.toThrow();
+  });
+});
+
+describe("excelColumnRow", () => {
+  let workbookData: Awaited<ReturnType<typeof parseBNFileFromPath>>;
```
connor-parke-icf, line 58: i dont think ive seen "Awaited" before, but looks interesting. I suppose its most useful when you want to define a type as the return type of an async function and dont want it to be a Promise, as you have it here. Neat

connor-parke-icf (approved):
only thing i might add is an explicit return type on excelColumnRow(), im guessing it might can be inferenced by the IDE anyway. but other than that lgtm!

PR #1076, opened 2026-04-13
(1/3) DEMOS-1603: Adds `DeliverableDetailsManagementPage` Foundation

tresdonicf:
Per DEMOS-1603: This PR adds the foundation for the `DeliverableDetailsManagement` page. This can be accessed via `/deliverable/{deliverableId}`

Here's where this screen is after this PR:

image


client/src/components/header/Header.tsx
```diff
@@ -44,6 +45,14 @@ const HeaderLower: React.FC = () => {
     }
   }
 
+  // Match /deliverable/:deliverableId
+  if (path.match(/^\/deliverable\/[^/]+$/)) {
```
connor-parke-icf, line 49: aside: its out of scope for this ticket, but regex matching on path feels dangerous.  would be nice to revisit the routing component now that we have a small handful of pages and see if theres a different way we can render the various layouts. Working on the document preview page showed some awkwardness with auth and the layout that would be nice to clean up

client/src/mock-data/demonstrationMocks.ts
```diff
@@ -114,37 +113,11 @@ export const demonstrationMocks: MockedResponse[] = [
...
-    },
-    result: {
-      data: {
-        demonstration: MOCK_DEMONSTRATION,
-      },
-    },
+  ...MOCK_DEMONSTRATION_BY_ID_QUERIES.map((query) => ({
+    request: { query, variables: { id: MOCK_DEMONSTRATION_ID } },
+    result: { data: { demonstration: MOCK_DEMONSTRATION } },
     maxUsageCount: Number.POSITIVE_INFINITY,
```
connor-parke-icf, line 119: nice; i like that we can have a common set of fully saturated objects then lets the different queries passed into the mocks do the filtering of that data

client/src/pages/deliverables/DeliverableDetailsManagementPage.tsx
```diff
@@ -0,0 +1,84 @@
...
+      dueDate
+      status
+    }
+  }
+`;
+
+export type DeliverableDetailsManagementDeliverable = Pick<Deliverable, "id" | "deliverableType" | "dueDate" | "status"> & {
+  demonstration: Pick<Demonstration, "id" | "name"> & { state: { id: string } };
+  cmsOwner: { person: { fullName: string } };
+};
```
connor-parke-icf, line 35: NB: Might can instantiate a TypedDocumentNode to tie the query response to the GQL query so we dont need to pass it into the useQuery generic param.  Keeps every usage (not that theres more than one, but still) of the query expecting the correct params and response just by using the Query itself

client/src/pages/DemonstrationDetail/DemonstrationDetailHeader.tsx
```diff
@@ -10,8 +10,8 @@ import { useDialog } from "components/dialog/DialogContext";
 import { useNavigate } from "react-router-dom";
 
 export const DEMONSTRATION_HEADER_DETAILS_QUERY = gql`
-  query DemonstrationHeaderDetails($demonstrationId: ID!) {
-    demonstration(id: $demonstrationId) {
+  query DemonstrationHeaderDetails($id: ID!) {
```
connor-parke-icf, line 13: i like this change. we're not super consistent about how we define these queries in the schema one the server, but i like the convention of keeping these params drier.  would love to update the server schema to be consistently just using ID when its self-explanitory

connor-parke-icf (approved):
looks good to me!

client/src/pages/DemonstrationDetail/DemonstrationDetailHeader.tsx
tresdonicf: You know this one was only kinda needed because this route is supposed to show the `DemonstrationDetailHeader` but there was a conflict with the way the id for this and the `deliverableId` were being resolved.

Truth be told I kinda like the explicitness of the previous approach but it was just incompatible with the way we're handling the header in the layout.

Would be fixed by just explicitly using the header in the page component either way

client/src/pages/deliverables/DeliverableDetailsManagementPage.tsx
tresdonicf: That's a good idea!

client/src/components/header/Header.tsx
tresdonicf: I totally agree, I think we just want to lay these out on the "page" components itself honestly. Didn't wanna blow up the scope too much but I don't like how clever this component tries to be

PR #1078, opened 2026-04-13
DEMOS-1838: Create or Edit Demonstration - DEMO ID - Medicaid or CHIP Selection

nunesmlucas:
Adds a new required **DEMO ID** field to the Create/Edit Demonstration dialog, allowing users to select Medicaid Demonstration, CHIP, or both via two inline checkboxes. Displays a CHIP ID alongside the existing Demonstration ID in both the demonstration header breadcrumb and the Details tab summary.

This is **frontend-only** work. No backend/GraphQL schema changes were made. CHIP ID values and the Edit dialog's default `demoIds` are hardcoded placeholders until the backend lands.

Note: The "CHIP ID Alert" toast from the original story was removed per PM guidance — out of scope for this ticket.

image
image
image

### DIFF
--- a/client/src/components/dialog/demonstration/DemonstrationDialog.tsx
+++ b/client/src/components/dialog/demonstration/DemonstrationDialog.tsx
@@ -104,7 +104,8 @@ export const checkFormHasChanges = (
     updatedDemonstration.expirationDate !== initialDemonstration.expirationDate ||
     updatedDemonstration.sdgDivision !== initialDemonstration.sdgDivision ||
     updatedDemonstration.signatureLevel !== initialDemonstration.signatureLevel ||
-    updatedDemonstration.demoIds.join(",") !== initialDemonstration.demoIds.join(",")
+    [...updatedDemonstration.demoIds].sort((a, b) => a.localeCompare(b)).join(",") !==
+      [...initialDemonstration.demoIds].sort((a, b) => a.localeCompare(b)).join(",")
   );
 };
 
@@ -118,7 +119,7 @@ export const checkFormIsValid = (demonstration: DemonstrationDialogFields) => {
   if (!demonstration.projectOfficerId) {
     return false;
   }
-  if (!demonstration.demoIds || demonstration.demoIds.length === 0) {
+  if (demonstration.demoIds.length === 0) {
     return false;
   }
   if (
--- a/client/src/components/dialog/demonstration/DemonstrationDialog.tsx
+++ b/client/src/components/dialog/demonstration/DemonstrationDialog.tsx
@@ -2,6 +2,7 @@ import React, { useEffect, useState } from "react";
 
 import { BaseDialog } from "components/dialog/BaseDialog";
 import { Textarea } from "components/input";
+import { Checkbox } from "components/input/Checkbox";
 import { SelectSdgDivision } from "components/input/select/SelectSdgDivision";
 import { SelectSignatureLevel } from "components/input/select/SelectSignatureLevel";
 import { SelectUSAStates } from "components/input/select/SelectUSAStates";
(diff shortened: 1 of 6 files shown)


client/src/components/dialog/demonstration/DemonstrationDialog.tsx
```diff
@@ -89,7 +103,8 @@ export const checkFormHasChanges = (
     updatedDemonstration.effectiveDate !== initialDemonstration.effectiveDate ||
     updatedDemonstration.expirationDate !== initialDemonstration.expirationDate ||
     updatedDemonstration.sdgDivision !== initialDemonstration.sdgDivision ||
-    updatedDemonstration.signatureLevel !== initialDemonstration.signatureLevel
+    updatedDemonstration.signatureLevel !== initialDemonstration.signatureLevel ||
+    updatedDemonstration.demoIds.join(",") !== initialDemonstration.demoIds.join(",")
```
nunesmlucas: Comparing the array for changes

client/src/components/dialog/demonstration/EditDemonstrationDialog.tsx
```diff
@@ -119,6 +119,7 @@ const getDemonstrationDialogFields = (demonstration: Demonstration): Demonstrati
   expirationDate: demonstration.expirationDate
     ? formatDateForServer(demonstration.expirationDate)
     : "",
+  demoIds: ["medicaid"],
```
nunesmlucas, line 122: temporarily defaulting this for the frontend to show "one stored" choice.

client/src/components/table/tables/SummaryDetailsTable.tsx
```diff
@@ -89,6 +89,16 @@ export const SummaryDetailsTable: React.FC<{ demonstrationId: string }> = ({ dem
...
+      <div className={FIELD_CONTAINER_CLASSES}>
+        <div className={LABEL_CLASSES}>Demonstration ID</div>
+        <div className={VALUE_CLASSES}>{demonstration.id}</div>
+      </div>
+
+      <div className={FIELD_CONTAINER_CLASSES}>
+        <div className={LABEL_CLASSES}>CHIP ID</div>
+        <div className={VALUE_CLASSES}>21-W-00014/8</div>
+      </div>
+
```
nunesmlucas, line 101: Hardcoded values for QA. 
Integration story for FE and BE connection will resolve this.

client/src/pages/DemonstrationDetail/DemonstrationDetailHeader.tsx
```diff
@@ -110,7 +110,7 @@ export const DemonstrationDetailHeader: React.FC<DemonstrationDetailHeaderProps>
               Demonstration List
             </a>
             {/* \u00A0 is unicode for non-breaking space */}
-            {"\u00A0 > \u00A0"} {demonstration.id}
+            {"\u00A0 > \u00A0"} {demonstration.id} {"\u00A0|\u00A0"} 21-W-00014/8
```
nunesmlucas, line 113: > Hardcoded values for QA. Integration story for FE and BE connection will resolve this.

Same here

client/src/components/dialog/demonstration/DemonstrationDialog.tsx
connor-parke-icf: hmm, if the ids end up in a different order this will detect a change even if there isnt a meaningful one. In other places in the database we dont specify an exact ordering of the returned rows and i imagine this will follow suite, so it may be better to loop through the values themselves or something just in case.  Or maybe they can be sorted before joining for concat? that would probably work

client/src/components/dialog/demonstration/DemonstrationDialog.tsx
```diff
@@ -103,6 +118,9 @@ export const checkFormIsValid = (demonstration: DemonstrationDialogFields) => {
   if (!demonstration.projectOfficerId) {
     return false;
   }
+  if (!demonstration.demoIds || demonstration.demoIds.length === 0) {
```
connor-parke-icf: if demonstration is of type DemonstrationDialogFields, and that type requires a defined demoIds field, we shouldnt need to check falsey on demonstration.demoIds

client/src/components/dialog/demonstration/DemonstrationDialog.tsx
```diff
@@ -113,6 +131,45 @@ export const checkFormIsValid = (demonstration: DemonstrationDialogFields) => {
...
+        <span
+          className="cursor-help text-text-placeholder"
+          title="Select Medicaid Demonstration, CHIP, or both. Selecting CHIP will generate a unique CHIP ID."
+        >
+          <HintIcon />
+        </span>
+      </div>
+      <div className="flex gap-8 items-center flex-nowrap whitespace-nowrap">
+        {DEMO_ID_OPTIONS.map((option) => (
+          <Checkbox
```
connor-parke-icf, line 161: not sure if this is covered in this story or a followup one, but we should disable the checkboxes of theyre preloaded with a value, so that the user cannot unset a previously saved choice.

connor-parke-icf (approved):
approving, but i think we should make a couple of adjustments before merge:
* the order of the fields in an array is not guaranteed by the database (at least, following the pattern we have everywhere else), so the string concat to check for changes would fail if a different order is loaded (unless we're sorting somewhere that i missed). 
* it may be covered in a different story, but we should disable the appropriate checkbox if it is preloaded with a value, as the user should not be able to unset a previously saved demoId.

other than those two things, Looks good to me! placeholders pending integration story are fine and i like breaking the DemoId checkboxes into their own component

client/src/components/dialog/demonstration/DemonstrationDialog.tsx
tresdonicf: Could also turn them into a `Set()` and check if the `symetricDifference()` is non-empty. A bit much for two values but wanted to throw it out there!

client/src/components/dialog/demonstration/DemonstrationDialog.tsx
nunesmlucas: This is going to be part of the integration / behavior story. We made a note of it with Dipika

PR #1079, opened 2026-04-14
(3/3) DEMOS-1609: Create Deliverables using the API Endpoint

tresdonicf:
Per DEMOS-1609, uses the api to create deliverable slots!

https://github.com/user-attachments/assets/322c81ee-91ef-43d8-9431-abf784f5be2d

The table doesn't have a query quite yet so we can't update it but that will be an easy fix


dustbuster (approved):
LES DO ETT!!

dustbuster (approved):
do it!

dustbuster (approved):

connor-parke-icf (approved):
the previous changes fit together remarkably well with the new API, nice work! nothin really to say here, lgtm!

PR #1083, opened 2026-04-15
DEMOS-1800: Edit Deliverable (Layout)

nunesmlucas:
# DEMOS-1800 — CMS User: Edit Deliverable (Layout)

## Summary

Wires the previously-stubbed Edit button on the Deliverables Management table to a
new `EditDeliverableDialog` modal, plus a success toast, per DEMOS-1800.

## What changed

- **`components/dialog/deliverable/EditDeliverableDialog.tsx` (new)** — Edit modal
  built on `BaseDialog`, reusing the existing field components from the Add slot
  flow (`DeliverableTypeField`, `DeliverableNameField`, `CMSOwnerField`,
  `DemonstrationTypeField`, `SingleDeliverableScheduleType`). Also exports the
  `isDeliverableEditable` / `NON_EDITABLE_DELIVERABLE_STATUSES` helpers used by
  the table to gate the action button.
- **`components/table/tables/DeliverableTable.tsx`** — opens the new dialog from
  the existing Edit `CircleButton` and disables it when the selected row's status
  is `Accepted`, `Approved`, or `Received and Filed` (UI-only restriction for MVP).
- **`components/dialog/deliverable/index.ts`** — re-exports the new dialog and
  helpers.
- **`util/messages.ts`** — adds `DELIVERABLE_UPDATED_MESSAGE`
  ("Changes have been saved to the deliverable") used by the success toast.

## Implementation notes

- **Reused existing primitives only** — no new fields, dialogs, toast types, or
  icons. The dialog mirrors the structure of `AddDeliverableSlotDialog` so the
  layout is consistent with the rest of the deliverable flows.
- **No `useEffect`** — derived values (`dueDateWasChanged`, `isFormValid`,
  `hasChanges`, `reasonInvalid`) are computed inline from `useState`.
- **Backend wiring is intentionally out of scope** — DEMOS-1800 is the layout
  ticket, and there is no `UpdateDeliverable` mutation in the schema yet. The
  dialog accepts an optional `onSave` callback so the mutation can be wired in a
  follow-up without touching the layout.
- **Status gating is UI-only**, matching the MVP note in the ticket.

https://github.com/user-attachments/assets/ec296711-6267-421e-bf0e-d09454116ef5


client/src/components/table/tables/DeliverableTable.tsx
```diff
@@ -13,49 +13,53 @@ import { selectionTooltip } from "./actionTooltips";
...
   const renderActionButtons: DeliverableActionButtons = (table) => {
-    const selectedCount = table.getSelectedRowModel().rows.length;
+    const selectedRows = table.getSelectedRowModel().rows;
+    const selectedCount = selectedRows.length;
+    const singleSelectedDeliverable = selectedCount === 1 ? selectedRows[0].original : null;
 
-    const editEnabled = selectedCount === 1;
+    const selectedIsEditable =
+      singleSelectedDeliverable === null || isDeliverableEditable(singleSelectedDeliverable.status);
+    const editEnabled = selectedCount === 1 && selectedIsEditable;
```
nunesmlucas, line 97: Like beating a dead horse. We have this across multiple tables, should probably create a story to reduce this into a helper component.

client/src/components/table/tables/DeliverableTable.tsx
```diff
@@ -13,49 +13,53 @@ import { selectionTooltip } from "./actionTooltips";
 import { ImportIcon } from "components/icons/Action/ImportIcon";
 import { EditIcon } from "components/icons/Navigation/EditIcon";
 import { sortDeliverablesByDefault } from "util/sortDeliverables";
+import { EditDeliverableDialog, isDeliverableEditable } from "components/dialog/deliverable";
 
 export type DeliverableTableRow = DeliverablesPageTableRow;
 export type DeliverableTableViewMode = Exclude<PersonType, "non-user-contact">;
```
dustbuster: I need to swap this to UserType... nothing to do with you!

client/src/components/table/tables/DeliverableTable.tsx
dustbuster: meh i only see two named that. But not a bad idea.

client/src/components/table/tables/DeliverableTable.tsx
```diff
@@ -13,49 +13,53 @@ import { selectionTooltip } from "./actionTooltips";
...
 export const DeliverableTable: React.FC<{
   deliverables: DeliverableTableRow[];
   emptyRowsMessage?: string;
   viewMode: DeliverableTableViewMode;
-}> = ({
-  deliverables,
-  emptyRowsMessage = EMPTY_ROWS_MESSAGE,
-  viewMode,
-}) => {
+}> = ({ deliverables, emptyRowsMessage = EMPTY_ROWS_MESSAGE, viewMode }) => {
```
dustbuster, line 80: NIT: I am also working in this file. Can we keep the white space changes to a minimum? It just makes it harder to merge.

client/src/components/dialog/deliverable/EditDeliverableDialog.tsx
```diff
@@ -0,0 +1,256 @@
...
+      id
+      demonstrationTypes {
+        demonstrationTypeName
+        approvalStatus
+      }
+    }
+  }
+`;
+
+type DemonstrationTypesQueryResult = {
```
dustbuster: NIT: Any reason not to move this to types?

client/src/components/dialog/deliverable/EditDeliverableDialog.tsx
```diff
@@ -0,0 +1,256 @@
...
+  demonstration: {
+    id: string;
+    demonstrationTypes: {
+      demonstrationTypeName: string;
+      approvalStatus: Tag["approvalStatus"];
+    }[];
+  } | null;
```
dustbuster: Is this `| null` required? Could mask bug. Or i could be wrong. idk!

dustbuster (approved):
Couple comments. but nothing crazy enough to block it.

client/src/components/dialog/deliverable/EditDeliverableDialog.tsx
```diff
@@ -0,0 +1,256 @@
...
+import { DeliverableTypeField } from "./fields/DeliverableTypeField";
+import { DemonstrationTypeField } from "./fields/DemonstrationTypeField";
+import { SingleDeliverableScheduleType } from "./fields/schedule-type/SingleDeliverableScheduleType";
+
+export const EDIT_DELIVERABLE_DIALOG_TITLE = "Edit Deliverable";
+export const EDIT_DELIVERABLE_DIALOG_NAME = "edit-deliverable-dialog";
+export const EDIT_DELIVERABLE_SAVE_BUTTON_NAME = "button-edit-deliverable-confirm";
+export const EDIT_DELIVERABLE_REASON_FIELD_NAME = "edit-deliverable-reason";
+
+export const NON_EDITABLE_DELIVERABLE_STATUSES: ReadonlySet<DeliverableStatus> = new Set([
```
connor-parke-icf, line 22: interesting, ive never seen ReadonlySet before; good to know this exists.  Makes me wonder if this would be a more appropriate way to define our consts in constants.ts

client/src/components/dialog/deliverable/EditDeliverableDialog.tsx
connor-parke-icf: in general, the Types and Constants files are reserved for things that are used across both Client and Server. if this type is specific to a client-defined query, it should probably stay clientside

client/src/components/dialog/deliverable/EditDeliverableDialog.tsx
```diff
@@ -0,0 +1,256 @@
...
+  !NON_EDITABLE_DELIVERABLE_STATUSES.has(status);
+
+export interface EditDeliverableInput {
+  id: string;
+  name: string;
+  deliverableType: DeliverableType;
+  cmsOwnerUserId: string;
+  dueDate: string;
+  demonstrationTypes: string[];
+}
```
connor-parke-icf, line 38: aside: when we do the integration ticket, we should probably pull whatever type gets defined in the server's types.ts for this. I think the dueDate and dueDate change reason might get bundled together in an object on a single field; but thats not applicable right now

client/src/components/dialog/deliverable/EditDeliverableDialog.tsx
```diff
@@ -0,0 +1,256 @@
...
+  query EditDeliverableDemonstrationTypes($id: ID!) {
+    demonstration(id: $id) {
+      id
+      demonstrationTypes {
+        demonstrationTypeName
+        approvalStatus
+      }
+    }
+  }
+`;
```
connor-parke-icf: i think we may have a component for selecting demo types on the CreateDeliverableSlot component already.  /client/src/components/dialog/deliverable/fields/DemonstrationTypeField.tsx. it may or may not be applicable, i just remembered it exists

client/src/components/dialog/deliverable/EditDeliverableDialog.tsx
```diff
@@ -0,0 +1,256 @@
...
+}) => {
+  const { showSuccess } = useToast();
+
+  const { data: demoTypesData } = useQuery<DemonstrationTypesQueryResult>(
+    EDIT_DELIVERABLE_DEMO_TYPES_QUERY,
+    {
+      variables: { id: deliverable.demonstration.id },
+      skip: demonstrationTypeTags !== undefined,
+    }
+  );
```
connor-parke-icf: hmm, the skip here feels kinda odd to me; is this a fallback in case the demonstration's list of tags is undefined? if we're following the pattern of passing everything needed for the dialog into it as a prop, i feel like we shouldnt have an additional conditional query here. If the demonstration types on the demonstration is undefined, i feel like that should indicate that something higher up is broken.

client/src/components/table/tables/DeliverableTable.tsx
```diff
@@ -114,6 +122,12 @@ export const DeliverableTable: React.FC<{
           actionButtons={actionButtons}
         />
       )}
+      {deliverableBeingEdited && (
+        <EditDeliverableDialog
+          deliverable={deliverableBeingEdited}
+          onClose={() => setDeliverableBeingEdited(null)}
+        />
+      )}
```
tresdonicf: This isn't quite the pattern that we've been using much after the introduction of the `<DialogContext>`. Now the flow is generally to add a new entry to that which will render this page and a button or some other action can be hooked to `showEditDeliverableDialog(...)`

client/src/components/table/tables/DeliverableTable.tsx
```diff
@@ -13,49 +13,53 @@ import { selectionTooltip } from "./actionTooltips";
...
 
-  const showAddDeliverableDialog = () => { };
-  const showEditDeliverableDialog = () => { };
-  const showRemoveDeliverableDialog = () => { };
+  const [deliverableBeingEdited, setDeliverableBeingEdited] = useState<DeliverableTableRow | null>(
+    null
+  );
+
+  const showAddDeliverableDialog = () => {};
+  const showRemoveDeliverableDialog = () => {};
```
tresdonicf: Related to my other comment here, are these doing anything yet?

tresdonicf (commented):
Would it be possible to move the dialog in this into the `DialogContext`?

client/src/components/dialog/deliverable/EditDeliverableDialog.tsx
nunesmlucas: `ReadonlySet` is nice when you need `.has()` checks (fast + can't be mutated). Probably overkill for most of the as const lists in constants.ts since those just get looped over, but handy anywhere we're checking "is X in this group?"

client/src/components/dialog/DialogContext.tsx
```diff
@@ -235,6 +241,34 @@ export const useDialog = () => {
     );
   };
 
+  const showEditDeliverableDialog = (
+    deliverable: DeliverableTableRow,
```
tresdonicf, line 245: Thanks for making this change! Generally the pattern we use here is passing just the `id` of the thing to edit (so `deliverableId` here) and adding a query to the dialog itself that fetches the rows needed to populate the fields initially.

Would it be possible to change this to that pattern?

client/src/components/table/tables/DeliverableTable.tsx
```diff
@@ -21,6 +23,10 @@ export type DeliverableTableRow = Omit<
   name: string;
   demonstration: Pick<Deliverable["demonstration"], "id" | "name"> & {
     state: Pick<State, "id">;
+    demonstrationTypes: {
+      demonstrationTypeName: string;
+      approvalStatus: "Approved" | "Unapproved";
+    }[];
```
tresdonicf, line 29: This is a good example of the rhyme to the reason about passing in only the id to the dialog rather than the whole record - it avoids some of these type expansions further up the callstack.

If we imagine this being called in multiple places that all fetch disparate but related sets of `deliverable` records and we need to add a new field to support some new functionality we would have to add it to all of those places

tresdonicf (commented):
Thanks for changing to `useDialog`! I have one more request to isolate the data for filling in the fields to the `EditDeliverableDialog` itself and that dialog should be able to be called with a nice and easy `showEditDeliverableDialog(deliverableId)`

PR #1084, opened 2026-04-15
DEMOS-1691 suggestedApplicationTag Resolvers

JeffreyM-GA:
- Adds mutators for suggested application tags
- Mutators perform a simple update of tag status, and (if applicable), call setApplicationTags (which requires current applicationTags to be fetched)
image


dustbuster (commented):
Seems alright by me. WOULD FOR SURE run this by @TARehman Or @connor-parke-icf

dustbuster (approved):
okay. Not seeing much traction. So I'll just rubber stamp it

TARehman:
@connor-parke-icf is going to review this, I asked him to take a look and he said he would so I could focus on updateDeliverable. 😄

server/src/model/applicationTagSuggestion/applicationTagSuggestionSchema.ts
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  type Mutation {
+    replaceApplicationTagSuggestion(input: UpdateApplicationTagSuggestionInput!): Application
+  }
+
+  type Mutation {
+    removeApplicationTagSuggestion(input: UpdateApplicationTagSuggestionInput!): Application
+  }
+`;
```
connor-parke-icf, line 18: looking at the resolvers i think theres a misalignment here
```
export async function acceptApplicationTagSuggestion(
  _: unknown,
  { input }: { input: UpdateApplicationTagSuggestionInput },
): Promise<PrismaApplication>
```
`acceptApplicationTagSuggestion` only uses input.applicationId to return the application, but we can probably simplify this since we grab the applicationId from the database's existing suggestion `suggestion.applicationId `. Would be nice to make this just take in a suggestionId if thats all it really needs
```
export async function replaceApplicationTagSuggestion(
  _: unknown,
  { suggestionId, newValue }: { suggestionId: string; newValue: string },
): Promise<void>
```
here the resolver returns a Promise<void> but the schema is expecting an application to be returned.  I think we should update the resolver to return a PrismaApplication.  Also the resolver names this "newValue" but the UpdateApplicationTagSuggestionInput type doesnt have that field
```
export async function removeApplicationTagSuggestion(
  _: unknown,
  { suggestionId }: { suggestionId: string },
): Promise<void> 
```
Here the resolver only needs the suggestion id, but the schema is requiring an application Id. I think the schema shouldbe updated to only require the suggestion id. Also the schema expects an Application back, but the resolver is returning Promise<void>

every graphql operation returns _something_. We can circumvent this on the frontend by returning __typename, but thats a field on something (demo, application, document, etc)

server/src/model/applicationTagSuggestion/applicationTagSuggestionResolvers.ts
```diff
@@ -0,0 +1,121 @@
...
+): Promise<void> {
+  try {
+    return prisma().$transaction(async (tx) => {
+      await tx.applicationTagSuggestion.update({
+        where: { id: suggestionId },
+        data: {
+          status: {
+            connect: { id: "Removed" },
+          },
+        },
```
connor-parke-icf: So i think this is a valid way to set the status, but I think a more explicit way to do this is just set the statusId to "Removed" so that we dont need to involve the static constraint ApplicationTagSuggestionStatus table.
```
data: {
    statusId: "Removed"
}
```

server/src/model/applicationTagSuggestion/applicationTagSuggestionResolvers.ts
```diff
@@ -0,0 +1,121 @@
...
+        tagNames.push(newValue);
+      }
+
+      // Apply updated tags
+      await setApplicationTags(_, {
+        input: {
+          applicationId: suggestion.applicationId,
+          applicationTags: tagNames,
+        },
+      });
```
connor-parke-icf: NB: wonder if exporting this find and replace functionality to its own helper might be a little easier to test in isolation, since the existing method is a "drop all and recreate"-type operation

connor-parke-icf (commented):
I think we need a few adjustments to this, mainly the mismatch between the expected input and return types between the schema and resolvers. Also, unless this is defined elsewhere, i dont think this is currently getting loaded onto the server itself; we need a change in server/src/model/graphql.ts so its actually accessible in the API

server/src/model/applicationTagSuggestion/applicationTagSuggestionSchema.ts
connor-parke-icf: also we should be able to put these all under the same type Mutation {}

server/src/model/applicationTagSuggestion/applicationTagSuggestionResolvers.ts
```diff
@@ -1,91 +1,71 @@
...
-        where: { id: input.suggestionId },
-        data: { status: { connect: { id: "Accepted" } } },
-      });
-
-      // Fetch existing tags
-      const existingTags = await tx.applicationTagAssignment.findMany({
-        where: { applicationId: suggestion.applicationId },
-        select: { tagNameId: true },
+        where: { id: suggestionId },
+        data: { statusId: "Accepted" },
```
connor-parke-icf, line 43: nice simplification here! yeah just finding and updating the relevant tag should be enough!

server/src/model/applicationTagSuggestion/applicationTagSuggestionResolvers.ts
```diff
@@ -1,91 +1,71 @@
 import { handlePrismaError } from "../../errors/handlePrismaError";
-import { prisma } from "../../prismaClient";
+import { prisma, PrismaTransactionClient } from "../../prismaClient";
 import { setApplicationTags } from "../applicationTagAssignment";
 import { PrismaApplication } from "../application/applicationTypes";
-import { getApplication } from "../application";
-import { UpdateApplicationTagSuggestionInput } from "./applicationTagSuggestionSchema";
+import { ApplicationTagSuggestion } from "@prisma/client";
+
+export async function appendApplicationTags(
```
connor-parke-icf, line 7: nice helper

server/src/model/applicationTagSuggestion/applicationTagSuggestionSchema.ts
```diff
@@ -1,25 +1,18 @@
...
-
-  type Mutation {
-    replaceApplicationTagSuggestion(input: UpdateApplicationTagSuggestionInput!): Application
-  }
-
-  type Mutation {
-    removeApplicationTagSuggestion(input: UpdateApplicationTagSuggestionInput!): Application
+    acceptApplicationTagSuggestion(suggestionId: ID!): Application
+    replaceApplicationTagSuggestion(suggestionId: ID!, newValue: String!): Application
+    removeApplicationTagSuggestion(suggestionId: ID!): ApplicationTagSuggestion
```
connor-parke-icf, line 16: NB: typically we have had the pattern of returning the parent object that is being changed, rather than the specific object the operation is acting on.  That being said, i would prefer to do it the way you have here to keep the API more atomic, but since we are largely refetching stuff after every mutation anyway on frontend none of that should matter. The Schema is consistent with the resolver so this is totally fine

connor-parke-icf (approved):
fantastic work, thanks for addressing those concerns!

PR #1085, opened 2026-04-15
DEMOS-1889: adding GET funcion to demo detail

dustbuster:
Screenshot 2026-04-15 at 14 26 24

Screenshot 2026-04-15 at 14 26 53

### DIFF
--- a/client/src/components/dialog/deliverable/AddDeliverableSlotDialog.tsx
+++ b/client/src/components/dialog/deliverable/AddDeliverableSlotDialog.tsx
@@ -13,6 +13,7 @@ import { QuarterlyDeliverableSchedule } from "./fields/schedule-type/QuarterlyDe
 import { CreateDeliverableInput, DeliverableType, Demonstration, LocalDate, Tag } from "demos-server";
 import { useToast } from "components/toast";
 import { DELIVERABLE_SLOTS_CREATED_MESSAGE } from "util/messages";
+import { DELIVERABLES_PAGE_QUERY } from "components/table/tables/DeliverableTable";
 
 export const CREATE_DELIVERABLE_MUTATION = gql`
   mutation CreateDeliverable($input: CreateDeliverableInput!) {
@@ -23,9 +24,10 @@ export const CREATE_DELIVERABLE_MUTATION = gql`
 `;
 
 export const useCreateDeliverable = () => {
-  // TODO: last piece needed on here is that we need to refetch the deliverable slots after creation
-  // There's not currently a query for that yet
-  const [createDeliverable, { loading }] = useMutation(CREATE_DELIVERABLE_MUTATION);
+  const [createDeliverable, { loading }] = useMutation(CREATE_DELIVERABLE_MUTATION, {
+    refetchQueries: [{ query: DELIVERABLES_PAGE_QUERY }],
+    awaitRefetchQueries: true,
+  });
 
   const createDeliverables = async (inputs: CreateDeliverableInput[]) => {
     await Promise.all(
--- a/client/src/components/table/tables/DeliverableTable.tsx
+++ b/client/src/components/table/tables/DeliverableTable.tsx
@@ -1,6 +1,6 @@
 import React from "react";
 import { gql } from "@apollo/client";
-import type { Deliverable, Person, UserType, State } from "demos-server";
+import type { Deliverable, Person, State, UserType } from "demos-server";
 
 import { DeliverableColumns } from "../columns/DeliverableColumns";
 import { Table, type TableProps } from "../Table";
--- a/client/src/components/table/tables/DeliverableTable.tsx
+++ b/client/src/components/table/tables/DeliverableTable.tsx
@@ -1,6 +1,6 @@
 import React from "react";
 import { gql } from "@apollo/client";
-import type { Deliverable, Person, PersonType, State } from "demos-server";
+import type { Deliverable, Person, UserType, State } from "demos-server";
 
 import { DeliverableColumns } from "../columns/DeliverableColumns";
 import { Table, type TableProps } from "../Table";
@@ -29,7 +29,7 @@ export type DeliverableTableRow = Omit<
   cmsDocuments: Pick<Deliverable["cmsDocuments"][number], "id">[];
   stateDocuments: Pick<Deliverable["stateDocuments"][number], "id">[];
 };
-export type DeliverableTableViewMode = Exclude<PersonType, "non-user-contact">;
+
 export type DeliverablesQueryResult = {
   deliverables: DeliverableTableRow[];
 };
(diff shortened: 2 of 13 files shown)


client/src/components/dialog/deliverable/AddDeliverableSlotDialog.tsx
```diff
@@ -23,9 +24,10 @@ export const CREATE_DELIVERABLE_MUTATION = gql`
 `;
 
 export const useCreateDeliverable = () => {
-  // TODO: last piece needed on here is that we need to refetch the deliverable slots after creation
-  // There's not currently a query for that yet
-  const [createDeliverable, { loading }] = useMutation(CREATE_DELIVERABLE_MUTATION);
+  const [createDeliverable, { loading }] = useMutation(CREATE_DELIVERABLE_MUTATION, {
+    refetchQueries: [{ query: DELIVERABLES_PAGE_QUERY }],
+    awaitRefetchQueries: true,
+  });
```
dustbuster, line 30: 🙌

client/src/components/table/tables/DeliverableTable.tsx
```diff
@@ -14,8 +14,61 @@ import { ImportIcon } from "components/icons/Action/ImportIcon";
...
+  demonstration: Pick<Deliverable["demonstration"], "id" | "name"> & {
+    state: Pick<State, "id">;
+  };
+  cmsOwner: Pick<Deliverable["cmsOwner"], "id"> & {
+    person: Pick<Person, "fullName" | "id">;
+  };
+  submissionDate?: string;
+  cmsDocuments: Pick<Deliverable["cmsDocuments"][number], "id">[];
+  stateDocuments: Pick<Deliverable["stateDocuments"][number], "id">[];
+};
```
dustbuster, line 31: MIgrated some junk from DeliverablesPage, since we need to use this table in two different place. then we need to put it lower on the tree.

client/src/components/table/tables/DeliverableTable.tsx
```diff
@@ -14,8 +14,61 @@ import { ImportIcon } from "components/icons/Action/ImportIcon";
...
+      expectedToBeSubmitted
+      cmsDocuments {
+        id
+      }
+      stateDocuments {
+        id
+      }
+      createdAt
+      updatedAt
+    }
```
dustbuster, line 69: Not sure we need all this stuff. but i think having more than we need is better than having less?

client/src/components/table/tables/DemonstrationDeliverableTable.test.tsx
```diff
@@ -3,7 +3,7 @@ import { describe, expect, it } from "vitest";
 import { render, screen, within } from "@testing-library/react";
 
 import { DemonstrationDeliverableTable } from "./DemonstrationDeliverableTable";
-import type { DeliverableTableRow } from "pages/DeliverablesPage";
+import type { DeliverableTableRow } from "./DeliverableTable";
```
dustbuster, line 6: Then from there, we just update imports.

client/src/pages/DemonstrationDetail/deliverables/DeliverablesTab.test.tsx
```diff
@@ -6,7 +6,7 @@ import { DialogProvider } from "components/dialog/DialogContext";
 import { ADD_DELIVERABLE_SLOT_DIALOG_TITLE } from "components/dialog/deliverable";
 import { ADD_DELIVERABLE_SLOT_BUTTON_NAME, DeliverablesTab } from "./DeliverablesTab";
 import { TestProvider } from "test-utils/TestProvider";
-import { MOCK_DELIVERABLES } from "mock-data/deliverableMocks";
```
dustbuster, line 9: NO MOCK FOR YOU!

client/src/components/table/tables/DeliverableTable.tsx
```diff
@@ -14,8 +14,61 @@ import { ImportIcon } from "components/icons/Action/ImportIcon";
...
+  cmsDocuments: Pick<Deliverable["cmsDocuments"][number], "id">[];
+  stateDocuments: Pick<Deliverable["stateDocuments"][number], "id">[];
+};
 export type DeliverableTableViewMode = Exclude<PersonType, "non-user-contact">;
+export type DeliverablesQueryResult = {
+  deliverables: DeliverableTableRow[];
+};
+
+export const DELIVERABLES_PAGE_QUERY = gql`
+  query GetDeliverablesPage {
```
tresdonicf, line 38: A couple things here:
1. Looks like this name is a little bit out of date after the move, maybe should be something like `GetDeliverablesTable` instead
2. It's a bit odd that the query is declared here but only used elsewhere -  it would be nice if you could pass nothing to `<DeliverablesTable>` and get all deliverables or pass `<DeliverablesTable demonstrationId={...}>` to get only deliverables for that demonstration. This can be addressed later just thinking through what components should probably be responsible for what

client/src/pages/DemonstrationDetail/deliverables/DeliverablesTab.tsx
```diff
@@ -3,24 +3,30 @@ import { IconButton } from "components/button";
...
   parentDemonstration: AddDeliverableSlotDemonstration;
-  deliverables: DeliverableTableRow[];
 }) => {
   const { showAddDeliverableSlotDialog } = useDialog();
   const rawPersonType = getCurrentUser().currentUser?.person.personType;
   const viewMode = rawPersonType as DeliverableTableViewMode;
+  const { data, loading, error } = useQuery<DeliverablesQueryResult>(DELIVERABLES_PAGE_QUERY);
+  const deliverables = data?.deliverables.filter(
+    (deliverable) => deliverable.demonstration.id === parentDemonstration.id
+  ) ?? [];
```
tresdonicf, line 29: This does seem like it leaks data for all deliverables in the system (they'll be visible in the network tab). It's a straightforward approach and perhaps for something like CMS users it's alright

The `<DeliverablesTable />` with an optional `{demonstrationId}` would allow us to tighten this up a little bit but really this problem is solved in auth (as simply knowing an id of a demonstration you shouldn't have access to is the same issue)

tresdonicf (approved):
I added some comments about some things that could probably be inspected a bit closer but nothing to block on here!

client/src/pages/DemonstrationDetail/deliverables/DeliverablesTab.tsx
dustbuster: yeah i kinda had that in the back of my mind. Would be really nice to just pass in DemoId and then let it query.

dustbuster: TO DO! Not s super easy win!

PR #1087, opened 2026-04-15
Demos 1741 amendment extension resolver auth refactor 2

connor-parke-icf:
same thing but for Extension. nearly identical to the previous on since amendments are nearly identical to extensions.  One difference here, is that the getter methods for applications can be removed, since all instances have been replaced with arrow functions in the resolvers.


connor-parke-icf:
https://jiraent.cms.gov/browse/DEMOS-1981

james-becwar-cms (approved):
Server stuff looks great, and the client stuff looks like it would work to me.

server/src/model/extension/queries/selectManyExtensions.ts
```diff
@@ -0,0 +1,10 @@
+import { Extension as PrismaExtension, Prisma } from "@prisma/client";
+import { prisma, PrismaTransactionClient } from "../../../prismaClient";
+
+export async function selectManyExtensions(
+  where: Prisma.ExtensionWhereInput,
+  tx?: PrismaTransactionClient
+): Promise<PrismaExtension[]> {
+  const prismaClient = tx ?? prisma();
+  return await prismaClient.extension.findMany({ where });
```
TARehman, line 9: NB: Do we need to spread these? I did for mine, but I guess it doesn't really make sense?

server/src/model/extension/queries/selectManyExtensions.test.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  it("uses the default prisma client when no transaction client is provided", async () => {
+    const extensions = [{ id: "extension-1" }, { id: "extension-2" }] as PrismaExtension[];
+    extensionFindMany.mockResolvedValueOnce(extensions);
+
+    const result = await selectManyExtensions(where);
+
+    expect(prisma).toHaveBeenCalledExactlyOnceWith();
+    expect(extensionFindMany).toHaveBeenCalledExactlyOnceWith({ where });
+    expect(result).toBe(extensions);
```
TARehman, line 42: NB: Small code smell but I think this isn't a good thing to do because you're literally just checking if the mock returns the mock. It gives a false sense of security. If there was logic transforming the object in the query (a map to flatten things, etc) then it makes sense, but if you just directly return the queried object, I don't think this is giving us much value.

server/src/model/extension/queries/selectManyExtensions.test.ts
```diff
@@ -0,0 +1,72 @@
...
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { prisma, PrismaTransactionClient } from "../../../prismaClient";
+import { selectManyExtensions } from "./selectManyExtensions";
+
+vi.mock("../../../prismaClient", () => ({
+  prisma: vi.fn(),
+}));
+
+describe("selectManyExtensions", () => {
+  const extensionFindMany = vi.fn();
```
TARehman, line 11: NB: Why do you define this as a function here, and then later you use a different vi.fn() construction for the mockTransaction where you pass it in directly (lines 19-23)? Just curious, especially because on line 47, you _redefine_ it as a mock again before setting the return value.

server/src/model/extension/queries/selectManyExtensions.test.ts
```diff
@@ -0,0 +1,72 @@
...
+  });
+
+  it("uses the default prisma client when no transaction client is provided", async () => {
+    const extensions = [{ id: "extension-1" }, { id: "extension-2" }] as PrismaExtension[];
+    extensionFindMany.mockResolvedValueOnce(extensions);
+
+    const result = await selectManyExtensions(where);
+
+    expect(prisma).toHaveBeenCalledExactlyOnceWith();
+    expect(extensionFindMany).toHaveBeenCalledExactlyOnceWith({ where });
```
TARehman, line 41: You should validate that the transaction mock _isn't_ being called in this case, and vice versa. Check how I did it for some of the deliverable queries.

server/src/model/extension/queries/selectExtension.test.ts
TARehman, line 1: A lot of my comments on selectManyExtensions also apply here.

server/src/model/demonstration/demonstrationResolvers.test.ts
```diff
@@ -255,6 +259,18 @@ describe("demonstrationResolvers", () => {
     );
   });
 
+  it("delegates `Demonstration.extensions` to `extensionData.getManyExtensions`", async () => {
```
TARehman, line 262: NB: I think it makes sense for these to structure the test statements something like:

```typescript
describe("thingResolvers", () => {
  describe("Mutation.otherThing", () => {
    it("should do the things", async () => {
      ...
    });
  });

  describe("Query.thing", () => {
    ...
  });

  describe("Thing.someChild", () => {
    ...
  });
});
```

This is how I started to do it for deliverables. So, basically, the describe hierarchy matches the structure of the resolvers object being returned. Totally NB but I think it makes it really easy to structure the tests.

server/src/model/extension/extensionResolvers.test.ts
```diff
@@ -138,23 +133,59 @@ describe("extensionResolvers", () => {
...
+  it("resolves `Extension.signatureLevel`", () => {
+    const extension = {
+      signatureLevelId: "OA" satisfies SignatureLevel,
+    } as PrismaExtension;
+
+    const result = extensionResolvers.Extension.signatureLevel(extension);
+    expect(result).toBe(extension.signatureLevelId);
+  });
+
+  it("resolves `Extension.status`", () => {
```
TARehman, line 164: NB: The simplicity of doing it this way where we just have an arrow function in each of the resolver files is at war in my brain with the fact that we have to write and test the same function three times. I think its fine and that the juice of abstraction isn't worth the squeeze, but something about it makes me go "hmm". Maybe we should have a generic function that gets a field from something?

TARehman (approved):
Looks good to me. However, I would recommend addressing the comments regarding the tests for the queries. The broader pattern of how to structure resolver tests is really more of a discussion item / FYI kind of thing so no worries there.

connor-parke-icf:
these are good suggestions.  in the interest of getting what i have currently merged, i will address them in a follow up PR

server/src/model/extension/extensionResolvers.test.ts
connor-parke-icf: what pushed me toward inlining the logic in the resolver here was that a) the logic itself takes up as much space as the call to the helper method, b) this keeps the resolver structure consistent; it just so happens that amendments, extensions, and demonstrations share field., but most objects also have an ID or a createdAt for instance. should those also be abstracted to a helper? found it was just easier to parse when you dont need to consult a "thing dereferencer"

PR #1089, opened 2026-04-16
DEMOS-1787 add default for state user contact type, remove All Users …

JeffreyM-GA:
…from admin role options

Fixes 2 bugs with Manage Contacts Modal
1) Applies default contact type for state users
2) Removes a role that shouldn't have appeared for admin users

### DIFF
--- a/server/src/constants.ts
+++ b/server/src/constants.ts
@@ -85,7 +85,13 @@ export const ROLES = [
 
 export const CONTACT_TYPES = ["DDME Analyst", "Project Officer", "State Point of Contact"] as const;
 
-export const ADMIN_DEMONSTRATION_ROLES = ROLES;
+export const ADMIN_DEMONSTRATION_ROLES = [
+  "Project Officer",
+  "State Point of Contact",
+  "DDME Analyst",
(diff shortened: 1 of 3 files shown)


server/src/constants.ts
```diff
@@ -85,7 +85,13 @@ export const ROLES = [
...
 export const CONTACT_TYPES = ["DDME Analyst", "Project Officer", "State Point of Contact"] as const;
 
-export const ADMIN_DEMONSTRATION_ROLES = ROLES;
+export const ADMIN_DEMONSTRATION_ROLES = [
+  "Project Officer",
+  "State Point of Contact",
+  "DDME Analyst",
+  "Policy Technical Director",
+  "Monitoring & Evaluation Technical Director",
+] as const;
```
connor-parke-icf, line 94: discussed this, these constants will likely end up getting reworked/renamed some with the authorization stuff, so this seems fine to me.  "All Users" is a System Role, so doesnt have much to do with Demonstration assignment, and is likely to be removed altogether eventually

connor-parke-icf (approved):
LGTM!

PR #1090, opened 2026-04-16
Demos 1988 row based auth refactor documents

connor-parke-icf:
next up we have Documents! Same pattern as before, just routing the fetching through a common data interface for permission checking. exported the common "isStatePointOfContact" where clause to its own helper.

Going to make a smaller cleanup PR after this one to address some typing inconsistencies, consts, and misc items with the set of changes so far

https://github.com/user-attachments/assets/83b35e8a-fe3a-412c-bb0b-974a8ba063e7


james-becwar-cms (approved):
Approved, but wasn't sure on the deleted queries, msg'ed Connor to double check.

server/src/model/extension/extensionData.ts
```diff
@@ -17,14 +18,7 @@ const getPermissionFilters = (userId: string) =>
...
     "View Extensions on Assigned Demonstrations": {
-      demonstration: {
-        demonstrationRoleAssignments: {
-          some: {
-            personId: userId,
-            roleId: "State Point of Contact",
-          },
-        },
-      },
+      demonstration: isStatePointOfContactOnDemonstration(userId),
```
TARehman, line 21: This is a good refactor, definitely makes sense.

server/src/model/deliverable/deliverableResolvers.test.ts
```diff
@@ -99,6 +97,12 @@ describe("deliverableResolvers", () => {
     },
   };
 
+  const testContext: GraphQLContext = {
```
TARehman, line 100: NB: Not sure if we should do this as a Partial and then do the `as GraphQLContext` later, but its fine either way. I like the `Partial` approach because it helps me fill in fields I know are valid even when I want to leave some out; the `as GraphQLContext` at the end just makes that go away if I understand correctly. That being said, that's entirely a developer ergonomics thing for the most part.

Side note: we should probably add some way to lint / build the test files so that when types change we can find all the type errors that pop up in there, that's not included in the regular build process.

server/src/model/document/queries/selectDocument.test.ts
```diff
@@ -0,0 +1,63 @@
...
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { prisma, PrismaTransactionClient } from "../../../prismaClient";
+import { selectDocument } from "./selectDocument";
+
+vi.mock("../../../prismaClient", () => ({
+  prisma: vi.fn(),
+}));
+
+describe("selectDocument", () => {
+  const documentFindAtMostOne = vi.fn();
```
TARehman, line 11: NB: Some comment as earlier about the mocking approach but I know you plan to make a second pass on that stuff.

TARehman (approved):
It's a lot because of how much we use documents, but it all seemed okay to me on a review as best I could see. 🚀 ⛵ ship it 😄

PR #1092, opened 2026-04-16
DEMOS-1957-update-deliverable-mutator

TARehman:
This ended up being larger than I anticipated, mostly because there's more logic than it initially seems inside of the update deliverable mutator. It's really a wrapper for a few different kinds of operations, which adds to the complexity. This PR does the following regarding deliverables.

- Added `checkDeliverableInputFunctions`, a set of functions for checking the inputs to various deliverable mutators, to enable validation.
- Switched the incoming demonstration types to use the Set() system in JS, which made some of the validation easier - this necessitated some changes to `createDeliverable`.
- Moved checking of some things to the `parseX` functions; `parseX` handles checking the due date input for validity, and checking for duplicated inputs on the demonstration types. This makes sense because the parser ensures that what comes out is a properly formatted date and a Set instead of a list, so it makes sense to check those things during parsing.
  - Note that here, we still check if the due date is in the future either in the `createDeliverable` mutator (since it is required to be present there) or in `manuallyUpdateDeliverableDueDate` (which is invoked by `updateDeliverable`). We could, in theory, move these into the `parseX` as well; I just felt like it was a little safer to do them in the function itself that changes the data, rather than assuming that the function is getting valid data.
  - Also note, I renamed the parse file to put all the parsing stuff in one file, which is why two files show up as deleted.
- Updated `deliverableResolvers` to add the `demonstrationTypes` as well as to expose the new API endpoints.
- Got rid of `deliverableTypes` because the only types that really mattered were the ones relating to parsing the inputs, so I just put them in the parsing file.
- Renamed update queries to `editX` per a discussion with @connor-parke-icf.
- Added `manuallyUpdateDeliverableDueDate` to handle cases where the `updateDeliverable` payload includes changes to the due date. This handles a number of things, including making sure that the change is real, properly logging it into `deliverable_action` if it is, and handling the case of when something that is `Past Due` is moved to `Upcoming`.
- Added a query for editing deliverables, which can be used for editing things not necessarily available via the API (for instance, the status).
- Added 'updateDeliverableDemonstrationTypes` to handle the cases where new demonstration types are part of an update payload. Much like `manuallyUpdateDeliverableDueDate`, this also bundles in logic necessary to deal with this.
- Rewrote the `validateDeliverableInput` to be more modular and use the functions created to make validation more clear (and make testing more streamlined).
- Dropped `deliverableActionTypes` for similar reasons as to the other type file; moved the relevant type to a single place and exported it from there.
- Added queries and updates for the `deliverableDemonstrationType` records to facilitate the `updateDeliverableDemonstrationTypes` functions.

Other small changes included in this are below.

- Small tweaks to the docker-compose.yml file for the devcontainer. We weren't using the data volume we make for the database. Also, localstack moved to a model where you need an auth token to use it, so we need to pin on an earlier version or it'll stop working if you rebuild.
- We currently have TagName as a scalar mostly to provide semantic meaning (telling you what kind of string needs to be there, but not necessarily what is in it). However, we should really avoid the possibility of an empty string as a tag. Simply adding to customScalarResolvers to use the NonEmptyStringResolver to resolve the TagName fields solves this issue.
- Small changes were made to the dateUtilities module. The confusingly named `getEasternNow()` wasn't actually getting you the current time, just the start of day and end of day. I added the actual current time of invocation, and also renamed some of the functions to get rid of the old `__foo` naming pattern that we're not using anymore. Tests were updated as well.
- I eliminated redundant .js file extensions throughout a bunch of the files related to deliverables.
- Added a new validation utility to check if two Sets are the same or not.

Tests are passing. SonarQube is passing at present, 100% coverage. Exercised the API locally to the best of my ability but entirely possible that I missed something.


client/src/components/table/tables/DeliverableTable.tsx
```diff
@@ -16,7 +16,7 @@ import { sortDeliverablesByDefault } from "util/sortDeliverables";
 
 export type DeliverableTableRow = Omit<
   Deliverable,
-  "demonstration" | "cmsOwner" | "cmsDocuments" | "stateDocuments" | "name"
+  "demonstration" | "cmsOwner" | "demonstrationTypes" | "cmsDocuments" | "stateDocuments" | "name"
 > & {
   name: string;
   demonstration: Pick<Deliverable["demonstration"], "id" | "name"> & {
```
connor-parke-icf, line 22: typically i tend to prefer Pick<> over Omit<> for this reason; if things are added to the main type we need to update all the omits to exclude it. With Pick, a field getting removed presumably already would require an update at its usage site

server/src/model/deliverable/queries/editDeliverable.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export type EditDeliverableInput = {
+  name?: NonEmptyString;
+  statusId?: DeliverableStatus;
+  deliverableTypeId?: DeliverableType;
+  cmsOwnerUserId?: string;
+  dueDate?: Date;
+};
+
+export async function editDeliverable(
```
connor-parke-icf, line 13: NB: hmm, i thought for the query functions we were going to go with the involved SQL operation as the name, so this would be "updateDeliverable" and the function that calls it to do the logical mutation would be "edit".

server/src/model/deliverable/queries/getDeliverable.test.ts
```diff
@@ -2,9 +2,9 @@ import { describe, it, expect, vi, beforeEach } from "vitest";
 import { getDeliverable } from "./getDeliverable";
 
 // Mock imports
-import { prisma } from "../../../prismaClient.js";
+import { prisma } from "../../../prismaClient";
```
connor-parke-icf, line 5: ill be making a followup PR for my permissions changes to make this same adjustment there (and utilize an index.ts)

server/src/model/deliverable/queries/insertDeliverable.test.ts
```diff
@@ -61,7 +61,7 @@ describe("insertDeliverable", () => {
       isEasternTZDate: true,
       easternTZDate: new TZDate(2025, 2, 14, 3, 17, 22, 931, "America/New_York"),
     },
-    demonstrationTypes: [],
+    demonstrationTypes: new Set<TagName>(),
```
connor-parke-icf, line 64: i like this change; wonder how far we can stretch this pattern since basically everywhere on the server we dont care about the ordering of things and control for duplicates.

server/src/model/deliverable/checkDeliverableInputFunctions.test.ts
```diff
@@ -0,0 +1,165 @@
...
+
+describe("checkDeliverableInputFunctions", () => {
+  describe("checkDemonstrationStatus", () => {
+    it("should return undefined if the demonstration is Approved", () => {
+      const testInput: Partial<PrismaDemonstration> = {
+        id: "abc123",
+        statusId: "Approved" satisfies ApplicationStatus,
+      };
+      const result = checkDemonstrationStatus(testInput as PrismaDemonstration);
+      expect(result).toBeUndefined();
```
connor-parke-icf, line 32: Nit: since we return errors if invalid and nothing if valid, maybe this could be renamed to be like getDemonstrationStatusError or something; definitely not needed though this is totally fine

server/src/model/deliverable/checkDeliverableInputFunctions.test.ts
```diff
@@ -0,0 +1,165 @@
...
+      const result = checkForDuplicateDemonstrationTypes(testInput);
+      expect(result).toBe(
+        "Duplicate demonstration types were included on the input: Free Insulin, Nutrition Counseling."
+      );
+    });
+  });
+
+  describe("checkDueDateInFuture", () => {
+    beforeEach(() => {
+      vi.useFakeTimers();
```
connor-parke-icf, line 137: interesting, never seen useFakeTimers either, but probably another example where we could utilize this testing on the frontend for timezone inconsistencies

server/src/model/deliverable/createDeliverable.ts
```diff
@@ -18,12 +17,13 @@ export async function createDeliverable(
...
 
     const newDeliverable = await insertDeliverable(parsedInput, tx);
+    const newDemonstrationTypes = parsedInput.demonstrationTypes ?? new Set();
 
     await setDeliverableDemonstrationTypes(
       {
         deliverableId: newDeliverable.id,
         demonstrationId: parsedInput.demonstrationId,
-        demonstrationTypes: parsedInput.demonstrationTypes ?? [],
+        demonstrationTypes: Array.from(newDemonstrationTypes),
```
connor-parke-icf, line 26: i take it, Prisma doesnt play well with converting Sets to arrays automatically?

server/src/model/deliverable/createDeliverable.test.ts
```diff
@@ -51,8 +47,9 @@ describe("createDeliverable", () => {
   const testContext: GraphQLContext = {
     user: {
       id: "57f92f14-7c5e-4c78-a774-5a54d7e9c2e7",
-      sub: "82d0e8e4-82d0-447c-b1bb-52227e49cf51",
-      role: "demos-cms-user",
+      cognitoSubject: "82d0e8e4-82d0-447c-b1bb-52227e49cf51",
+      personTypeId: "demos-cms-user",
+      permissions: ["View All Demonstrations"],
```
connor-parke-icf, line 52: i saw this when doing document permissions as well, im curious how this didnt get caught by the linter when GraphQLContext changed

server/src/model/deliverable/deliverableResolvers.test.ts
```diff
@@ -213,6 +221,29 @@ describe("deliverableResolvers", () => {
     });
   });
 
+  describe("resolveDeliverableDemonstrationTypes", () => {
+    it("should query the demonstraiton types of the parent deliverable", async () => {
```
connor-parke-icf: NB: typo

server/src/model/deliverable/deliverableResolvers.test.ts
```diff
@@ -213,6 +221,29 @@ describe("deliverableResolvers", () => {
...
+        {
+          demonstrationId: testDemonstrationId,
+          deliverableId: testDeliverableId,
+          demonstrationTypeTagNameId: "Vitamin A Supplementation for Newborns",
+        },
+      ]);
+
+      const result = await resolveDeliverableDemonstrationTypes(
+        testDeliverable as PrismaDeliverable
+      );
```
connor-parke-icf, line 241: ive started leaning into this pattern more and more, where in the test we just cast an object that has only the fields we care about as the saturated object

server/src/model/deliverable/deliverableResolvers.ts
```diff
@@ -84,6 +87,14 @@ export function resolveDeliverableDueDateType(parent: PrismaDeliverable): Delive
   return parent.dueDateTypeId as DeliverableDueDateType;
 }
 
+export async function resolveDeliverableDemonstrationTypes(
+  parent: PrismaDeliverable
+): Promise<TagName[]> {
+  return (await getDeliverableDemonstrationTypes(parent.id)).map(
+    (demonstrationType) => demonstrationType.demonstrationTypeTagNameId
+  );
+}
```
connor-parke-icf, line 96: NB: for resolvers that are just one simple call, i feel like we can inline this in the resolver array itself.  that way testing that the deliverableResolvers.Deliverable.demonstrationTypes correctly points to this function is the same test as the function itself

server/src/model/deliverable/deliverableSchema.ts
```diff
@@ -21,6 +21,7 @@ export const deliverableSchema = gql`
     cmsOwner: User!
     dueDate: DateTime!
     dueDateType: DeliverableDueDateType!
+    demonstrationTypes: [TagName!]!
```
connor-parke-icf, line 24: might want this to be Tag instead of TagName in case we want to also show the approvalStatus

server/src/model/deliverable/index.ts
```diff
@@ -1,12 +1,32 @@
...
+export { updateDeliverableDemonstrationTypes } from "./updateDeliverableDemonstrationTypes";
 
 // Queries
 export { getDeliverable } from "./queries/getDeliverable";
 export { getManyDeliverables } from "./queries/getManyDeliverables";
+export { insertDeliverable } from "./queries/insertDeliverable";
+export { editDeliverable } from "./queries/editDeliverable";
 
 // Types & Constants
-export type { ParsedCreateDeliverableInput } from "./deliverableTypes.js";
```
connor-parke-icf, line 12: as an aside, i feel like we have some linting issue going on in server where some things arent getting prettified/adjusted. I feel like we dont have this ".js" issue clientside

server/src/model/deliverable/manuallyUpdateDeliverableDueDate.ts
```diff
@@ -0,0 +1,67 @@
...
+  // If the dates don't match, do an update
+  if (!datesMatch) {
+    await editDeliverable(
+      deliverableId,
+      {
+        dueDate: input.dueDate.newDueDate.easternTZDate,
+        statusId: newStatus,
+      },
+      tx
+    );
```
connor-parke-icf, line 50: i agree that the datesMatch check belongs here instead of in the more general parser/validator. It requires context outside of the input itself

server/src/model/deliverable/manuallyUpdateDeliverableDueDate.test.ts
```diff
@@ -0,0 +1,242 @@
...
+  const mockDemonstrationId = "6ba5407b-3706-4795-b8a8-3e32e8fa77ac";
+  const mockDeliverable: Partial<PrismaDeliverable> = {
+    id: testDeliverableId,
+    demonstrationId: mockDemonstrationId,
+    dueDate: mockGeneralDate,
+    statusId: "Under CMS Review" satisfies DeliverableStatus,
+  };
+
+  // Mock transaction
+  const mockTransaction: any = "Test!";
```
connor-parke-icf, line 52: yeah, in the linter "any"s should be getting flagged, but here its not. not sure if intentional, but there have been some times where i catch no linting issues, then things pop up later

server/src/model/deliverable/parseDeliverableInputs.ts
```diff
@@ -0,0 +1,88 @@
...
+export type ParsedUpdateDeliverableInput = {
+  name?: NonEmptyString;
+  deliverableType?: DeliverableType;
+  cmsOwnerUserId?: string;
+  dueDate?: ParsedUpdateDueDate;
+  demonstrationTypes?: Set<TagName>;
+  status?: DeliverableStatus;
+};
+
+export function parseCreateDeliverableInput(
```
connor-parke-icf, line 36: iirc, we decided to defer the checking for "deliverable of type x requires at least one demonstrationType" for post-mvp/later yeah?

server/src/model/deliverable/updateDeliverable.ts
```diff
@@ -0,0 +1,46 @@
...
+  checkOptionalNotNullFields(
+    ["name", "deliverableType", "cmsOwnerUserId", "dueDate", "demonstrationTypes"],
+    input
+  );
+  const parsedInput = parseUpdateDeliverableInput(input);
+
+  const updatedDeliverable = await prisma().$transaction(async (tx) => {
+    await validateUpdateDeliverableInput(deliverableId, parsedInput, tx);
+
+    // Directly edit name, deliverable type, and CMS owner
```
connor-parke-icf, line 30: NB: i feel like there miiiight be some redundancy here. Does passing in a field as undefined and not including a field have different results to prisma?

server/src/model/deliverable/updateDeliverableDemonstrationTypes.ts
```diff
@@ -0,0 +1,42 @@
...
+  // Can turn these into a set because DB guarantees uniqueness
+  const oldDemonstrationTypes: Set<TagName> = new Set(
+    (await getDeliverableDemonstrationTypes(deliverableId, tx)).map(
+      (demonstrationType) => demonstrationType.demonstrationTypeTagNameId
+    )
+  );
+
+  // If there's a difference, do an update on the database
+  const diff = findSetDifferences(oldDemonstrationTypes, updateInput.demonstrationTypes);
+  if (!diff.setsMatch) {
```
connor-parke-icf, line 32: might not need a key of setsMatch if we return something falsey from findSetDifferences if there are none, but havnt gotten to the utils get to know exactly what happening there

server/src/model/deliverableDemonstrationType/deliverableDemonstrationTypeTypes.ts
```diff
@@ -0,0 +1,7 @@
+import { TagName } from "../../types";
+
+export type SetDeliverableDemonstrationTypesInput = {
+  deliverableId: string;
+  demonstrationId: string;
+  demonstrationTypes: TagName[];
+};
```
connor-parke-icf, line 7: if this is the only type def in here, i feel like we can move this to the Schema or somewhere else.  Generally, im not big into the standalone types file

server/src/model/deliverableDemonstrationType/setDeliverableDemonstrationTypes.ts
```diff
@@ -1,7 +1,10 @@
 import { prisma, PrismaTransactionClient } from "../../prismaClient";
-import { SetDeliverableDemonstrationTypesInput } from "../../types";
+import {
+  deleteAllDeliverableDemonstrationTypes,
+  insertDeliverableDemonstrationTypes,
+  SetDeliverableDemonstrationTypesInput,
```
connor-parke-icf, line 5: yeah, we could probably define the Type here instead of its own file

server/src/validationUtilities.ts
```diff
@@ -15,3 +15,38 @@ export function findDuplicates<T>(items: T[]): T[] {
 
   return duplicates;
 }
+
+export function findSetDifferences<T>(
```
connor-parke-icf, line 19: could "setsMatch" and "common" be derived from the exclusive lists of inS1Only and inS2Only? this util might be a little bigger in scope than needed.  Im honestly surprised there isnt a built-in method on Set that does this already

connor-parke-icf (approved):
very nice work, i dont _think_ there are any blockers here, just suggestions and nitpicks and questions.  Most of those i think i would rather be resolved in a followup PR than this one since they might be name changes, which git would probably get confused by and be hard to validate. 

Happy to sync directly on this, but i think largely everything here checks out. I like the separation of concerns here, especially for validation.

client/src/components/table/tables/DeliverableTable.tsx
TARehman: TBCH I just added this so the front would build.

server/src/model/deliverable/queries/editDeliverable.ts
TARehman: Oh, I guess I had the opposite understanding because we already have several logical operations called update. But we could change that later if we want.

server/src/model/deliverable/createDeliverable.ts
TARehman: Prisma might, but TS doesn't like it for sure.

server/src/model/deliverable/createDeliverable.test.ts
TARehman: It does, it's just in a test so it doesn't get caught by build because we don't build the test files.

server/src/model/deliverable/deliverableResolvers.test.ts
TARehman:
```suggestion
    it("should query the demonstration types of the parent deliverable", async () => {
```

server/src/model/deliverable/deliverableResolvers.ts
TARehman: Sure, it's just not the pattern the file follows currently, so I kept it the same.

server/src/model/deliverable/deliverableSchema.ts
TARehman: I don't think that's going to be relevant here, to be honest. For deliverables, they're more like application tags than they are full-on demonstration types.

server/src/model/deliverable/manuallyUpdateDeliverableDueDate.test.ts
TARehman: The tests aren't being linted as aggressively. And the `any` here is actually useful to make the tests work correctly without extensive extra mocking. I would never want to use `any` in the actual code but I'm more comfortable with type hacks to make the tests work fine.

server/src/model/deliverable/parseDeliverableInputs.ts
TARehman: Correct, its not fully defined yet, but hopefully this framework would allow us to add it without much pain. Though I'd put that in the validation functions, not the parser.

server/src/model/deliverable/updateDeliverable.ts
TARehman: I don't think there's any practical difference, and I don't even think it's a Prisma thing - it's a JS thing. If a field is undefined, it's the same as if you ask for a field that doesn't exist. That being said, it's slightly easier to reason about if the keys don't change over time. If you have something like `{ foo: undefined }` then `foo` does exist as a key, even if the value is undefined.

server/src/model/deliverableDemonstrationType/deliverableDemonstrationTypeTypes.ts
TARehman: I don't think there's going to be a schema for this eventually, which is the issue. The only path to resolve these is going to go through `deliverableResolvers` most likely.

server/src/validationUtilities.ts
TARehman: Shockingly, there's very little in the way of built-ins for this stuff. Also, the utility used to be something looking at list differences until I just decided to swap in and use sets directly, which is the source of some of the structure residually.

That being said, I also don't entirely know what you mean. Of course common could be derived from the two sets, but like, that's what this function does? :grin: Similarly, the `setsMatch` is done by checking if the sets match or not, which is what this function...does?

PR #1094, opened 2026-04-17
(3/3) DEMOS-1603: Mock data cleanup and polish

tresdonicf:
Last PR for DEMOS-1603, this one cleans up some of the mock-data for deliverables and does a couple thing for the deliverable details page:

- fix comment box color
- hide side nav for this page
- Use `deliverables/{id}` instead of `deliverable/{id}` for consistency with `demonstrations/`

image


client/src/components/table/tables/DeliverableTable.test.tsx
tresdonicf, line 1: Moved the mock data relevant to the test to this test

client/src/components/table/tables/DeliverableTable.tsx
```diff
@@ -56,16 +54,6 @@ export const DELIVERABLES_PAGE_QUERY = gql`
...
-      dueDateType
-      expectedToBeSubmitted
-      cmsDocuments {
-        id
-      }
-      stateDocuments {
-        id
-      }
-      createdAt
-      updatedAt
```
tresdonicf, line 68: Unused

client/src/layout/PrimaryLayout.tsx
```diff
@@ -4,11 +4,18 @@ import { SideNav } from "./SideNav";
...
 import { DialogProvider } from "components/dialog/DialogContext";
 
+const ROUTES_WITH_HIDDEN_SIDENAV = ["/demonstrations/", "/deliverables/"];
+
+export const shouldHideSideNav = (pathname: string) =>
+  ROUTES_WITH_HIDDEN_SIDENAV.some((route) => {
+    const pathStartsWithRoute = pathname.startsWith(route);
+    const pathIsLongerThanRoute = pathname.length > route.length;
+    return pathStartsWithRoute && pathIsLongerThanRoute;
+  });
```
tresdonicf, line 14: Moved this into a lil function

client/src/mock-data/deliverableMocks.ts
tresdonicf, line 1: Cleaned up this file a bit, used the spread operator instead of the function creation / separate type mechanism.

client/src/pages/DeliverablesPage.test.tsx
tresdonicf, line 1: Same here moved relevant tests to this file

client/src/components/table/tables/DeliverableTable.tsx
```diff
@@ -16,7 +16,7 @@ import { sortDeliverablesByDefault } from "util/sortDeliverables";
 
 export type DeliverableTableRow = Omit<
   Deliverable,
-  "demonstration" | "cmsOwner" | "demonstrationTypes" | "cmsDocuments" | "stateDocuments" | "name"
+  "demonstration" | "cmsOwner" | "demonstrationTypes" | "cmsDocuments" | "stateDocuments" | "name" | "dueDateType" | "expectedToBeSubmitted" | "createdAt" | "updatedAt"
```
connor-parke-icf, line 19: i feel like it may be better to specify the fields we need rather than specify the fields we dont; though i dont suspect too many more changes coming to the deliverable schema at this point

client/src/layout/PrimaryLayout.test.tsx
```diff
@@ -0,0 +1,50 @@
...
+    it("returns true for a string id", () => {
+      expect(shouldHideSideNav("/deliverables/abc-def")).toBe(true);
+    });
+
+    it("returns false for a bare trailing slash", () => {
+      expect(shouldHideSideNav("/deliverables/")).toBe(false);
+    });
+  });
+
+  describe("list / other routes", () => {
```
connor-parke-icf, line 33: good that we have a dedicated test for this now; didnt realize that was missing before

client/src/pages/DeliverablesPage.test.tsx
```diff
@@ -17,23 +18,40 @@ vi.mock("components/user/UserContext", async (importOriginal) => {
   };
 });
 
+const MOCK_DELIVERABLE_TABLE_ROWS = [
```
connor-parke-icf, line 21: im not super huge into importing data from the mocks file into tests; i think i would rather explicitly spell out the base object in the test file to keep it decoupled from the mock view.  that being said, i do really like the spread pattern here.  More and more ive been leaning into just making a very sparse object and casting it as the full object for tests where we only need one or two fields. For instance in the server GraphQLContext includes a user with id, cognitoSubject, personTypeId, and permissions, but for 90% of those tests we only care about the user id. so i can make a 
```
const mockContest: GraphQLContext = { 
  user: {
    id: "abc123"
  }
} as GraphQLContext;
```
and use that in place of a whole context object

client/AGENTS.md
```diff
@@ -19,6 +19,7 @@ This file provides instructions for AI agents to use when generating or editing
 
 - Reuse shared types from `demos-server` when available; do not duplicate local copies.
 - Use `Pick<>` or similar utility types when a component needs only part of a larger type.
+- Prefer `Pick` over `Omit` for type derivations.
```
connor-parke-icf, line 22: HA beat me to it!

connor-parke-icf (approved):
looks good to me! most of my comments looks like you already considered looking at the changes to agents.md, so nice!

total aside, id rather have the path be singular <base_url>/deliverable/:id instead of plural; same with demonstrations.  "Demonstrations" feels like a listing, and "Demonstration" feels like a detail view. but like, this is completely inconsequential.

client/src/pages/DeliverablesPage.test.tsx
tresdonicf: Not my favorite either! If we only have one "default" type export to override without needing to redeclare every field I'm kind of into that but it should probably be sentinel values like empty strings and `0`.

For now just didn't want to change this too much since it was wrapped up in the test assertions! Should be easier to take that additional hop when needed now

tresdonicf:
> looks good to me! most of my comments looks like you already considered looking at the changes to agents.md, so nice!
> 
> total aside, id rather have the path be singular <base_url>/deliverable/:id instead of plural; same with demonstrations. "Demonstrations" feels like a listing, and "Demonstration" feels like a detail view. but like, this is completely inconsequential.

Totally, haha - trying to codify rules into that as I come across them, it's getting better and might be about time for a real `.instructions.md` file!

Valid feeling! The canonical wisdom here is to use the plural version always - though our app uses client-side routing and GQL (for real "resources") and this guidance generally comes from server-side routing / REST APIs. I do think there's still a valid wisdom that holds true which is that the paths being additive bodes well for fallbacks.

So you go to `/resources` and get a list and `/resources/` and get a list and `/resources/123` and get a single resource. This means you've avoided the need to ever maintain a separate route `/resource/...`. which can be helpful for fallback matching if a particular resource doesn't exist compared to the need to check if the id exists at `resource/` and send the user back to `resources/` if not which is mostly just unneeded complexity

PR #1095, opened 2026-04-17
task/clean-up-sq-smells

TARehman:
Cleans up or suppresses a few SQ code smells identified by the scanner.

### DIFF
--- a/server/src/model/application/applicationResolvers.ts
+++ b/server/src/model/application/applicationResolvers.ts
@@ -30,7 +30,7 @@ export async function resolveApplicationPhases(
       applicationId: parent.id,
     },
   });
-  return result!;
+  return result;
 }
 
 export async function resolveApplicationTags(parent: PrismaApplication): Promise<Tag[]> {
--- a/server/src/model/demonstration/demonstrationResolvers.ts
+++ b/server/src/model/demonstration/demonstrationResolvers.ts
@@ -191,10 +191,10 @@ export async function __resolveDemonstrationState(
   parent: PrismaDemonstration
 ): Promise<PrismaState> {
   // State can never be null thanks to the database
-  const result = await prisma().state.findUnique({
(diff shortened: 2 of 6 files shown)


server/src/model/application/applicationResolvers.ts
```diff
@@ -30,7 +30,7 @@ export async function resolveApplicationPhases(
       applicationId: parent.id,
     },
   });
-  return result!;
+  return result;
```
connor-parke-icf, line 33: i feel like 95% of the time the exclamation point should not be used

server/src/model/demonstration/demonstrationResolvers.ts
```diff
@@ -191,10 +191,10 @@ export async function __resolveDemonstrationState(
   parent: PrismaDemonstration
 ): Promise<PrismaState> {
   // State can never be null thanks to the database
-  const result = await prisma().state.findUnique({
+  const result = await prisma().state.findUniqueOrThrow({
```
connor-parke-icf, line 194: im not entirely sure which is better. using findUniqueOrThrow, or using findUnique and doing a custom nullcheck.  either way, get that exclamation point outa here

connor-parke-icf (approved):

server/src/model/application/applicationResolvers.ts
TARehman: This one was me being too database-minded, IMHO. We should just use `findUniqueOrThrow` when we know there will be something, and accept empty lists from `findMany` otherwise. 👍🏻

server/src/model/demonstration/demonstrationResolvers.ts
TARehman: I think the custom nullchecks should only be done when we need to do something particular. Otherwise, just allow Prisma to toss it if needed.

PR #1097, opened 2026-04-17
Demos 1653 - Adds V1 validation function for BN notebooks.

james-becwar-cms:
This PR adds the "new" demos BN workbook validations to the share library and some psudo code on how it can be ran from the browser and the server.

Currently validation takes about 2 seconds on my machine.

Notes that one validation we are still waiting for clarification on how it should work, and I expect these validation rules to be modified once we start using with populated files in test.


shared_library/src/BN/rulesets/v1/index.ts
```diff
@@ -0,0 +1,253 @@
...
+    const A700 = excelColumnRow("A700", "C Report", data);
+
+    if (
+      A100 !== "Waiver Name" ||
+      A300 !== "Waiver Name" ||
+      A500 !== "Waiver Name" ||
+      A700 !== "Waiver Name"
+    ) {
+      return {
+        code: "1",
```
connor-parke-icf, line 23: might could use a natural key for this error code, so its easy to parse in a log or something.  "Waiver Name value misplaced" or something

shared_library/src/BN/rulesets/v1/index.ts
```diff
@@ -0,0 +1,253 @@
...
+        continue;
+      }
+      for (let colNumber =  excelColumnToNumber(columnStart); colNumber <= excelColumnToNumber(columnEnd); colNumber++) {
+        const col = numberToExcelColumn(colNumber);
+        const cellValue = excelColumnRow(`${col}${row}`, "C Report", data);
+        if (cellValue !== "" && cellValue !== null) {
+          return {
+            code: "4",
+            message:
+              "TODO: Error: C Report - Expenditure data is present for DY(s) that extend beyond the Reporting DY.\n\nPlease only enter data through the current Reporting DY.",
```
connor-parke-icf: its not entirely clear if TODO here is a dev note or part of the actual error string; maybe a typo?

shared_library/src/BN/rulesets/v1/index.ts
```diff
@@ -0,0 +1,253 @@
...
+
+  function error5(data: ExcelData) {
+    const D2 = excelColumnRow("D2", "C Report", data);
+    if (D2 === "" || D2 === null) {
+      return {
+        code: "5",
+        message:
+          "Error: C Report - 'Data Pulled On' field is blank.\n\nPlease enter the date the data was pulled in cell D2.",
+      };
+    }
```
connor-parke-icf, line 92: might be good to have a centralized mapper of error code to error message somewhere, so if we have to look something up we dont need to search across multiple files.

shared_library/src/BN/index.test.ts
```diff
@@ -94,4 +94,24 @@ describe("excelColumnRow", () => {
       "Row 99 not found",
     );
   });
+
+  it("numberToExcelColumn converts numbers to Excel column letters correctly", () => {
+    expect(numberToExcelColumn(0)).toBe("A");
```
connor-parke-icf, line 99: worth noting here that Excel is 1 indexed, so it may be more appropriate if "A" corresponded with 1. 

the cell A1 would be (0, 1) here, which feels off. Unless we are subtracting 1 from the row number as well.

connor-parke-icf (approved):
this seems good; i recognize all of the rules and what theyre for! i think there are some organizational adjustments that could be made, but with a grain of salt:
* itd be nice to have the error codes be natural keys of some kind, so theyre easier to lookup
* a centralized mapping of error code to error description would be nice
* breaking up the different validation rules into their own file in a "rules" folder i think would help make each rule more atomic
* something feels off about the column letter - number conversions, i think its because excel is 1-indexed. May or may not be applicable, but itd be nice for the conventions to match

shared_library/src/BN/index.test.ts
james-becwar-cms: Its mapping the A1 to a square array, so A1 should map to [0],[0].  I have another test that reads an actual file to make sure I got them right, lol.

shared_library/src/BN/rulesets/v1/index.ts
james-becwar-cms: Its in the shared_library/validation_rules.xlsx file.  Its a copy of the file that Vivian has been working on.

shared_library/src/BN/rulesets/v1/index.ts
james-becwar-cms: Good catch, thank you.

shared_library/src/BN/rulesets/v1/index.ts
james-becwar-cms: its what Vivian has been using to talk about them with CMS, its in the file: shared_library/validation_rules.xlsx

PR #1100, opened 2026-04-20
Demos 1988 test cleanup

connor-parke-icf:
minimal functional changes here, mainly cleanup:

- query tests reformatted to more consistently mock out prisma queries
- Data functions updated to utilize an optional transaction parameter
- resolver tests updated to wrap each field in its own "describe" block, such that resolver tests match resolver structure. Eventually, the entire resolver test file should match this structure. 
- added explicit "type" keyword to index.ts files where needed

### DIFF
--- a/server/src/auth/index.ts
+++ b/server/src/auth/index.ts
@@ -1,7 +1,7 @@
-export { GraphQLContext } from "./auth.util";
-export { ContextUser } from "./userContext";
+export { type GraphQLContext } from "./auth.util";
+export { type ContextUser } from "./userContext";
 export {
   buildAuthorizationFilter,
   isStatePointOfContactOnDemonstration,
-  PermissionFilters,
+  type PermissionFilters,
 } from "./buildAuthorizationFilter";
--- a/server/src/auth/index.ts
+++ b/server/src/auth/index.ts
@@ -1,0 +1,7 @@
+export { GraphQLContext } from "./auth.util";
+export { ContextUser } from "./userContext";
+export {
+  buildAuthorizationFilter,
+  isStatePointOfContactOnDemonstration,
+  PermissionFilters,
+} from "./buildAuthorizationFilter";
--- a/server/src/model/amendment/amendmentData.test.ts
+++ b/server/src/model/amendment/amendmentData.test.ts
@@ -64,10 +64,27 @@ describe("amendmentData", () => {
 
       expect(buildAuthorizationFilter).toHaveBeenCalledOnce();
       expect(buildAuthorizationFilter).toHaveBeenCalledWith(user, expect.any(Function));
-      expect(selectAmendment).toHaveBeenCalledExactlyOnceWith({
-        AND: [where, authFilter],
-      });
+      expect(selectAmendment).toHaveBeenCalledExactlyOnceWith(
+        {
+          AND: [where, authFilter],
+        },
+        undefined
+      );
       expect(result).toBe(amendment);
+    });
+
+    it("passes transaction client to selectAmendment if provided", async () => {
+      const mockTransactionClient = {} as any;
+      vi.mocked(buildAuthorizationFilter).mockReturnValueOnce(authFilter);
+
+      await getAmendment(where, user, mockTransactionClient);
+      expect(buildAuthorizationFilter).toHaveBeenCalledOnce();
+      expect(selectAmendment).toHaveBeenCalledExactlyOnceWith(
+        {
+          AND: [where, authFilter],
+        },
+        mockTransactionClient
+      );
     });
   });
 
@@ -91,10 +108,27 @@ describe("amendmentData", () => {
 
       expect(buildAuthorizationFilter).toHaveBeenCalledOnce();
       expect(buildAuthorizationFilter).toHaveBeenCalledWith(user, expect.any(Function));
-      expect(selectManyAmendments).toHaveBeenCalledExactlyOnceWith({
-        AND: [where, authFilter],
-      });
+      expect(selectManyAmendments).toHaveBeenCalledExactlyOnceWith(
+        {
+          AND: [where, authFilter],
+        },
+        undefined
+      );
       expect(result).toBe(amendments);
+    });
+
+    it("passes transaction client to selectManyAmendments if provided", async () => {
+      const mockTransactionClient = {} as any;
+      vi.mocked(buildAuthorizationFilter).mockReturnValueOnce(authFilter);
+
+      await getManyAmendments(where, user, mockTransactionClient);
+      expect(buildAuthorizationFilter).toHaveBeenCalledOnce();
+      expect(selectManyAmendments).toHaveBeenCalledExactlyOnceWith(
+        {
+          AND: [where, authFilter],
+        },
+        mockTransactionClient
+      );
     });
   });
 });
--- a/server/src/model/amendment/amendmentData.test.ts
+++ b/server/src/model/amendment/amendmentData.test.ts
@@ -1,20 +1,15 @@
 import { Amendment as PrismaAmendment, Prisma } from "@prisma/client";
 import { beforeEach, describe, expect, it, vi } from "vitest";
-import { buildAuthorizationFilter } from "../../auth/buildAuthorizationFilter.js";
-import { getAmendment, getManyAmendments } from "./amendmentData.js";
-import { selectAmendment } from "./queries/selectAmendment.js";
-import { selectManyAmendments } from "./queries/selectManyAmendments.js";
-import { ContextUser } from "../../auth/userContext.js";
+import { buildAuthorizationFilter, ContextUser } from "../../auth";
+import { getAmendment, getManyAmendments } from "./amendmentData";
+import { selectAmendment, selectManyAmendments } from "./queries";
 
-vi.mock("../../auth/buildAuthorizationFilter.js", () => ({
+vi.mock("../../auth", () => ({
   buildAuthorizationFilter: vi.fn(),
 }));
 
-vi.mock("./queries/selectAmendment.js", () => ({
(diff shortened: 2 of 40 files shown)


server/src/model/amendment/queries/selectAmendment.test.ts
```diff
@@ -8,56 +8,65 @@ vi.mock("../../../prismaClient", () => ({
...
-    amendmentFindAtMostOne.mockResolvedValueOnce(null);
-
+    regularMocks.amendment.findAtMostOne.mockResolvedValueOnce(null);
     const result = await selectAmendment(where);
-
+    expect(regularMocks.amendment.findAtMostOne).toHaveBeenCalledExactlyOnceWith(expectedCall);
     expect(result).toBeNull();
   });
+
+  it("returns amendment that is found", async () => {
```
connor-parke-icf, line 64: while this does seem like a redundant test, i actually think its important to keep in. Though it appears that this is just tersting that the mock is what the mock is defined, its only because the selectAmendment function is so simple. If it were to contain more complicated logic, that would not be the case. We're mocking the inner prisma select, and verifying that the selectAmendment correctly interprets the value of that return; it just happens that the interpreted value is identical to the returned value.

server/src/model/amendment/queries/selectAmendment.test.ts
```diff
@@ -8,56 +8,65 @@ vi.mock("../../../prismaClient", () => ({
 }));
 
 describe("selectAmendment", () => {
-  const amendmentFindAtMostOne = vi.fn();
-
+  const regularMocks = {
```
connor-parke-icf, line 11: restructured these mocks to align with the test structure of deliverables. Before, we were redefining a mock as a standalone function, created some redundancy; now each of the test functions consistently use the same set of mocks.

server/src/model/amendment/amendmentData.ts
```diff
@@ -34,14 +36,18 @@ export async function getAmendment(
...
 
-  return await selectAmendment({
-    AND: [where, authFilter],
-  });
+  return await selectAmendment(
+    {
+      AND: [where, authFilter],
+    },
+    tx
+  );
```
connor-parke-icf, line 44: realized that this function was not able to utilize a the select queries transactionally. Now takes in an optional tx param and forwards it to the appropriate query.

server/src/model/amendment/amendmentResolvers.test.ts
```diff
@@ -130,65 +130,84 @@ describe("amendmentResolvers", () => {
     mockPrismaClient.$transaction.mockImplementation((callback) => callback(mockTransaction));
   });
 
-  it("delegates `Query.amendment` to `amendmentData.getAmendment`", async () => {
-    await amendmentResolvers.Query.amendment(undefined, { id: "abc123" }, mockContext);
-    expect(getAmendment).toHaveBeenCalledExactlyOnceWith({ id: "abc123" }, mockUser);
+  describe("Query.amendment", () => {
```
connor-parke-icf, line 133: shifted some things around in these tests to create a structure of "describe Type.field" for each field. Eventually, all fields in the resolver tests will conform to this pattern as more types are refactored.

server/src/auth/index.ts
```diff
@@ -1,7 +1,7 @@
-export { GraphQLContext } from "./auth.util";
-export { ContextUser } from "./userContext";
+export { type GraphQLContext } from "./auth.util";
+export { type ContextUser } from "./userContext";
```
TARehman, line 2: NB: We've historically put the type exports in a separate section from the others so that we can directly use `export type`. That being said, we don't have any formal guidance on that, so I'm down to change or keep it the same, and it certainly doesn't block.

server/src/model/amendment/queries/selectAmendment.test.ts
TARehman: I agree with your general logic that there _can_ be value to this type of thing, specifically when you get a value, modify or format it a certain way, and then return it. It's just in this case that we're directly returning the mocked value, so this is, more or less, checking that the system doesn't do any modifications of what is returned. It does cover a small possibility of failure in the future (e.g. we change how the function works to modify things and then forget to update the test); I don't know that the juice is there for the squeeze, but it's already here so no reason to pull it out.

server/src/model/amendment/amendmentData.test.ts
```diff
@@ -64,11 +64,28 @@ describe("amendmentData", () => {
 
       expect(buildAuthorizationFilter).toHaveBeenCalledOnce();
       expect(buildAuthorizationFilter).toHaveBeenCalledWith(user, expect.any(Function));
```
TARehman, line 66: NB: For future reference, you can use `.toHaveBeenCalledExactlyOnceWith` to do the same thing as you've done here in a single call. Also, I'd normally suggest making some kind of transaction object to verify that its called / passed correctly, just because of how undefined works in JS, but this is fine. These tests feel somewhat redundant, since you're passing a transaction client in both cases, but it covers it okay so no concerns.

server/src/model/amendment/amendmentData.ts
TARehman: I might recommend moving the process of getting a client up the chain here. What I mean is that you can have `getManyAmendments` retrieve a Prisma client if one is not provided, and pass that to `selectAmendment`, just like you can have `selectAmendment` get a client if one is not provided as well.

Then, you don't have to worry about passing a potentially undefined parameter through the functions, and it makes the behavior of both the same; if a tx is provided, it uses it, and otherwise, it fetches a new client.

I don't think its "wrong" to do it this way, but I had this debate in some of the deliverable code and settled on "get a client if one is not provided" at all levels. I thought it was easier to understand, and would be more efficient in any cases where more than one tx-consuming function is requested inside (because you'd get a single client and use it for all transactions). This isn't relevant here because this is basically passing through to the `selectX` function, so I can see not including it, but it feels like the general pattern should be to retrieve it if it allows an optional tx and one is not provided.

server/src/model/demonstration/queries/selectManyDemonstrations.test.ts
```diff
@@ -8,74 +8,66 @@ vi.mock("../../../prismaClient", () => ({
...
+    demonstration: {
+      findMany: transactionMocks.demonstration.findMany,
+    },
+  } as any;
 
+  const testDemonstrationId = "demonstration-1";
+  const testDemonstrationId2 = "demonstration-2";
   const where = {
-    id: "demonstration-1",
+    id: testDemonstrationId,
```
TARehman, line 35: NB: It's a tiny bit confusing that the `where` filter used here in the test is something you'd probably never do, which is use selectMany to select one demonstration, but I don't think its worth changing really. Just the brief brain fuzz of "that where statement couldn't return two values).

server/src/model/demonstration/queries/selectManyDemonstrations.test.ts
```diff
@@ -8,74 +8,66 @@ vi.mock("../../../prismaClient", () => ({
...
+  };
+  const expectedCall = {
+    where: { id: testDemonstrationId },
   };
 
   beforeEach(() => {
     vi.clearAllMocks();
     vi.mocked(prisma).mockReturnValue(mockPrismaClient as never);
```
TARehman, line 43: NB: I don't think it matters but `as any` feels better here somehow.

server/src/model/demonstration/queries/selectManyDemonstrations.test.ts
```diff
@@ -8,74 +8,66 @@ vi.mock("../../../prismaClient", () => ({
...
-      { id: "demonstration-1" },
-      { id: "demonstration-2" },
-    ] as PrismaDemonstration[];
-    demonstrationFindMany.mockResolvedValueOnce(demonstrations);
+    const demonstrations = [{ id: testDemonstrationId }, { id: testDemonstrationId2 }] as PrismaDemonstration[];
+    regularMocks.demonstration.findMany.mockResolvedValueOnce(demonstrations);
 
     const result = await selectManyDemonstrations(where);
-
+    expect(regularMocks.demonstration.findMany).toHaveBeenCalledExactlyOnceWith(expectedCall);
```
TARehman, line 70: NB: Doesn't hurt anything but I wouldn't put assertions that aren't relevant to the `it` statement necessarily, because if that assertion fails, it'll cause two different tests to break and that could add a tiny amount of cognitive overhead in debugging because both test failures are caused by the same underlying issue. Unlikely to matter because if one is failing the other probably will be as well.

server/src/model/document/queries/selectManyDocuments.test.ts
```diff
@@ -8,65 +8,66 @@ vi.mock("../../../prismaClient", () => ({
...
+    document: {
+      findMany: transactionMocks.document.findMany,
+    },
+  } as any;
 
+  const testDocumentId = "document-1";
+  const testDocumentId2 = "document-2";
   const where = {
-    id: "document-1",
+    id: testDocumentId,
```
TARehman, line 35: NB: Same point about the test case not matching expectations as above; a where that can only return one row is fine because we're just testing mocks anyway, but it feels cognitively odd when you first read it.

TARehman (approved):
Left some comments, but in general this looks fine. My only comment would be that I think we should retrieve the Prisma client at the highest point in the call chain possible. So anywhere that takes an optional argument for a transaction, we should get the client there and then pass it down through all the functions called below if they also take optional transaction arguments. That being said, its not something to block over, just a comment.

server/src/model/amendment/amendmentData.ts
connor-parke-icf: discussed this; going to defer any changes here until we have a more concrete usecase since there are a number of different directions this can go.  Not sure which way is best

PR #1101, opened 2026-04-20
Demos 1986 row based auth application phase

connor-parke-icf:
More of the same, this time with applicationPhase.  this one was a little trickier to verify, but was able to manually remove the restrictions on Demonstration while keeping the restriction on ApplicationPhase. With that, was able to see that while i had access to every demonstration, i was only able to see the ApplicationPhases on the demonstration that was assigned to me. 

Followed the same patterns as introduced in the previous PR. Queries for the actual database fetching, Data functions for injecting permissions, and finally calling the data functions from the Resolvers. The only places this had to happen was in the Demonstration, Amendment, and Extension resolvers, as there is no way to query for applicationPhase directly; its always though another object.

### DIFF
--- a/server/src/model/applicationPhase/applicationPhaseResolvers.test.ts
+++ b/server/src/model/applicationPhase/applicationPhaseResolvers.test.ts
@@ -1,8 +1,6 @@
 import { describe, it, expect, vi, beforeEach } from "vitest";
 import {
   __resolveApplicationPhaseDates,
-  __resolveApplicationPhaseName,
-  __resolveApplicationPhaseStatus,
   applicationPhaseResolvers,
 } from "./applicationPhaseResolvers";
 import { ApplicationPhase as PrismaApplicationPhase } from "@prisma/client";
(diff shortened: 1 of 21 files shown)


server/src/model/applicationPhase/queries/index.ts
TARehman, line 1: NB: We've historically just put the queries in the index.ts file of the containing module in their own section (see for instance `deliverable/index.ts`). Not sure we want to change that.

server/src/model/applicationPhase/applicationPhaseResolvers.test.ts
```diff
@@ -119,18 +117,4 @@ describe("applicationPhaseResolvers", () => {
       expect(mockFindMany).toHaveBeenCalledExactlyOnceWith(expectedCall);
     });
   });
-
```
TARehman, line 122: Did we not add corresponding tests for the fact that these were moved to arrow functions?

TARehman (approved):
LGTM but I would add tests in the one place where I asked about it - seems that we got rid of the functions but didn't add corresponding tests for the new arrow functions. I know they're simple, but it should be a matter of just moving the existing tests to a slightly different invocation.

server/src/model/applicationPhase/queries/index.ts
connor-parke-icf: trying to introduce some stratification to the exports to discourage invoking the query functions directly, and opting to use the Data functions. This may change depending on the mutation implementation

server/src/model/applicationPhase/applicationPhaseResolvers.test.ts
connor-parke-icf: good catch, these are missing. i will add before merge

PR #1102, opened 2026-04-20
DEMOS-1795: Phase 3 Declare Incomplete clears dates in Phases 2 and 3

nunesmlucas:
## Bug

On the Completeness phase (Phase 3), clicking **Declare Incomplete** left stale dates showing in the date pickers:

- **Phase 2 (Application Intake):** "State Application Submitted Date", "Completeness Review Due Date" not cleared (QA follow-up).

I also found that after the initial fix, users couldn't type a new date into either phase while the phase was Incomplete — the input appeared to wipe on every keystroke.

## Why

The server's `declareCompletenessPhaseIncomplete` mutation correctly nullified all relevant DB dates. The problem was on the client:

- `CompletenessPhase.tsx` auto-fills "State Application Deemed Complete" from the uploaded `Application Completeness Letter` document's `createdAt`; the federal-comment dates derive from that.
- `ApplicationIntakePhase.tsx` auto-fills "State Application Submitted Date" from the uploaded `State Application` document's `createdAt`; "Completeness Review Due Date" derives as `+15 days`.

Because Declare Incomplete leaves the documents in place, the auto-calculations kept re-populating the pickers after refetch.

The first attempted fix hardcoded the displayed value to `""` when the phase was Incomplete, which also wiped any new input on each keystroke.

## Fix

Client-only. Two touched components follow the same pattern.

### `ApplicationIntakePhase.tsx`

1. `getApplicationIntakeComponentFromApplication` looks up the Completeness phase and passes a new `completenessPhaseStatus: PhaseStatus` prop.
2. Component derives `completenessIncomplete = completenessPhaseStatus === "Incomplete"`.
3. Display only gates the document-derived auto-fill, never the user's override:
   ```tsx
   const stateApplicationSubmittedDate =
     submittedDateOverride ||
     (completenessIncomplete ? "" : calculatedStateApplicationSubmittedDate);
   ```
   `PhaseSelector` unmounts Phase 2's component tree on navigation, so `submittedDateOverride` is always `""` on remount after Declare Incomplete — no hook or manual state reset needed.

### `CompletenessPhase.tsx`

1. `getApplicationCompletenessFromApplication` passes `completenessPhaseStatus: PhaseStatus`.
2. Component derives `completenessComplete` and `completenessIncomplete` from the status.
3. `calculateStateDeemedCompleteDate` returns `""` when the phase is Incomplete and no DB date exists — suppresses the stale document auto-fill without affecting the user's override.
4. Local `userSelectedStateDeemedCompleteDate` is cleared imperatively inside `handleDeclareIncomplete` after the mutation succeeds — Phase 3 stays mounted across the click, so we need to reset it explicitly to avoid the old typed value reappearing.

https://github.com/user-attachments/assets/9bd81f7d-0606-4eab-bd67-1bf1bb4dbaa0


client/src/components/application/phases/ApplicationIntakePhase.tsx
```diff
@@ -218,6 +220,7 @@ export const getApplicationIntakeComponentFromApplication = (
       tags={application.tags}
       setSelectedPhase={setSelectedPhase}
       phaseStatus={applicationIntakePhase.phaseStatus ?? "Not Started"}
+      completenessPhaseStatus={completenessPhase?.phaseStatus ?? "Not Started"}
```
connor-parke-icf: rather than default here; i feel like if we are unable to find the completenessPhase that should indicate a bigger issue and we should error, similar to how we handle the applicationIntakePhase

client/src/components/application/phases/CompletenessPhase.tsx
```diff
@@ -248,6 +242,7 @@ export const CompletenessPhase = ({
     showDeclareIncompleteDialog(async () => {
       try {
         await declareCompletenessPhaseIncomplete(applicationId);
+        setUserSelectedStateDeemedCompleteDate("");
```
connor-parke-icf, line 254: i feel like with this there may be some off stateful behavior going on. I would think that declaring incomplete would cause a refetch or something to cause the component to get rerendered and we wouldnt need to manually reset this date back to blank, but i could be wrong about that.

connor-parke-icf (approved):
looks good! i dont think theres anything blocking here, but i would recommend that if we fail to get a phase on a demonstration that we throw an error rather than defaulting. We should be able to confidently expect that all of the phases exist on the application. 

Also, i feel like we shouldnt have to manually reset the form data when declaring incomplete, but im not entirely sure. I would think that the refetch or invalidation would cause a fresh rerender

client/src/components/application/phases/ApplicationIntakePhase.tsx
```diff
@@ -246,16 +252,18 @@ export const ApplicationIntakePhase = ({
   const [submittedDateOverride, setSubmittedDateOverride] = useState<string>("");
 
   // Calculate the dates to display based on the following rules:
-  // 1. If a date was passed to this component (most likely finalized), use this
-  // 2. If the user has manually entered a date (submittedDateOverride), use this
-  // 3. If no manual date, calculate based on the latest createdAt date of State Application documents
+  // 1. If the user has manually entered a date (submittedDateOverride), use this
+  // 2. If a date was passed to this component (most likely finalized), use this
+  // 3. If no manual date and Completeness is Incomplete (downstream Declare Incomplete just
+  //    reset this phase), show blank — any document auto-fill is stale until the user re-enters.
```
tresdonicf, line 262: I wonder if this will ever be seen as a bug hahah.

I really don't think declaring incomplete needs to remove these dates as they'll just be changed or resubmitted anyway but 🤷

tresdonicf (approved):

client/src/components/application/phases/ApplicationIntakePhase.tsx
nunesmlucas: Just following orders :) and trying to avoid headache

client/src/components/application/phases/ApplicationIntakePhase.tsx
nunesmlucas: Good call, updated — throws now if either phase is missing. Matches the existing pattern

client/src/components/application/phases/CompletenessPhase.tsx
nunesmlucas: Refetch does rerender, but `userSelectedStateDeemedCompleteDate` is local `useState` — React only resets that on unmount or a `key` change, not on prop changes. Phase 3 stays mounted across the click, so the typed value would survive without the manual reset unfortunately

client/src/components/application/phases/ApplicationIntakePhase.tsx
tresdonicf: Definitely, just wondering what kinda odds I can get on this 👀😂

PR #1103, opened 2026-04-20
Demos 1982 application note application date

connor-parke-icf:
Same idea, just for Application Date and Application Note. With this, all fields on ApplicationPhase are covered!  Verified in a similar way, by granting permissions for demonstration, phases, and either applicationNotes or Dates depending on what im testing. Found that i could only see the applicationNotes or Dates of Demonstrations i was assigned to as a state user.


dustbuster (approved):
Would probs get @TARehman to double check all this. But looks good to me.

PR #1104, opened 2026-04-20
(1/4) DEMOS-1541: Collapse Comment Box

tresdonicf:
Per DEMOS-1541: This PR collapses the comment box while preserving the comment

**Note:** I also adjusted the `Textarea` name field to be a drop-in for the data-test-id to simplify this, we don't gain anything from prepending `textarea-` to it. I updated relevant tests to reflect this

https://github.com/user-attachments/assets/af0a96d9-d3dc-4ed2-9d34-e433aa5a0e6e

### DIFF
--- a/client/src/components/input/Textarea.tsx
+++ b/client/src/components/input/Textarea.tsx
@@ -51,7 +51,7 @@ export const Textarea: React.FC<TextareaProps> = ({
       <textarea
         id={name}
         name={name}
-        data-testid={`textarea-${name}`}
+        data-testid={name}
         className={`${TEXTAREA_BASE_CLASSES} ${getInputColors(validationMessage ?? "")}`}
         placeholder={placeholder ?? ""}
         required={isRequired ?? false}
--- a/client/src/pages/deliverables/DeliverableDetailsManagementPage.tsx
+++ b/client/src/pages/deliverables/DeliverableDetailsManagementPage.tsx
@@ -69,9 +69,9 @@ export const DeliverableDetailsManagementPage: React.FC = () => {
           <DeliverableInfoFields deliverable={data.deliverable} />
           <DeliverableButtons />
         </div>
-        <div className="grid grid-cols-4 w-full gap-2 flex-1">
-          <div className="col-span-3"><FileAndHistoryTabs /></div>
-          <div className="col-span-1"><CommentBox /></div>
+        <div className="flex w-full gap-2 flex-1">
+          <div className="flex-1"><FileAndHistoryTabs /></div>
+          <div><CommentBox /></div>
         </div>
       </div>
     </div>
--- a/client/src/pages/deliverables/sections/CommentBox.tsx
+++ b/client/src/pages/deliverables/sections/CommentBox.tsx
@@ -21,10 +21,10 @@ const CommentBoxHeader = ({ onCollapse }: { onCollapse: () => void }) => (
   </div>
 );
 
-const CommentBoxTextArea = () => {
+const CommentBoxTextArea = ({ value, onChange }: { value: string; onChange: (e: React.ChangeEvent<HTMLTextAreaElement>) => void }) => {
   return (
     <div className="flex-1">
-      <Textarea label="Comments" initialValue="" name={COMMENT_BOX_TEXT_AREA_NAME} onChange={() => {}} />
+      <Textarea label="Comments" initialValue={value} name={COMMENT_BOX_TEXT_AREA_NAME} onChange={onChange} />
     </div>
   );
 };
@@ -38,6 +38,7 @@ const CommentBoxHistory = () => (
 
 export const CommentBox = () => {
   const [isCollapsed, setIsCollapsed] = useState(false);
+  const [currentComment, setCurrentComment] = useState("");
 
   if (isCollapsed) {
     return (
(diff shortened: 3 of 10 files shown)


client/src/pages/deliverables/DeliverableDetailsManagementPage.tsx
```diff
@@ -69,9 +69,9 @@ export const DeliverableDetailsManagementPage: React.FC = () => {
           <DeliverableInfoFields deliverable={data.deliverable} />
           <DeliverableButtons />
         </div>
-        <div className="grid grid-cols-4 w-full gap-2 flex-1">
-          <div className="col-span-3"><FileAndHistoryTabs /></div>
-          <div className="col-span-1"><CommentBox /></div>
+        <div className="flex w-full gap-2 flex-1">
+          <div className="flex-1"><FileAndHistoryTabs /></div>
+          <div><CommentBox /></div>
```
connor-parke-icf, line 74: if im understanding this right, the strategy is to mount the component upfront so that it can persist its own state (open/close, textarea content). So even when hidden, the component isnt getting unmounted thus persists its data. This kinda reminds me of how a Dialog remains mounted, but closed, and the show and hide is basically just a change in rendering (though, we deliberately unmount our dialogs). makes sense to me!

client/src/pages/deliverables/sections/CommentBox.tsx
```diff
@@ -1,23 +1,30 @@
...
+    </SecondaryButton>
   </div>
 );
 
-const CommentBoxTextArea = () => {
+const CommentBoxTextArea = ({ value, onChange }: { value: string; onChange: (e: React.ChangeEvent<HTMLTextAreaElement>) => void }) => {
   return (
     <div className="flex-1">
-      <Textarea label="Comments" initialValue="" name={COMMENT_BOX_TEXT_AREA_NAME} onChange={() => {}} />
+      <Textarea label="Comments" initialValue={value} name={COMMENT_BOX_TEXT_AREA_NAME} onChange={onChange} />
```
connor-parke-icf, line 27: is "initialValue" here just what we call the "value"? not that theres anything on the page that would update the value in the textarea without user input, but seems a little different from the patterns we've used elsewhere (excluding the date inputs, which have proved special)

connor-parke-icf (approved):
LGTM! i hadnt really thought about persisting values in unrendered fields before, but this seems like a good way to do it. I think in all of our previous implements of fields on pages, we've been fine to fully unmount the component and lose what may have been in the field. For instance in our dialogs if you type something, close the dialog, then reopen it i would expect that whatever i typed before is lost (especially since we have the warning message). Same idea with navigating between tabs in the application workflow. Nice catch on making sure the closing and opening is just a rendering behavior and doesnt affect the value held in the state.

client/src/components/input/Textarea.tsx
```diff
@@ -51,7 +51,7 @@ export const Textarea: React.FC<TextareaProps> = ({
       <textarea
         id={name}
         name={name}
-        data-testid={`textarea-${name}`}
+        data-testid={name}
```
dustbuster, line 54: 👍  Much cleaner.

client/src/pages/deliverables/sections/CommentBox.tsx
```diff
@@ -1,23 +1,30 @@
...
 import { Textarea } from "components/input";
+import { CommentIcon } from "components/icons";
+import { SecondaryButton } from "components/button";
 
 export const COMMENT_BOX_NAME = "comment-box";
 export const COMMENT_BOX_TEXT_AREA_NAME = "textarea-comment-box";
+export const COLLAPSE_COMMENTS_BUTTON_NAME = "button-collapse-comments";
 
-const CommentBoxHeader = () => (
+const CommentBoxHeader = ({ onCollapse }: { onCollapse: () => void }) => (
```
dustbuster, line 12: Like ` = () => (` vs. `= ({ onCollapse }: { onCollapse: () => void }) => (`

But closures gonna clos.

client/src/pages/deliverables/sections/CommentBox.tsx
dustbuster: There is quite a few initialValues also there is a `value={stuff}` as well.

Image

Image

Not sure if this helps i just noticed this.

dustbuster (approved):

client/src/pages/deliverables/sections/CommentBox.tsx
tresdonicf: This seems to just be used in the beginning with the state being held internally and the callback propagating it back to the caller, this could probably go for a rename at some point though

client/src/pages/deliverables/DeliverableDetailsManagementPage.tsx
tresdonicf: I might not fully understand but I do think some of that behavior wouldn't be persisted and would always default (open / close) and as for the textarea content that is persisted by putting the value into a state variable, thus avoiding the implicit re-rendering / resetting

I could be misunderstanding the comment / question though, based on the highlighted section of code

connor-parke-icf: yeah sorry, was kinda thinking my way through this as i was typing lol.  The mounting of the CommentBox (and thus the state of its content) is not conditional to the expansion behavior, so opening and closing doesnt make the content go away. 

I could see an implementation of this where closing the panel would unmount the CommentBox and lose its current state, which wouldnt be desired

tresdonicf: Totally, I observed that behavior of it getting rid of the state at first when I went to take the first screencap and noticed "hmm we probably don't want that" so fixed it hahah

PR #1105, opened 2026-04-21
DEMOS 1806 Allow CMS or Admin user to have more than one contact role for a demonstration 

JeffreyM-GA:
https://github.com/user-attachments/assets/c695a21d-5c7b-47c8-9172-e293de558b59

What I've implemented is roughly:
Another row for the user can be added by simply searching for the user and adding as you would the first row
The "Contact Types" options for the new row will be filtered to not include existing contact types. EG - If a user already has "Project Officer" as a contact type, that contact type will not be presented as an option for the next row.
If a user would have no available contact types (all of their options have been exhausted), the user will no longer appear in search results and cannot be added to the table again

One concern I had was that if a user were to add a bunch of rows for the same person without selecting a contact type, the above logic gets flaky. Example: Jeff has 4 available contact types. You add Jeff as a contact 5x while leaving contact type as "Select Type...", and then go and select the type for the first 4 instances. The fifth row gets into an unresolvable state where there are no options available. I was talking with Dipika about this before each of us left for vacation, and she said:

 "I think they need to assign a type and add the contact to the list before they search for the same or a different contact"
I interpreted this as "The Search Field should be disabled if there are unassigned contact types". I thought this would be a little confusing without any indication of why the search field is being disabled, so I added a line that says "Please select a contact type for all contacts before adding more."

### DIFF
--- a/client/src/components/dialog/ManageContactsDialog.tsx
+++ b/client/src/components/dialog/ManageContactsDialog.tsx
@@ -186,7 +186,7 @@ export const ManageContactsDialog: React.FC<ManageContactsDialogProps> = ({
         (c) =>
           c.personId === personId &&
           c.contactType &&
-          c.id !== currentRowId // 👈 key fix
+          c.id !== currentRowId
       )
       .map((c) => c.contactType);
 
--- a/client/src/components/dialog/ManageContactsDialog.tsx
+++ b/client/src/components/dialog/ManageContactsDialog.tsx
@@ -1,4 +1,4 @@
-import React, { ChangeEvent, useCallback, useEffect, useMemo, useState } from "react";
+import React, { ChangeEvent, useCallback, useEffect, useMemo, useRef, useState } from "react";
 
 import { Button } from "components/button";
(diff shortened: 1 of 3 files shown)


client/src/components/dialog/ManageContactsDialog.tsx
```diff
@@ -161,12 +162,35 @@ export const ManageContactsDialog: React.FC<ManageContactsDialogProps> = ({
...
+    else options = optionsByRole.Default;
+
+    if (!personId) return options;
+
+    const usedRoles = selectedContacts
+      .filter(
+        (c) =>
+          c.personId === personId &&
+          c.contactType &&
+          c.id !== currentRowId // 👈 key fix
```
tresdonicf: Might not need this comment since it'll just be how works, could replace it with a comment about the implication of the comparsion

tresdonicf (approved):
We talked about this a bit in slack - it is a bit odd that the user can add another contact by "searching" for them but this is an HCD concern about how to best surface this functionality and not within the scope of this PR

PR #1106, opened 2026-04-21
Demos 1985 row based auth application tag

connor-parke-icf:
this is the first instance of a more complicated resolver/query change, since applicationTagAssignment is what needs to be queried, but we only care about the Tag. So, a type ApplicationTagAssignmentQueryResult is used to handle that composite type of both. Other than that though, this is largely the same as the rest of the permissions updates. 

This is one of likely 4 PRs for the handling of Tags and the related fields. next is DemonstrationTypes, then DeliverableDemonstrationTypes, then likely suggestedTags.

https://github.com/user-attachments/assets/469f167a-eb4e-4180-8da5-ca7cd012a579


server/src/model/applicationTagAssignment/queries/index.ts
```diff
@@ -0,0 +1,10 @@
+import {
+  ApplicationTagAssignment as PrismaApplicationTagAssignment,
+  Tag as PrismaTag,
+} from "@prisma/client";
+
+export { selectApplicationTagAssignment } from "./selectApplicationTagAssignment";
+export { selectManyApplicationTagAssignments } from "./selectManyApplicationTagAssignments";
+export type ApplicationTagAssignmentQueryResult = PrismaApplicationTagAssignment & {
+  tag: PrismaTag;
+};
```
connor-parke-icf, line 10: introduced a type for the result of the Prisma query, since we care about both the tag assignment and the tag itself.

server/src/model/applicationTagAssignment/queries/selectApplicationTagAssignment.ts
```diff
@@ -0,0 +1,16 @@
+import { Prisma } from "@prisma/client";
+import { prisma, PrismaTransactionClient } from "../../../prismaClient";
+import { ApplicationTagAssignmentQueryResult } from ".";
+
+export async function selectApplicationTagAssignment(
+  where: Prisma.ApplicationTagAssignmentWhereInput,
+  tx?: PrismaTransactionClient
+): Promise<ApplicationTagAssignmentQueryResult | null> {
+  const prismaClient = tx ?? prisma();
+  return await prismaClient.applicationTagAssignment.findAtMostOne({
```
connor-parke-icf, line 10: rather than performing the mapping here, we only care about mapping when formatting for GQL, so that has been moved to the resolver.  This just fetches

james-becwar-cms (approved):

PR #1110, opened 2026-04-21
(2/4) DEMOS-1541: Show Tabs under the CommentBox to only CMS / Admin Users (+ UserContext Improvements)

tresdonicf:
For this PR I focused on the idea of "showing an element to one type of user and not another". Ended up coming up with a bunch of improvements to the way we handle `currentUser` and adding some tooling / documenting future improvements as well. Almost net-negative change after all is said and done!

https://github.com/user-attachments/assets/e34251ad-035d-47d5-ab4d-dd3f202ae6e0

### DIFF
--- a/client/package.json
+++ b/client/package.json
@@ -62,6 +62,8 @@
     "depcheck": "depcheck",
     "dev": "vite --host",
     "dev:mocks": "VITE_USE_MOCKS=true npm run dev",
+    "dev:mocks:state": "VITE_MOCK_PERSON_TYPE=demos-state-user npm run dev:mocks",
+    "dev:mocks:admin": "VITE_MOCK_PERSON_TYPE=demos-admin npm run dev:mocks",
     "format": "prettier --write src",
     "format:check": "prettier --check src",
     "lint": "eslint src; npm run tsc",
--- a/client/package.json
+++ b/client/package.json
@@ -63,6 +63,7 @@
     "dev": "vite --host",
     "dev:mocks": "VITE_USE_MOCKS=true npm run dev",
     "dev:mocks:state": "VITE_MOCK_PERSON_TYPE=demos-state-user npm run dev:mocks",
+    "dev:mocks:cms": "VITE_MOCK_PERSON_TYPE=demos-cms-user npm run dev:mocks", 
     "dev:mocks:admin": "VITE_MOCK_PERSON_TYPE=demos-admin npm run dev:mocks",
     "format": "prettier --write src",
     "format:check": "prettier --check src",
--- a/client/src/components/user/UserContext.tsx
+++ b/client/src/components/user/UserContext.tsx
@@ -4,7 +4,7 @@ import { useAuth } from "react-oidc-context";
 import { Person, User } from "demos-server";
 import { Loading } from "components/loading/Loading";
 
-type CurrentUser = Pick<User, "id" | "username"> & {
+export type CurrentUser = Pick<User, "id" | "username"> & {
   person: Pick<Person, "id" | "personType" | "fullName" | "firstName" | "lastName" | "email">;
 };
 
@@ -61,4 +61,14 @@ export function getCurrentUser() {
   }
 
   return ctx;
+}
+
+export function TestUserProvider({
+  children,
+  currentUser,
+}: {
+  children: React.ReactNode;
+  currentUser: CurrentUser;
+}) {
+  return <Ctx.Provider value={{ currentUser }}>{children}</Ctx.Provider>;
 }
--- a/client/src/pages/DeliverablesPage.test.tsx
+++ b/client/src/pages/DeliverablesPage.test.tsx
@@ -1,22 +1,12 @@
 import React from "react";
 import { render, screen, fireEvent } from "@testing-library/react";
 import { beforeEach, describe, expect, it, vi } from "vitest";
-import * as UserContext from "components/user/UserContext";
-
 import { MockedResponse } from "@apollo/client/testing";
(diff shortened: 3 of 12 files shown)


client/src/pages/DemonstrationDetail/deliverables/DeliverablesTab.test.tsx
```diff
@@ -1,22 +1,12 @@
...
-import * as UserContext from "components/user/UserContext";
-import { mockUsers } from "mock-data/userMocks";
-
-vi.mock("components/user/UserContext", async (importOriginal) => {
-  const actual = await importOriginal<typeof UserContext>();
-  return {
-    ...actual,
-    getCurrentUser: vi.fn(),
-  };
-});
```
tresdonicf, line 19: Could unmock some of these tests now!

client/src/test-utils/TestProvider.tsx
```diff
@@ -36,12 +39,15 @@ export const TestProvider: React.FC<TestProviderProps> = ({
...
+  currentUser = developmentMockUser,
 }) => {
   return (
     <MemoryRouter initialEntries={routerEntries}>
       <ToastProvider>
         <MockedProvider mocks={mocks} addTypename={addTypename}>
-          {children}
+          <TestUserProvider currentUser={currentUser}>
+            {children}
+          </TestUserProvider>
```
tresdonicf, line 50: Can set the currentUser in test provider to test specific functionality now!

client/package.json
```diff
@@ -62,6 +62,9 @@
     "depcheck": "depcheck",
     "dev": "vite --host",
     "dev:mocks": "VITE_USE_MOCKS=true npm run dev",
+    "dev:mocks:state": "VITE_MOCK_PERSON_TYPE=demos-state-user npm run dev:mocks",
+    "dev:mocks:cms": "VITE_MOCK_PERSON_TYPE=demos-cms-user npm run dev:mocks", 
+    "dev:mocks:admin": "VITE_MOCK_PERSON_TYPE=demos-admin npm run dev:mocks",
```
tresdonicf, line 67: Allows the mock user to take on different shapes, can quickly see the differences between users

client/src/components/user/UserContext.tsx
```diff
@@ -62,3 +62,13 @@ export function getCurrentUser() {
 
   return ctx;
 }
+
+export function TestUserProvider({
```
tresdonicf, line 66: A thin wrapper that doesn't call any auth functionality, thus usable in tests

client/src/pages/DemonstrationDetail/deliverables/DeliverablesTab.test.tsx
dustbuster: You are such a bad test your mother was a submit button! Oh unmock. right.

dustbuster: I got monty python in my head. 

Image

client/src/pages/DeliverablesPage.test.tsx
```diff
@@ -51,11 +41,11 @@ const DELIVERABLES_TABLE_MOCKS: MockedResponse[] = [
...
-  const mockGetCurrentUser = vi.mocked(UserContext.getCurrentUser);
+  const DEFAULT_TEST_USER = { ...mockUsers[0], id: CURRENT_USER_ID };
 
-  const renderDeliverablesPage = async () => {
+  const renderDeliverablesPage = async (currentUser = DEFAULT_TEST_USER) => {
     render(
-      <TestProvider mocks={DELIVERABLES_TABLE_MOCKS}>
+      <TestProvider mocks={DELIVERABLES_TABLE_MOCKS} currentUser={currentUser}>
         <DeliverablesPage />
       </TestProvider>
```
dustbuster, line 50: I kinda like this!

client/package.json
dustbuster: `frenchguy:monty-python:mocks:silly-english-type`

dustbuster (approved):
Looks good to me!

client/src/pages/deliverables/sections/CommentBox.tsx
```diff
@@ -37,9 +47,17 @@ const CommentBoxHistory = () => (
 );
 
 export const CommentBox = () => {
+  const { currentUser } = getCurrentUser();
   const [isCollapsed, setIsCollapsed] = useState(false);
   const [currentComment, setCurrentComment] = useState("");
 
+  if (!currentUser) {
+    return null;
+  }
```
connor-parke-icf, line 56: hmm, is there a good way we can guarantee that the user exists on the context? i would love it if we could just assume that if it gets to this point in the rendering that there is definitely a usercontext so we dont need to null check in a bunch of places

connor-parke-icf (approved):
looks good to me! eventually, im planning on returning a "SystemRoles" and a "Permissions" field on the user/person object that we might can switch to.  Something like that might become more relevant once we have more than one Role type within an IDM role (like "readonly CMS User").  We can cross that bridge when we get to it though; this is great for supporting testing and at a glance rendering

client/src/pages/deliverables/sections/CommentBox.tsx
tresdonicf: I threw an item in the AGENTS.md for this - the tests kinda blew up when I removed the `?` but totally, if we can't find a current user than `<UserContext>` should itself just render `<div>Can't find user, something is wrong!</div>` instead of any children and down the line we should just be able to use it without any conditionality.

I'll take another crack at this at some point unless somebody else gets to it before I do!

PR #1111, opened 2026-04-21
Demos 1987 row based auth contacts

connor-parke-icf:
More permissions! this time, contacts. This does not include the Primary Project Officer field, since that returns a Person and the logic there is a little tricky (especially with how Person is used in logging/events). Same idea though, route through a Data layer to apply permissions. 

https://github.com/user-attachments/assets/8034a26c-87b0-4529-9455-d65055751cca


server/src/model/demonstrationRoleAssignment/queries/index.ts
```diff
@@ -0,0 +1,7 @@
+import { DemonstrationRoleAssignment as PrismaDemonstrationRoleAssignment } from "@prisma/client";
+
+export { selectDemonstrationRoleAssignment } from "./selectDemonstrationRoleAssignment";
+export { selectManyDemonstrationRoleAssignments } from "./selectManyDemonstrationRoleAssignments";
+export type DemonstrationRoleAssignmentQueryResult = PrismaDemonstrationRoleAssignment & {
+  isPrimary: boolean;
+};
```
connor-parke-icf, line 7: slight deviation from the pattern, but here i opted to do the isPrimary mapping in the prisma query function since the primaryDemonstrationRoleAssignment holds no actual data; its just an indicator of primary-ness. Theres no reason to hold the entire primaryDemonstrationRoleAssignment on an object _other_ than whether it exists or not.

james-becwar-cms (approved):

PR #1115, opened 2026-04-22
DEMOS-1956-modify-action-required-fields

TARehman:
Adds a configuration and check for when the extension ID is optional. This ensures that it is required in cases where we must be able to definitively link back to the extension, eg for extension-related actions.


server/src/model/migrations/20260422200131_required_extension_for_certain_actions/migration.sql
```diff
@@ -0,0 +1,29 @@
...
+        'Denied Extension Request',
+        'Withdrew Extension Request'
+    );
+
+ALTER TABLE
+    demos_app.deliverable_action
+ADD CONSTRAINT
+    require_extension_id_for_extension_actions
+CHECK (
+    (extension_id_optional = TRUE) OR (extension_id_optional = FALSE AND active_extension_id IS NULL)
```
connor-parke-icf, line 28: might be easier to read as "not the invalid condition", i.e. NOT (extension_id_optional = TRUE and active_extension_id is NULL) or something like that. doesnt matter either way

connor-parke-icf (approved):
.mmd changes appear to align with the prisma and the migration changes. odd not seeing these changes uncrunched after all that time on the deliverable feature branch; LGTM!

PR #1119, opened 2026-04-23
Demos 2013 application tag suggestion unique key fix

connor-parke-icf:
Remove the id column from application_tag_suggestion and application_tag_suggestion_extract
Update History tables appropriately
update UK and FK to include only value and application_id
update trigger upsert_suggestion_record to upsert on conflict of (application_id, value) instead of id
update resolvers and schema to require input of (application_id, value) instead of id
update associated tests
update mermaid diagram 
Add section in seeder to populate the database with mock suggestion records
Create new column "replaced_value" on application_tag_suggestion
create check constraint to validate that when status_id is "Replaced", replaced_value must not be null and vice verse.


server/src/sql/functions.sql
```diff
@@ -942,7 +942,8 @@ BEGIN
     FROM
         demos_app.application_tag_suggestion_extract
     WHERE
-        suggestion_id = OLD.suggestion_id
+        application_id = OLD.application_id
+        and value = OLD.value
```
TARehman: NIT: format / capitalization

```suggestion
        application_id = OLD.application_id AND
        value = OLD.value
```

server/src/sql/functions.sql
```diff
@@ -952,9 +953,10 @@ BEGIN
         FROM
             demos_app.application_tag_suggestion
         WHERE
-            id = OLD.suggestion_id
+            application_id = OLD.application_id
+            and value = OLD.value
```
TARehman: NIT: format / capitalization

```suggestion
            application_id = OLD.application_id AND
            value = OLD.value
```

server/src/sql/functions.sql
```diff
@@ -1016,9 +1016,10 @@ BEGIN
         FROM
             demos_app.application_tag_suggestion_extract
         WHERE
-          suggestion_id = NEW.id
+          application_id = NEW.application_id
+          and value = NEW.value
```
TARehman: NIT: format / capitalization

```suggestion
          application_id = NEW.application_id AND
          value = NEW.value
```

server/src/model/applicationTagSuggestionExtract/applicationTagSuggestionExtract.prisma
```diff
@@ -10,10 +9,10 @@ model ApplicationTagSuggestionExtract {
   updatedAt     DateTime @updatedAt @map("updated_at") @db.Timestamptz()
 
   // DEFERRED: will be manually replaced by new deferred key through migration
-  suggestion  ApplicationTagSuggestion                  @relation(fields: [suggestionId, applicationId, value], references: [id, applicationId, value])
+  suggestion  ApplicationTagSuggestion                  @relation(fields: [applicationId, value], references: [applicationId, value])
   uiPathValue UiPathValue                               @relation(fields: [uiPathValueId, applicationId, fieldId, value], references: [id, applicationId, fieldId, value])
   field       ApplicationTagSuggestionExtractFieldLimit @relation(fields: [fieldId], references: [id])
 
-  @@id([suggestionId, uiPathValueId])
+  @@id([uiPathValueId])
```
TARehman: The key here needs to be `uiPathValueId, applicationId, value` I think. That is what would replace `suggestionId`. I'm not sure if it makes a tremendous difference because of the fact that uiPathValueId is already 1:1 with `uipath_value`, but we'd essentially be relying on that 1:1 relationship to prevent the same value from being used in two different suggestions. I think we want to be more explicit about that.

server/src/model/applicationTagSuggestion/applicationTagSuggestionSchema.ts
```diff
@@ -11,8 +10,12 @@ export const applicationTagSuggestionSchema = gql`
   }
 
   type Mutation {
-    acceptApplicationTagSuggestion(suggestionId: ID!): Application
-    replaceApplicationTagSuggestion(suggestionId: ID!, newValue: String!): Application
-    removeApplicationTagSuggestion(suggestionId: ID!): ApplicationTagSuggestion
+    acceptApplicationTagSuggestion(applicationId: ID!, value: String!): Application
```
TARehman, line 13: NB: Should the input be something like TagName for the type and "tag" vs "value" for the arguments?

server/src/model/applicationTagSuggestion/applicationTagSuggestionResolvers.ts
```diff
@@ -52,17 +62,27 @@ export async function acceptApplicationTagSuggestion(
 
 export async function replaceApplicationTagSuggestion(
   _: unknown,
-  { suggestionId, newValue }: { suggestionId: string; newValue: string },
+  { applicationId, value, newValue }: { applicationId: string; value: string; newValue: string }
 ): Promise<PrismaApplication> {
   try {
     return await prisma().$transaction(async (tx) => {
       const suggestion = await tx.applicationTagSuggestion.findUniqueOrThrow({
```
TARehman, line 69: NB: Didn't we figure out that this is redundant because of the fact that update returns?

server/src/model/applicationTagSuggestion/applicationTagSuggestionSchema.ts
connor-parke-icf: here, i dont think so. Since its referencing a Suggestion's value, which is not itself a Tag yet.  kinda in between a generic string and a Tag.

server/src/model/applicationTagSuggestion/applicationTagSuggestionResolvers.ts
connor-parke-icf: yeah, but didnt want to go through and refactor stuff with this update. figured itd be better to leave as much in place for this PR. I can go back and adjust when doing mutato permissions.

server/src/sql/functions.sql
connor-parke-icf: i like the "AND" to be on the second line. idk why it just reads better to me. We do both formats in this file

server/src/model/applicationTagSuggestionExtract/applicationTagSuggestionExtract.prisma
connor-parke-icf: discussed, will adjust

server/src/model/applicationTagSuggestion/applicationTagSuggestionSchema.ts
connor-parke-icf: i actually thing "value" is more appropriate than the Tag scalar here, since its kinda in a weird nebulous "prospective tag" state. better to keep it aligned with that than overconstrain i think.

TARehman (approved):
LGTM! 🚀

PR #1121, opened 2026-04-23
DEMOS-1543 — File Management: Deliverables Management Page (Layout)

nunesmlucas:
Replaces the three "Coming Soon" placeholders on the Deliverable Details page
with real State Files, CMS Files, and History tab layouts per the Jira story
mockups. Layout-only work — mirrors the DEMOS-1573 / DEMOS-1800 pattern: tables
and buttons are wired up, but actual file/history mutations are intentionally
out of scope until the backend endpoints land.

image
image
image


client/src/pages/deliverables/sections/DeliverableFileTable.tsx
```diff
@@ -0,0 +1,108 @@
...
+  editAriaLabel: string;
+  deleteAriaLabel: string;
+  emptyMessage: string;
+  files: DeliverableFileRow[];
+  columns: TableProps<DeliverableFileRow>["columns"];
+  onAdd?: () => void;
+  onEdit?: (file: DeliverableFileRow) => void;
+  onDelete?: (fileIds: string[]) => void;
+  footer?: React.ReactNode;
+};
```
tresdonicf, line 31: This is a lot of props!! Based on the way the component is shared I can imagine they're all needed right now but this might give us some insight about where to put this shared boundary or perhaps breaking up the subcomponents so those can be shared instead.

It's okay for now but just want to mention this for the future.

client/src/pages/deliverables/sections/fileColumns.tsx
tresdonicf, line 1: We should definitely revisit these column definitions at some point, they can be so verbose and a bit obscure once you start getting into the `filterFn / meta` sort of keys

tresdonicf (approved):
Overall LGTM for stubbing out the place to implement the functionality! Do you have any screenshots for this one by chance?

dustbuster:
Can you please add a screenshot? I'm working on something really similar and would like to see if we are overlapping

PR #1123, opened 2026-04-24
Demos 1985 row based auth suggested application tags

connor-parke-icf:
aaaand now for suggested application tags. Same idea. With this one though, took the opportunity to remove a lot of the unneeded mocks from ApplicationResolvers.test.ts.

This one does contain a somewhat controversial change; in the prisma model an extra relation was added between application_tag_suggestion and application, since getting the application from the suggestion in order to assign permissions was going to be pretty difficult. This relation is already handled transitively, so no actual logical changes are occurring, but this is one of the few instances where we find use of such a shortcut.  I have opted to not include this in the mermaid diagram since its not part of the core relational logic

https://github.com/user-attachments/assets/72de1072-c851-4110-b5f4-dce43481688f


server/src/constants.ts
```diff
@@ -491,6 +491,8 @@ export const PERMISSIONS = [
   "View DeliverableDemonstrationTypes on Assigned Demonstrations",
   "View All DemonstrationRoleAssignments",
   "View DemonstrationRoleAssignments on Assigned Demonstrations",
+  "View All ApplicationTagSuggestions",
+  "View ApplicationTagSuggestions on Assigned Demonstrations",
```
TARehman, line 495: NB: Should this be assigned demonstrations, or assigned applications? Suggestions are applicable to extensions and amendments as well. ETA: I realize now that the job assignment is on a demonstration, and then the extensions and amendments are children of the demonstration, but something about this term feels a little off. Probably not worth changing right now, but a good note to look at later.

server/src/model/migrations/20260424125309_prisma_virtual_application_relation_on_suggestions/migration.sql
```diff
@@ -0,0 +1,4 @@
+SET search_path TO demos_app;
+
+-- AddForeignKey
+ALTER TABLE "application_tag_suggestion" ADD CONSTRAINT "application_tag_suggestion_application_id_fkey" FOREIGN KEY ("application_id") REFERENCES "application"("id") ON DELETE RESTRICT ON UPDATE CASCADE;
```
TARehman, line 4: NB: Is this reflected in the ER diagram? I'm of mixed feelings about it because it's entirely for ergonomics of writing Prisma queries - but I also like the drawing matching the actual DB layout. Maybe we could style the line or something?

server/src/model/applicationTagSuggestion/queries/selectApplicationTagSuggestion.test.ts
```diff
@@ -0,0 +1,84 @@
...
+  const where = {
+    applicationId: testApplicationId,
+  };
+  const expectedCall = {
+    where: { applicationId: testApplicationId },
+  };
+
+  beforeEach(() => {
+    vi.clearAllMocks();
+    vi.mocked(prisma).mockReturnValue(mockPrismaClient as never);
```
TARehman: NB: I usually use as any here, but I don't know if it matters. More of a stylistic thing.

server/src/model/application/application.prisma
```diff
@@ -17,6 +17,10 @@ model Application {
   events                    Event[]
   applicationTagAssignments ApplicationTagAssignment[]
 
+  // relation is already handled transitively, but explicitly adding aids in simplifying queries 
+  applicationTagSuggestions ApplicationTagSuggestion[]
+
+
```
TARehman: NIT: Extra line here?

server/src/model/applicationTagSuggestion/queries/index.ts
```diff
@@ -0,0 +1,3 @@
+export { selectApplicationTagSuggestion } from "./selectApplicationTagSuggestion";
+export { selectManyApplicationTagSuggestions } from "./selectManyApplicationTagSuggestions";
+
```
TARehman: NIT: Extra line here?

TARehman (approved):
All fine. A few things that aren't blocking or aren't important. 🚢 it! 🙂

server/src/constants.ts
connor-parke-icf: yeah, the naming convention is really annoying. Kinda optimizing for speed of copy/paste/find/replace. Like ultimately the permissions are all based on a user's relationship to a demonstration, but sometimes that relationship is transitive through a number of separate objects, leading to the names of the permissions being slightly misaligned.

server/src/model/migrations/20260424125309_prisma_virtual_application_relation_on_suggestions/migration.sql
connor-parke-icf: yeah, i intentionally did not include this in the ER since its purely for the sake of the application.  In the prisma, i try to highlight it as being supplemental, but its not a great solution, especially when looking at just the SQL.  Prob worth revisiting

server/src/model/applicationTagSuggestion/queries/selectApplicationTagSuggestion.test.ts
connor-parke-icf: tbh i have no idea which is more appropriate. im guessing "any" is better since its actually a thing that we care about returning.

PR #1135, opened 2026-04-24
Install shared_library before running npm-updates

cms-jesse:
The most recent npm-updates run failed because the shared_library dependencies hadn't been installed prior to running. This PR adds the install and also adds a check for updates in the shared_library package.

I have also updated the buildDiscarder configuration to delete Jenkins pipeline runs earlier for merge queue branches and PRs. The previous configuration would delete everything except the last build, so old builds would hang around longer than needed and take up too much Jenkins disk space. This update will make sure that merge queue branches are deleted immediately after completion, and PR runs will be deleted after 7 days.

### DIFF
--- a/pipelines/Jenkinsfile.npm-updates
+++ b/pipelines/Jenkinsfile.npm-updates
@@ -35,6 +35,7 @@ pipeline {
           npmUpdates(dir: 'lambdas/fileprocess')
           npmUpdates(dir: 'lambdas/deleteinfectedfile')
           npmUpdates(dir: 'lambdas/UIPath')
+          npmUpdates(dir: 'lambdas/budgetNeutrality')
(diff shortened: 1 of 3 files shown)


pipelines/Jenkinsfile.npm-updates
```diff
@@ -14,18 +14,28 @@ pipeline {
   }
 
   stages {
+    stage('Install shared library deps') {
+      steps {
+        dirCon('shared_library', 'node') {
+          sh 'npm ci'
```
james-becwar-cms, line 20: Do we need to build here too?

james-becwar-cms (approved):
Approve with a comment

dustbuster (approved):

PR #1140, opened 2026-04-24
Demos 1987 row based auth states, person, primaryProjectOfficer

connor-parke-icf:
This one is a little strange in that it doesnt actually add any new permissions.  All users have access to all states.  Honestly, State as its own object could probably be removed from the API entirely and it would make no difference. Regardless, keeping with the pattern established for all the other ones ive refactored it to follow suit. 

People on the other hand doesnt have permission either, but thats because of the use of "person" in action/history tables. if someone is unassigned from a demonstration, we still want anyone with access to a demonstration to be able to see other persons' actions. Nonetheless, updated the resolvers to use the established pattern to clean up similarly. In doing this, Demosntration.primaryProjectOfficer was updated to now support permissions. 

Finally, the searchPeople query was found to have a bug in it and does not currently have tests associated. So im going to open up a new ticket for it as there are clientside implications that should be considered (highly considering removing that query altogether and letting the frontend do the appropriate filtering, but ultimately depends on what that might look like with respect to Person permissions).

Funtionally, the only change this PR introduces is the permission on Demonstration.primaryProjectOfficer, which follows the permissions of DemonstrationRoleAssignment.


server/src/model/state/stateResolvers.ts
TARehman, line 1: I thought we were just going to nuke this entirely, as there's no use case for being able to get states. Is that still the plan?

TARehman (approved):
This looks fine to me, I just thought we were gonna nuke the state resolvers.

server/src/model/state/stateResolvers.ts
connor-parke-icf: it makes sense to remove if we're sure that we wont have any queries that are going to traverse object > state > object, or state > object. Im leaving it in now cause im hesistent to take away the functionality with all the other stuff permissions-wise at this moment, but i think it would be fine to remove in general

PR #1141, opened 2026-04-24
DEMOS-1956-add-history-records-for-deliverables

TARehman:
This PR adds the `deliverableActions` key to the `deliverable` object which will allow populating the history table in the front-end. This turned out to need a bit more code than originally expected largely to meet the formatting requirements for what is returned. It also led to a few permissions discussions between myself and @connor-parke-icf. What is here will need to be tweaked as we add additional permissions items to the system. Here's what's changing.

- Remove the `note` field from the `deliverable_extension` table. It just makes more sense to keep all the note field values in a single place, `deliverable_action`. This was a leftover from a previous iteration of the design.
- Added a function to `dateUtilities` to get a nice `MM/DD/YYYY` date string from an `EasternTZDate`. This supports formatting elsewhere in the response.
- Also updated `dateUtilities` to return the actual current time from the previously misnamed `getEasternNow` function. This type change cascaded to have minor affects on several other test files.
- Added a utility type for tests called `DeepPartial` which is useful for constructing partial objects for testing purposes that we later cast to the full object when passing in.
- Fixed a bug in the check constraint requiring an deliverable extension ID for certain kinds of deliverable actions.
- Added a custom scalar in GQL for the different types of deliverable actions.
- Refactored the query structure to better match to what @connor-parke-icf has been doing for the rest of the permission systems. Got rid of a previously used `getDeliverableDemonstrationTypes` function and used new queries instead.
- Fixed several typing and import issues in the test files that are not caught by `npm run build`. The tests themselves were not impacted, which is why this hasn't come up before.
- Added the `Deliverable.deliverableActions` key which calls the requisite functions. Also, I reordered the resolvers to put the demonstration types one higher to make it match the schema layout.
- Added functions to format the requested actions into the format specified by the ticket.
- Added a function to get the actions and pass them through the formatting before they are returned.
- Added new test cases throughout.

Tests passing in all areas, SonarQube had no issues. Exercised the API locally by creating a bunch of fake actions in my DB and then running the query.

```
query Query($deliverableId: ID!) {
  deliverable(id: $deliverableId) {
    id
    deliverableActions {
      actionType
      userFullName
      details
    }
  }
}
```

This returns:

```json
{
  "data": {
    "deliverable": {
      "id": "5a1c1f29-2f0a-4973-93f1-ec36e9697490",
      "deliverableActions": [
        {
          "actionType": "Created Deliverable Slot",
          "userFullName": "Bypassed User (Admin User)",
          "details": ""
        },
        {
          "actionType": "Marked as Past Due",
          "userFullName": "System Update",
          "details": "The deliverable is past its due date"
        },
        {
          "actionType": "Manually Changed Due Date",
          "userFullName": "Bypassed User (Admin User)",
          "details": "Old Due Date: 10/01/2026\nNew Due Date: 10/01/2026\nReason Details: Changed the due date"
        },
        {
          "actionType": "Requested Resubmission",
          "userFullName": "Bypassed User (Admin User)",
          "details": "Old Due Date: 10/01/2026\nNew Due Date: 10/31/2026\nReason Details: Please resubmit"
        },
        {
          "actionType": "Submitted Deliverable",
          "userFullName": "Bypassed User (Admin User)",
          "details": ""
        },
        {
          "actionType": "Started Review",
          "userFullName": "Bypassed User (Admin User)",
          "details": ""
        },
        {
          "actionType": "Approved Deliverable",
          "userFullName": "Bypassed User (Admin User)",
          "details": ""
        },
        {
          "actionType": "Received and Filed Deliverable",
          "userFullName": "Bypassed User (Admin User)",
          "details": ""
        },
        {
          "actionType": "Approved Extension Request",
          "userFullName": "Bypassed User (Admin User)",
          "details": "Approved Due Date: 10/01/2026"
        },
        {
          "actionType": "Withdrew Extension Request",
          "userFullName": "Bypassed User (Admin User)",
          "details": ""
        },
        {
          "actionType": "Denied Extension Request",
          "userFullName": "Bypassed User (Admin User)",
          "details": "Denial Reason: No extensions allowed sorry"
        },
        {
          "actionType": "Accepted Deliverable",
          "userFullName": "Nola Zieme (CMS User)",
          "details": ""
        },
        {
          "actionType": "Requested Extension",
          "userFullName": "Bridgette Becker (State User)",
          "details": "Current Due Date: 10/01/2026\nNew Due Date Requested: 04/24/2026\nReason: Technical Difficulties\nReason Details: Need an extension please"
        }
      ]
    }
  }
}
```


client/src/components/table/tables/DeliverableTable.tsx
```diff
@@ -19,7 +19,17 @@ import { getDeliverableFilterOptions } from "./deliverablesFilterOptions";
 
 export type DeliverableTableRow = Omit<
```
TARehman, line 20: Question: Is there a reason to use `Omit` here instead of `Pick`, especially while the API is still under active development and fields are being added or removed? Nothing is using `deliverableActions` so I didn't even think to check for build errors in `client` before pushing my code for review.

connor-parke-icf: I believe we are pushing towards using Pick over Omit in general for exactly that reason; lets us update the type definitions without having to propagate the change through the various usages.

server/src/model/applicationDate/createPhaseStartDate.test.ts
```diff
@@ -25,6 +25,10 @@ describe("createPhaseStartDate", () => {
       easternTZDate: new TZDate("2025-01-02T04:59:59.999Z"),
       isEasternTZDate: true,
     },
+    "Current Time": {
+      easternTZDate: new TZDate("2025-01-01T11:14:32.141Z"),
+      isEasternTZDate: true,
+    },
```
connor-parke-icf, line 31: still kinda wonder about the overall strategy of how we are propagating dates to the database. We want to transform the dates into a timestamp for storage, but we might be able to get all the information we need about what that timestamp should be closer to the interface with the database. Not worth changing, but would be a good item to review for later refactors.

server/src/model/deliverable/deliverableResolvers.test.ts
```diff
@@ -362,5 +374,14 @@ describe("deliverableResolvers", () => {
         ]);
       });
     });
+
+    describe("Deliverable.deliverableActions", () => {
+      it("should query the deliverable actions of the parent deliverable", async () => {
+        await deliverableResolvers.Deliverable.deliverableActions(
+          testDeliverable as PrismaDeliverable
```
connor-parke-icf, line 381: more and more i lean towards not having a single "test object" which all the tests pull from, opting instead of just define an object with the relevant data in the test itself.  keeps each test as self contained as possible. But i see merit to both strategies; certainly not a blocker

server/src/model/deliverable/deliverableResolvers.ts
```diff
@@ -125,6 +128,18 @@ export const deliverableResolvers = {
...
+        (deliverableDemonstrationType) => {
+          const { statusId, tagNameId, ...tag } =
+            deliverableDemonstrationType.demonstrationTypeTagAssignment.tag;
+          return {
+            ...tag,
+            tagName: tagNameId,
+            approvalStatus: statusId,
+          };
+        }
+      ),
```
connor-parke-icf, line 142: im kinda wishy-washy on how i originally implemented this, whether this should be a dedicated "resolver" method, or if in-lining it here is fine. ultimately, its just a mapping, which is within the pattern of what should be fine for just inlining, but its just a few lines more than id like. certainly not a blocker; youre just moving it

server/src/model/deliverable/deliverableResolvers.ts
```diff
@@ -147,17 +162,8 @@ export const deliverableResolvers = {
...
-            ...tag,
-            tagName: tagNameId,
-            approvalStatus: statusId,
-          };
-        }
-      ),
+    deliverableActions: async (parent: PrismaDeliverable): Promise<DeliverableAction[]> => {
+      return await getFormattedDeliverableActions(parent.id);
+    },
   },
```
connor-parke-icf, line 168: NB: we might not need to make this async since we're not doing anything with the awaited result. im not 100% on this, but awaiting deliverableResolver.Deliverable.deliverableActions has the same result regardless of whether we await the result of getFormattedDeliverableActions. 
```
deliverableActions: async (parent: PrismaDeliverable): Promise<DeliverableAction[]> => getFormattedDeliverableActions(parent.id),
```
i think would be fine

server/src/model/deliverable/updateDeliverableDemonstrationTypes.test.ts
```diff
@@ -35,25 +37,40 @@ describe("updateDeliverableDemonstrationTypes", () => {
...
-    {
-      tagName: "Free Insulin",
-      approvalStatus: "Approved" satisfies TagStatus,
-    },
-    {
-      tagName: "Low Cost Vitamin A Supplements for Newborns",
-      approvalStatus: "Unapproved" satisfies TagStatus,
-    },
-  ];
+  const mockDeliverableDemonstrationTypes: DeepPartial<DeliverableDemonstrationTypeQueryResult>[] =
```
connor-parke-icf, line 40: NB: im on the fence on whether i like the deepPartial vs defining the subtypes and building out the main type as a composite of those subtypes. I seems super useful, but also the fact that this doesnt exist in vanilla TS makes me think that it doesnt exist because its deliberately discouraged.  Not that ts is unwavering in its design wisdom...

server/src/model/deliverable/updateDeliverableDemonstrationTypes.ts
```diff
@@ -22,8 +20,8 @@ export async function updateDeliverableDemonstrationTypes(
 
   // Can turn these into a set because DB guarantees uniqueness
   const oldDemonstrationTypes: Set<TagName> = new Set(
-    (await getDeliverableDemonstrationTypes(deliverableId, tx)).map(
-      (demonstrationType) => demonstrationType.tagName
+    (await selectManyDeliverableDemonstrationTypes({ deliverableId: deliverableId }, tx)).map(
```
connor-parke-icf, line 23: nit: { deliverableId } here is the same as { deliverableId: deliverableId }, so could shorten this by like a token. not that it matters at all, or that explicitly mapping out the relation is better or worse. just ramblin as i go through this lol.

server/src/model/deliverableAction/queries/insertDeliverableAction.test.ts
```diff
@@ -3,7 +3,7 @@ import { describe, it, expect, vi, beforeEach, afterEach } from "vitest";
 import { DeliverableActionType as PrismaDeliverableActionType } from "@prisma/client";
 
 // Types
-import { InsertDeliverableActionInput } from "..";
+import { InsertDeliverableActionInput } from ".";
```
connor-parke-icf, line 6: nit: generally, from within the module itself ive been opting to import directly. That way the index is relegated to be used by everything outside the module, as bundle of exports for the container. Almost as if the index.ts is invisible to everything inside the folder, and is the main means of access to everything outside the folder. There are definitely exceptions to this (like a place to put a composite type used by a bunch of separate queries), but thats more or less the distinction ive been leaning towards.

server/src/model/deliverableAction/deliverableActionFormattingFunctions.ts
```diff
@@ -0,0 +1,97 @@
...
+  ["demos-state-user", "State User"],
+]);
+
+// Note: casts below are enforced by the database
+export function formatFullUserName(input: SelectManyDeliverableActionsRowResult): NonEmptyString {
+  if (input.user) {
+    const firstName = input.user.person.firstName;
+    const lastName = input.user.person.lastName;
+    const personType = PERSON_TYPE_DISPLAY_MAP.get(input.user.personTypeId as PersonType);
+    return `${firstName} ${lastName} (${personType})`.trim();
```
connor-parke-icf, line 26: NIT: not huge in having these trim functions anywhere except at the point of rendering/usage (mainly, frontend). If the database holds the truth, and the truth has a weird space at the end of it, is it wrong to propagate that space along? i.e. format on the way in, pull directly on the way out.

server/src/model/document/handleDeleteDocument.test.ts
```diff
@@ -26,7 +26,7 @@ describe("handleDeleteDocument", () => {
   const testDocumentId = "doc-123-456";
   const testApplicationId = "app-123-456";
 
-  const mockDeletedDocument: PrismaDocument = {
+  const mockDeletedDocument: Partial<PrismaDocument> = {
```
connor-parke-icf, line 29: is it worth going ahead and unexcluding the "test" files in the linter?

server/src/model/migrations/20260424173449_fix_check_on_extension_id/migration.sql
```diff
@@ -0,0 +1,13 @@
...
+    require_extension_id_for_extension_actions;
+
+ALTER TABLE
+    demos_app.deliverable_action
+ADD CONSTRAINT
+    require_extension_id_for_extension_actions
+CHECK (
+    (extension_id_optional = TRUE) OR
+    (extension_id_optional = FALSE AND active_extension_id IS NOT NULL)
+);
```
connor-parke-icf, line 13: i coulda sworn we had implemented this check constraint in a previous PR, but i may be getting mixed up with some other recently added check constraint

connor-parke-icf (approved):
LGTM! dont think anything blocking, mainly nits and NBs. Happy to discuss. Most of what im commenting is opinion and either way is totally fine, most just asserting context for some of those opinions

server/src/model/applicationDate/createPhaseStartDate.test.ts
TARehman: So, the server-side stuff basically "just works" because everywhere we deal with dates, we handle them as Eastern time. I'm curious what we'd want to change, really?

server/src/model/deliverable/deliverableResolvers.test.ts
TARehman: In general we're getting big enough that having a set of test objects might start to make sense, but obviously it's a big refactor. I would save that for once we've released because then the API is also more stable too.

server/src/model/deliverable/deliverableResolvers.ts
TARehman: This is just because the enclosing function doesn't do anything but call the one inside of it, so there's no difference between awaiting and not awaiting the one inside, since the external async will await anyway? I think that sounds right but also I don't think there's any harm to awaiting it?

server/src/model/deliverable/updateDeliverableDemonstrationTypes.test.ts
TARehman: I dunno, base TS has a lot of stuff omitted that is kind of surprising to me. But really, all this is doing is letting us do a single call, rather than three separate `Partial<X>` calls that we composite together.

server/src/model/deliverable/updateDeliverableDemonstrationTypes.ts
TARehman: Yeah, purely a style thing. Blame the Python guy in me. 🤪 "Explicit > implicit" and whatnot.

server/src/model/deliverableAction/queries/insertDeliverableAction.test.ts
TARehman: We don't really have a set approach to this yet. There may be a small reason to take one approach versus the other, dealing with mocking in the tests; I'm not sure how important it is, the behavior of mocks is a bit unpredictable to me in TS/JS.

My approach to tests has been to import most things that are being mocked from the `index.ts` file, and to import the item that is under test directly from the source file. That lets me define all the mocks in one place, and also avoids issues of only mocking some of the things from a module while testing other things. It works, I think, because the `index.ts` file just imports from the source, so when you mock it, it gets inserted into the chain at that point.

I'm not sure how important this is, I just know that I've had tests fail for seemingly inexplicable reasons, and when I looked more closely, I realized that the mocks were being applied to `.` but that the module was importing from the actual source.

server/src/model/deliverableAction/deliverableActionFormattingFunctions.ts
TARehman: The use of `trim` here is just to match 1:1 to how its done elsewhere in the server codebase.

server/src/model/document/handleDeleteDocument.test.ts
TARehman: I did that to check for issues, but it also popped on a bunch of stuff that I don't think we actually care about, so I think there's probably a middle ground of having a command or something we can run to check. VSCode identifies issues in open files for me, its just not wanting to open them all.

server/src/model/migrations/20260424173449_fix_check_on_extension_id/migration.sql
TARehman: We did, this was the one that was backward that I had to fix.

server/src/model/deliverable/deliverableResolvers.ts
connor-parke-icf: certainly no harm and yeah, only cause we're not like assigning the awaited result to a variable and using that for some other operation; we're just returning it.

server/src/model/deliverableAction/queries/insertDeliverableAction.test.ts
connor-parke-icf: ah yep, ive ran into that a bunch of times. Like changing something to import from the module but forgetting to update the test which is still importing from the file directly or something. so the mock "misses" and it tries to use the actual implementation. we definitely dont have a set convention for this, so either is totally fine; whichever is more ergonomic

server/src/model/migrations/20260424173449_fix_check_on_extension_id/migration.sql
connor-parke-icf: ahh right of course, ok great

PR #1143, opened 2026-04-27
Task: Remove @axe-core/cli from package.json

tresdonicf:
This package seems to be having a lot of trouble with its security posture lately. Since this is a dev tool and not an actual project dependency we can move this out of our `package.json` and into its own self-contained install via `npx`. Appears to preserve all existing functionality that was there (though it does only test the single page and not crawl so we'll want to address this at some point)

Crazy how many transitive dependencies this removed in the `package-lock`!

image

### DIFF
--- a/client/scripts/ally.sh
+++ b/client/scripts/ally.sh
@@ -4,12 +4,12 @@ set -e
 # @axe-core/cli is run via npx to avoid a global install.
 # It was removed from devDependencies due to security findings in axe-core.
 
-CI=${CI:-false}
+DEMOS_AXE_CI=${DEMOS_AXE_CI:-false}
 
 echo "Installing browser driver..."
 npx browser-driver-manager install chrome
 
-if [ "$CI" = "true" ]; then
+if [ "$DEMOS_AXE_CI" = "true" ]; then
   npx @axe-core/cli http://localhost:3000 \
     --save accessibility-report.json \
     --exit \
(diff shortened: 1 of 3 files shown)


client/scripts/ally.sh
tresdonicf, line 1: codified this into a script to document the removal reasons for this architecture and to modify it down the line as needed - just plucked what was already in the `package.json` out of here

client/scripts/ally.sh
```diff
@@ -0,0 +1,26 @@
...
+# @axe-core/cli is run via npx to avoid a global install.
+# It was removed from devDependencies due to security findings in axe-core.
+
+DEMOS_AXE_CI=${DEMOS_AXE_CI:-false}
+
+echo "Installing browser driver..."
+npx browser-driver-manager install chrome
+
+if [ "$DEMOS_AXE_CI" = "true" ]; then
+  npx @axe-core/cli http://localhost:3000 \
```
tresdonicf, line 13: Now this will just run via `npx` directly

TARehman (approved):
I think this is a good solution, smart thinking @tresdonicf. We only need this available as a developer tool so this should work fine I think. 🚀 LGTM!

connor-parke-icf (approved):
yeesh, did this module basically only serve as a way to run as a npm command and for that it needed all these dependencies? good fix, these security findings are getting pretty tiring

PR #1145, opened 2026-04-27
(2/2) DEMOS-1881: Admin Page SideNav

tresdonicf:
Per DEMOS-1881 This PR finishes adding the side nav to the admin page as well as not showing the button for the admin page if the user is not an admin.

image
Screenshot 2026-04-27 113010


client/src/components/card/Card.tsx
tresdonicf, line 1: Added a `Card` component to re-use instead of copying these styles all over the place, just replaced the `DeliverablesPage` instance for now

client/src/router/RequireAdmin.tsx
```diff
@@ -0,0 +1,13 @@
+import React from "react";
+import { Navigate } from "react-router-dom";
+import { getCurrentUser } from "components/user/UserContext";
+
+export const RequireAdmin: React.FC<{ children: React.ReactNode }> = ({ children }) => {
```
tresdonicf, line 5: A simple utility that redirects the user away from the `/admin` route if they're not an admin user.

client/src/mock-data/userMocks.ts
```diff
@@ -9,14 +9,27 @@ import { mockPeople, MockPerson } from "./personMocks";
...
 export const developmentMockUser: MockUser = {
   id: "999",
   username: "mock.dev.user",
   person: {
     id: "999",
-    firstName: "Mock",
+    firstName: getPrettyFirstName(getMockPersonType()),
     lastName: "User",
-    fullName: "Mock User",
+    fullName: `${getPrettyFirstName(getMockPersonType())} User`,
```
tresdonicf, line 32: Just so we which type of user we're "logged in" as when using the mocks

client/src/components/header/ProfileBlock.tsx
```diff
@@ -4,6 +4,8 @@ import { Avatar } from "./Avatar";
 import { getCurrentUser } from "components/user/UserContext";
 import { SignoutLink } from "../auth/AuthLinks";
 
+export const PROFILE_BLOCK_TEST_ID = "profile-block";
```
james-becwar-cms, line 7: What is this doing?

james-becwar-cms (approved):

client/src/components/header/ProfileBlock.tsx
tresdonicf: Good question! this stabilizes the reference for it to be used in tests - in this case it had two benefits:

1. I moved the test to use `getByTestId()` instead of `findByText()` which is in itself a healthy change (unless of course you are trying to assert that some text exists)
2. The `getByTestId` can directly refer to the `PROFILE_BLOCK_TEST_ID` exported from this file instead of declaring it in two places where it may drift over time

dustbuster (approved):
ok

PR #1148, opened 2026-04-27
DEMOS-1625: Added some updates to the state user table

dustbuster:
Still need to finish the button and the link before this is merged as it will severely limit the SU access 

Screenshot 2026-04-27 at 13 50 18


dustbuster:
i think this is stale.

PR #1150, opened 2026-04-28
(3/4) DEMOS-1541: Show comments + Textarea Improvements

tresdonicf:
Per DEMOS-1541: This one adds the ability to add comments (locally until the endpoints are ready)

I simplified the  Textarea component types a bit as well,  now the handler it takes should just operate on a string. It also uses `value` instead of `initialValue`

https://github.com/user-attachments/assets/223e9900-d1bd-47bc-9b57-c873461a8af9

### DIFF
--- a/client/src/components/input/Textarea.tsx
+++ b/client/src/components/input/Textarea.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React from "react";
 
 import { tw } from "tags/tw";
 
@@ -14,8 +14,8 @@ const VALIDATION_MESSAGE_CLASSES = tw`text-error-dark`;
 export interface TextareaProps {
   name: string;
   label: string;
-  initialValue: string;
-  onChange: (e: React.ChangeEvent<HTMLTextAreaElement>) => void;
+  value: string;
+  onChange: (newString: string) => void;
   isRequired?: boolean;
   isDisabled?: boolean;
   placeholder?: string;
@@ -25,22 +25,18 @@ export interface TextareaProps {
 export const Textarea: React.FC<TextareaProps> = ({
   name,
   label,
-  initialValue,
+  value,
   onChange,
   isRequired,
   isDisabled,
   placeholder,
   getValidationMessage,
 }) => {
-  const [value, setValue] = useState(initialValue);
   const validationMessage = getValidationMessage ? getValidationMessage(value) : "";
   const rowsToDisplay = Math.min(Math.max(value.split("\n").length, 1), MAX_ROWS);
 
   const handleChange = (e: React.ChangeEvent<HTMLTextAreaElement>) => {
-    setValue(e.target.value);
-    if (onChange) {
-      onChange(e);
-    }
+    onChange(e.target.value);
   };
   return (
     <div className="flex flex-col gap-xs">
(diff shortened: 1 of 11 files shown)


client/src/components/input/Textarea.tsx
```diff
@@ -14,8 +14,8 @@ const VALIDATION_MESSAGE_CLASSES = tw`text-error-dark`;
 export interface TextareaProps {
   name: string;
   label: string;
-  initialValue: string;
-  onChange: (e: React.ChangeEvent<HTMLTextAreaElement>) => void;
+  value: string;
+  onChange: (newString: string) => void;
```
tresdonicf, line 18: a bit simpler here

client/src/pages/deliverables/sections/CommentBox.tsx
```diff
@@ -3,53 +3,92 @@ import React, { useState } from "react";
...
+const Comment = ({ comment }: { comment: CommentBoxComment }) => {
+  const formattedDate = format(comment.timestamp, "MM/dd/yyyy, hh:mm:ss a");
 
-const CommentBoxHistory = () => (
+  return (
+    <div className="p-1 bg-gray-secondary-layout rounded">
+      {comment.userFullName} - {formattedDate}
+      {comment.comment}
+    </div>
+  );
```
nunesmlucas, line 73: Can we space this out a bit? Stacked.
Lucas Nunes - DateComment

client/src/pages/deliverables/sections/CommentBox.tsx
```diff
@@ -3,53 +3,92 @@ import React, { useState } from "react";
...
 };
 
-const CommentBoxTabs = () => (
-  <div data-testid={COMMENT_BOX_TABS_NAME}>
-    ONLY CMS / ADMIN USERS CAN SEE THIS
-    {/* TODO: implement tab content */}
-  </div>
-);
+const Comment = ({ comment }: { comment: CommentBoxComment }) => {
+  const formattedDate = format(comment.timestamp, "MM/dd/yyyy, hh:mm:ss a");
```
nunesmlucas, line 66: Dont we have formatDate util we can use? I think the `seconds` are aggressive lol.

client/src/pages/deliverables/sections/CommentBox.tsx
```diff
@@ -3,53 +3,92 @@ import React, { useState } from "react";
...
+  if (!currentUser) {
+    throw new Error("Current user is required to render CommentBoxTextArea");
+  }
+
+  const handleAddComment = () => {
+    if (currentComment.trim() !== "") {
+      addComment({
+        comment: currentComment,
+        userFullName: currentUser.person.fullName,
+        timestamp: new Date(),
```
nunesmlucas, line 51: `getTodayEst()` ? could be wrong but just trying to minimize

client/src/pages/deliverables/sections/CommentBox.tsx
tresdonicf: Yeah that just formats the date without the time, you're right we don't need the seconds, I'll make this change next PR!

client/src/pages/deliverables/sections/CommentBox.tsx
tresdonicf: This one we want to create a real `new Date()` since it's a timestamp (needs hour + minute) is for making sure we get the right single day in EST at the time but doesn't give back time info

client/src/pages/deliverables/sections/CommentBox.tsx
tresdonicf: Yeah that's coming in the next PR! The final one here will just be polish / weird stuff I've noticed (like pressing enter when you're in the comment box should _probably_ add the comment)

nunesmlucas (approved):

PR #1151, opened 2026-04-28
(3/2) DEMOS-1881: "Close Admin" Navigates back to previous URL

tresdonicf:
New requirements came in for DEMOS-1881 to navigate back to the URL the user was previously at instead of home,  this does that. Also adds the settings icon to the header (not in screencap)

https://github.com/user-attachments/assets/cfe02326-d20b-463f-ab86-f49971541ef5

### DIFF
--- a/client/src/pages/admin/AdminHeader.tsx
+++ b/client/src/pages/admin/AdminHeader.tsx
@@ -9,12 +9,12 @@ export const AdminHeader: React.FC = () => {
   return (
     <header className="w-full">
       <div className="w-full flex items-center justify-between">
-        <span className="text-2xl font-bold">Admin Dashboard</span>
+        <span className="text-xl font-bold">Admin</span>
         <IconButton
           icon={<ExitIcon />}
           name="close-admin"
           data-testid="close-admin"
-          onClick={() => navigate("/")}
+          onClick={() => navigate(-1)}
         >
           Close Admin
         </IconButton>
--- a/client/src/pages/admin/AdminHeader.tsx
+++ b/client/src/pages/admin/AdminHeader.tsx
@@ -1,7 +1,7 @@
 import React from "react";
 import { useNavigate } from "react-router-dom";
 import { IconButton } from "components/button";
-import { ExitIcon } from "components/icons";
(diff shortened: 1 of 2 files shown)


client/src/pages/admin/AdminHeader.tsx
```diff
@@ -1,20 +1,22 @@
...
-        <span className="text-2xl font-bold">Admin Dashboard</span>
+        <div className="text-xl font-bold flex gap-0-5 items-baseline">
+          <SettingsIcon />Admin
+        </div>
         <IconButton
           icon={<ExitIcon />}
           name="close-admin"
           data-testid="close-admin"
-          onClick={() => navigate("/")}
+          onClick={() => navigate(-1)}
```
nunesmlucas, line 19: LGTM - super simple, I haven't used this yet i dont think.

client/src/pages/admin/AdminHeader.tsx
```diff
@@ -1,20 +1,22 @@
...
 export const AdminHeader: React.FC = () => {
   const navigate = useNavigate();
 
   return (
     <header className="w-full">
       <div className="w-full flex items-center justify-between">
-        <span className="text-2xl font-bold">Admin Dashboard</span>
+        <div className="text-xl font-bold flex gap-0-5 items-baseline">
+          <SettingsIcon />Admin
+        </div>
```
nunesmlucas, line 14: Is this going to be protected based on user profile? Is it implemented higher in index level already or driven through BE?

nunesmlucas:
With the whole "Close Admin" this almost seems like it would be a modal??? Just weird terminology. I would imagine "Back" would be more appropriate. Either way i think this is fine.

nunesmlucas (approved):

client/src/pages/admin/AdminHeader.tsx
tresdonicf: Currently it's driven by the `<DemosRouter>` component (has a `<RequiresAdmin>` route guard sort of component). As far as some of the B/E operations though I imagine those might additionally be role protected but I'm not certain about that

PR #1153, opened 2026-04-28
DEMOS-1959-adding-start-review-mutator

TARehman:
This PR adds the mutator which starts the review process. It also mildly refactors a few tests to use the `DeepPartial` type added earlier.

I'm pretty sure there's going to be some redundancy in how I get all of the various deliverable items implemented right now. This will be ripe for a refactor later to make it more code efficient; there will be check functions that we can combine into more general approaches, etc. However, in the interests of time, I am going to continue going forward with things that potentially could be better abstracted for now.

Works locally in the API, tests passing, SQ passing.

### DIFF
--- a/server/src/model/deliverable/checkDeliverableInputFunctions.ts
+++ b/server/src/model/deliverable/checkDeliverableInputFunctions.ts
@@ -4,9 +4,11 @@ import {
   DemonstrationTypeTagAssignment as PrismaDemonstrationTypeTagAssignment,
   User as PrismaUser,
 } from "@prisma/client";
+import { PrismaTransactionClient } from "../../prismaClient";
 import { ApplicationStatus, DeliverableStatus, PersonType, TagName } from "../../types";
 import { findDuplicates } from "../../validationUtilities";
 import { EasternTZDate, getEasternNow } from "../../dateUtilities";
+import { selectManyDocuments } from "../document";
 
 export function checkDemonstrationStatus(demonstration: PrismaDemonstration): string | undefined {
   const approvedStatus: ApplicationStatus = "Approved";
@@ -18,6 +20,7 @@ export function checkDemonstrationStatus(demonstration: PrismaDemonstration): st
 export function checkDeliverableStatusNotFinalized(
   deliverable: PrismaDeliverable
 ): string | undefined {
+  // Cast enforced by DB constraints
   const deliverableStatus = deliverable.statusId as DeliverableStatus;
   const finalDeliverableStatuses: DeliverableStatus[] = [
     "Approved",
@@ -25,12 +28,13 @@ export function checkDeliverableStatusNotFinalized(
     "Received and Filed",
   ];
   if (finalDeliverableStatuses.includes(deliverableStatus)) {
-    return `Cannot modify deliverable ${deliverable.id} as it has already been finalized.`;
+    return `Cannot submit or modify deliverable ${deliverable.id} as it has already been finalized.`;
   }
 }
 
 export function checkOwnerPersonType(ownerUser: PrismaUser): string | undefined {
   const permittedOwnerPersonTypes: readonly PersonType[] = ["demos-admin", "demos-cms-user"];
+  // Cast enforced by DB constraints
   if (!permittedOwnerPersonTypes.includes(ownerUser.personTypeId as PersonType)) {
     return `User ${ownerUser.id} is not a CMS user; cannot own deliverable.`;
   }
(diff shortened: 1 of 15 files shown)


server/src/model/deliverable/startDeliverableReview.ts
```diff
@@ -0,0 +1,45 @@
...
+import { DeliverableStatus, PersonType } from "../../types";
+import { prisma } from "../../prismaClient";
+import { editDeliverable, getDeliverable, validateStartDeliverableReviewInput } from ".";
+import { insertDeliverableAction } from "../deliverableAction/queries";
+
+export async function startDeliverableReview(
+  deliverableId: string,
+  context: GraphQLContext
+): Promise<PrismaDeliverable> {
+  // This probably will be modified when permissions are updated
```
TARehman, line 12: Note: In #1155 I generalize this so its consistent across the functions.

james-becwar-cms (approved):

server/src/model/deliverable/checkDeliverableInputFunctions.ts
```diff
@@ -83,3 +83,14 @@ export async function checkDeliverableHasAtLeastOneDocument(
     return `Cannot submit deliverable ${deliverable.id} because it has no state documents attached.`;
   }
 }
+
+export function checkDeliverableHasStatus(
```
connor-parke-icf, line 87: nit: maybe can rename cause this implies that we're checking whether or not the deliverable's status exists, rather than that it matches the expected status.

connor-parke-icf: also in a bunch of our other validation checks, i think we tend to just throw an error rather than returning a string.  Curious what pushed towards this pattern instead. 

Edit: nevermind, figured it out

server/src/model/deliverable/manuallyUpdateDeliverableDueDate.test.ts
```diff
@@ -127,7 +125,7 @@ describe("manuallyUpdateDeliverableDueDate", () => {
     await manuallyUpdateDeliverableDueDate(
       testDeliverableId,
       testInput,
-      testContext,
+      testContext as GraphQLContext,
```
connor-parke-icf, line 128: nit: something feels off to me about casting this at every usage instead of casting the variable up above as the saturated type. but i dont know what the resolution might be

server/src/model/deliverable/manuallyUpdateDeliverableDueDate.test.ts
```diff
@@ -234,7 +232,7 @@ describe("manuallyUpdateDeliverableDueDate", () => {
         note: testInput.dueDate!.dateChangeNote,
         oldDueDate: mockDeliverable.dueDate,
         newDueDate: testInput.dueDate!.newDueDate.easternTZDate,
-        userId: testContext.user.id,
+        userId: testContext.user!.id,
```
connor-parke-icf, line 235: hmm... this may be an issue with using the DeepPartial. like we know for sure that testContext has a user on it, because we define it so above. But because DeepPartial allows the fields to be undefined we need to add this assurance

server/src/model/deliverable/validateDeliverableInputs.ts
```diff
@@ -126,3 +127,22 @@ export async function validateSubmitDeliverableInput(
     });
   }
 }
+
+export function validateStartDeliverableReviewInput(deliverable: PrismaDeliverable): void {
+  const errors: (string | undefined)[] = [];
+
+  errors.push(checkDeliverableHasStatus(deliverable, "Submitted"));
```
connor-parke-icf, line 134: oh, ok i see why we would want the validator to return a string instead of throw

connor-parke-icf (commented):
LGTM, couple of nits but nothin blocking

server/src/model/deliverable/manuallyUpdateDeliverableDueDate.test.ts
TARehman: It is, but since we're only using it for tests, I don't _really_ care that much. I think the benefit of DeepPartial outweighs this here and there stuff in tests.

server/src/model/deliverable/manuallyUpdateDeliverableDueDate.test.ts
TARehman: The issue is that if you do `Partial` or whatever at the top and immediately cast it, essentially all the type checking immediately goes away anyway, so you might as well not even bother with partialness at all. By doing it in two steps it lets TS detect certain kinds of type mismatches even with the partial hackery.

PR #1155, opened 2026-04-28
DEMOS-1959-check-user-type-consistently

TARehman:
Refactor that pulls checking for whether the user is authorized into its own function. This should be reviewed after #1153 is merged as it is branched off that.

### DIFF
--- a/server/src/constants.ts
+++ b/server/src/constants.ts
@@ -1,5 +1,3 @@
-import { DateType, DocumentType, ExpectedTimestamp, PhaseName } from "./types.js";
-
 export const CLEARANCE_LEVELS = ["COMMs", "CMS (OSORA)"] as const;
 
 export const APPLICATION_STATUS = [
@@ -71,6 +69,15 @@ export const DELIVERABLE_STATUSES = [
   "Approved",
   "Received and Filed",
 ] as const;
+export type DeliverableStatus = (typeof DELIVERABLE_STATUSES)[number];
+
+export const FINAL_DELIVERABLE_STATUSES = [
+  "Accepted",
+  "Approved",
+  "Received and Filed",
+] as const satisfies DeliverableStatus[];
+
+export type FinalDeliverableStatus = (typeof FINAL_DELIVERABLE_STATUSES)[number];
 
 export const DEMONSTRATION_TYPE_STATUSES = ["Expired", "Pending", "Active"] as const;
 
@@ -153,6 +160,7 @@ export const DOCUMENT_TYPES = [
   "Summative Evaluation Report",
   "Transition Plan",
 ] as const;
+export type DocumentType = (typeof DOCUMENT_TYPES)[number];
 
 export const PHASE_NAMES = [
   "Concept",
@@ -164,6 +172,7 @@ export const PHASE_NAMES = [
   "Approval Package",
   "Approval Summary",
 ] as const;
+export type PhaseName = (typeof PHASE_NAMES)[number];
 
 export const PHASE_STATUS = [
   "Not Started",
@@ -228,8 +237,10 @@ export const DATE_TYPES = [
   "Approval Summary Start Date",
   "Approval Summary Completion Date",
 ] as const;
+export type DateType = (typeof DATE_TYPES)[number];
 
 export const EXPECTED_TIMESTAMPS = ["Start of Day", "End of Day"] as const;
+export type ExpectedTimestamp = (typeof EXPECTED_TIMESTAMPS)[number];
 
 type DateTypeExpectedTimestampRecord = Record<DateType, { expectedTimestamp: ExpectedTimestamp }>;
 
@@ -465,7 +476,7 @@ export const BUDGET_NEUTRALITY_VALIDATION_STATUSES = [
   "Failed",
   "Pending",
   "In Progress",
-];
+] as const;
 
 export const PERMISSIONS = [
   "View All Demonstrations",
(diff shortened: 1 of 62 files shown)


server/src/model/deliverable/createDeliverable.ts
```diff
@@ -1,7 +1,12 @@
 import { Deliverable as PrismaDeliverable } from "@prisma/client";
 import { CreateDeliverableInput } from "../../types";
 import { GraphQLContext } from "../../auth/auth.util";
-import { parseCreateDeliverableInput, validateCreateDeliverableInput, insertDeliverable } from ".";
+import {
+  parseCreateDeliverableInput,
+  validateCreateDeliverableInput,
+  insertDeliverable,
+  validateUserPersonTypeAllowed,
+} from ".";
```
james-becwar-cms, line 9: index.js might be more clear then from "."

james-becwar-cms (approved):

server/src/model/deliverable/startDeliverableReview.ts
```diff
@@ -1,20 +1,23 @@
...
   deliverableId: string,
   context: GraphQLContext
 ): Promise<PrismaDeliverable> {
-  // This probably will be modified when permissions are updated
-  const allowedPersonTypes: PersonType[] = ["demos-admin", "demos-cms-user"];
-  if (!allowedPersonTypes.includes(context.user.personTypeId)) {
-    throw new Error("Only Admin Users and CMS Users may start the review process.");
-  }
-
+  validateUserPersonTypeAllowed(context, "startDeliverableReview", [
```
connor-parke-icf, line 17: ahh, ok i see. this was offloaded to its own helper which explains why the context was different in the test. Nice

server/src/model/deliverable/validateDeliverableInputs.ts
```diff
@@ -16,6 +16,23 @@ import { getUser } from "../user";
 import { getDemonstrationTypeAssignments } from "../demonstrationTypeTagAssignment";
 import { GraphQLError } from "graphql";
 import { Deliverable as PrismaDeliverable } from "@prisma/client";
+import { GraphQLContext } from "../../auth";
+import { PersonType } from "../../types";
+
+// This probably will be modified when permissions are updated more generally
+// Temporary solution for deliverables
+export function validateUserPersonTypeAllowed(
```
connor-parke-icf, line 24: just verifying for later, this is a true Permission and not a constraint by the database yeah? i.e. its allowed by the database if an unpermitted user was to get through here

server/src/model/deliverableStatus/deliverableStatusSchema.ts
```diff
@@ -2,4 +2,5 @@ import gql from "graphql-tag";
 
 export const deliverableStatusSchema = gql`
   scalar DeliverableStatus
+  scalar FinalDeliverableStatus
```
connor-parke-icf, line 5: interesting, is this Type used in an input or just a type returned? im not seeing where we actually use this in the GQL schema

server/src/constants.ts
```diff
@@ -228,8 +237,10 @@ export const DATE_TYPES = [
   "Approval Summary Start Date",
   "Approval Summary Completion Date",
 ] as const;
+export type DateType = (typeof DATE_TYPES)[number];
```
connor-parke-icf, line 240: hmm... this feels a little odd to me that we're defining these types outside of the types.ts. What is the distinction for why these are here instead of in types.ts?

server/src/types.ts
```diff
@@ -125,6 +120,16 @@ export type { DeliverableAction } from "./model/deliverableAction/deliverableAct
 // Alias is not enforced in TS but maintains alignment between GQL and TS
 export type NonEmptyString = string; // NOSONAR: typescript:S6564
 
+// These types are defined in constants.ts and re-exported here to avoid a circular reference
+export type {
```
connor-parke-icf, line 124: i see... NB: this feels to me like there may be a way to structure these consts and types so we dont have to make an exception for these types

connor-parke-icf (approved):
LGTM, no blockers just some curiosities. seems like we're breaking convention in a few places and that might indicate a more general organizational change is on the horizon

server/src/model/deliverable/validateDeliverableInputs.ts
TARehman: That's...mostly true. There's certain restrictions on who can be assigned as a cmsOwner but in general this is a true permission.

server/src/model/deliverableStatus/deliverableStatusSchema.ts
TARehman: It's actually for my next PR, it was just done already when I decided to fix the way the permissions work. I'll be using it for when we complete a deliverable because the API will accept one of three options - Accepted, Approved, Received and Filed.

server/src/constants.ts
TARehman: We moved these specific ones here to avoid a circular reference where we create a variable in `constants.ts`, export it to `types.ts` to make a type, then import the type from there back into `constants.ts` to make a composite type (in this case, `DateType` is used in one of the mapping objects that assigns a timestamp type, IIRC). At one point SQ was complaining about it, so moving these here and just exporting them through `types.ts` avoids the issue.

server/src/types.ts
TARehman: We should just put both of them in one file, probably? ETA: Part of the issue is that types aren't real in TS so you really have to go from the constant to the type, instead of the other way around.

server/src/model/deliverableStatus/deliverableStatusSchema.ts
connor-parke-icf: ahhh ok yeah, i guess that is the one instance where the user needs to specify a status

PR #1156, opened 2026-04-28
DEMOS-1800 - Deliverable Modal Improvements

nunesmlucas:
- **Due Date cannot be in the past** in either Edit or Create New Deliverable modal. Save is disabled and an inline message ("Date must be on or after MM/DD/YYYY.") appears under the field.
- **Demonstration Types now persist** from Create → Edit. The modal pre-fills the previously saved tags instead of always opening empty.
- **Demonstration Types required rule** is conditional in both modals: required for *Implementation Plan* and *Monitoring Protocol*, optional otherwise.
- **DatePicker year-typing fix.** Switched the input from fully controlled (`value`) to uncontrolled (`defaultValue` + `useRef`/`useEffect` sync). Native `<input type="date">` reports `value=""` while the year is mid-typed; a controlled input was re-rendering on every keystroke and wiping the in-progress digits.
- **Min-date display fix.** Format the YYYY-MM-DD string directly to MM/DD/YYYY for the validation message, avoiding `new Date(string)` parsing that shifted the date back a day in EST.

https://github.com/user-attachments/assets/706fdfe5-d752-49d2-a32c-6a0d4daed953


client/src/components/dialog/deliverable/AddDeliverableSlotDialog.tsx
```diff
@@ -103,17 +105,22 @@ export const buildAddDeliverableSlotPayloads = (
...
-    ? data.dueDate.length > 0
-    : data.quarterlyDueDates.every((dueDate) => dueDate.length > 0);
+const hasValidDueDateForScheduleType = (
+  data: AddDeliverableSlotFormData,
+  today: string
+): boolean => {
+  if (data.scheduleType === "Single") {
+    return data.dueDate.length > 0 && !isBefore(data.dueDate, today);
+  }
+  return data.quarterlyDueDates.every((dueDate) => dueDate.length > 0);
```
connor-parke-icf: should we also check that the due dates are not before today in the quarterly due dates check? might not be a concern based on how the frontend is implemented

client/src/components/table/tables/DeliverableTable.tsx
```diff
@@ -75,6 +76,10 @@ export const DELIVERABLES_PAGE_QUERY = gql`
         }
       }
       dueDate
+      demonstrationTypes {
+        tagName
+        approvalStatus
+      }
```
connor-parke-icf, line 96: hmm, i feel like there might be a field name mismatch. looking at how we query deliverables.demonstration.demonstrationTypes vs deliverables.demonstrationTypes i would expect those to have the same field structure. but maybe theyre not the same schema type

client/src/components/input/date/DatePicker.tsx
```diff
@@ -27,8 +29,23 @@ export const DatePicker: React.FC<DatePickerProps> = ({
...
+  const inputRef = useRef<HTMLInputElement>(null);
+  const inputMin = minDate && isAfter(minDate, DEFAULT_MIN_DATE) ? minDate : DEFAULT_MIN_DATE;
+
+  // The input is uncontrolled (defaultValue + ref-sync) instead of fully controlled.
+  // Native <input type="date"> fires input events with value="" while the user is mid-typing the
+  // year subfield (the date is briefly incomplete). A controlled input would re-render on every
+  // such event and wipe the in-progress digits. With this ref-sync pattern, React only touches
+  // the DOM when the value prop changes externally.
+  useEffect(() => {
+    if (inputRef.current && inputRef.current.value !== (value ?? "")) {
```
connor-parke-icf, line 43: i see, so this is kinda a custom implementation of an onChange behavior? id be extra sure to validate this on a couple different browsers, since we've had issues with this component having divergent behaviors between browsers

connor-parke-icf (approved):
LGTM, i dont think anything really blocking id just make sure to verify the field list in the GQL and the date picker behavior on different browsers

client/src/components/input/date/DatePicker.tsx
```diff
@@ -48,16 +70,17 @@ export const DatePicker: React.FC<DatePickerProps> = ({
...
         name={name}
         data-testid={name}
         className={`${INPUT_BASE_CLASSES} ${getInputColors(validationMessage ?? "")}`}
         required={isRequired ?? false}
         disabled={isDisabled ?? false}
-        value={value ?? ""}
+        defaultValue={value ?? ""}
         onChange={handleChange}
-        min={DEFAULT_MIN_DATE}
+        min={inputMin}
```
tresdonicf, line 77: Can this line just be `minDate ?? DEFAULT_MIN_DATE` once the date is provided? This should prevent the datepicker from showing values less than that

client/src/components/input/date/DatePicker.tsx
```diff
@@ -39,7 +56,12 @@ export const DatePicker: React.FC<DatePickerProps> = ({
     }
   };
 
-  const validationMessage = getValidationMessage ? getValidationMessage() : "";
+  const externalValidationMessage = getValidationMessage ? getValidationMessage() : "";
+  const minViolationMessage =
+    value && minDate && isBefore(value, minDate)
+      ? `Date must be on or after ${formatDate(parseISO(minDate))}.`
+      : "";
+  const validationMessage = externalValidationMessage || minViolationMessage;
```
tresdonicf: Generally the reason to pass in the validation message function is to avoid having to adopt a pattern like this where we decide on an internal or external validation message.

So at the callsite we should be able to provide a function `getValidationMessage` that does check for the minimum date against what we want and surfaces that, versus adding it to this component directly.

tresdonicf (dismissed):
I'd like to see if we can keep our main`DatePicker` component a bit less complicated, we should be able to move the validation outside of the component itself and pass it in to be called based on the input.

This will just be a bit easier to maintain going forward

client/src/components/table/tables/DeliverableTable.tsx
nunesmlucas: `Demonstration.demonstrationTypes` is `[DemonstrationTypeAssignment!]!` (uses `demonstrationTypeName`), while `Deliverable.demonstrationTypes` is `[Tag!]!` (uses tagName). Different schema types, intentional.

client/src/components/input/date/DatePicker.tsx
nunesmlucas: https://jiraent.cms.gov/browse/DEMOS-2004
for that :/

client/src/components/table/tables/DeliverableTable.tsx
connor-parke-icf: ahhh nice ok. makes sense since for the deliverable we dont care about the effective/expiration date, only the tag itself

client/src/components/dialog/deliverable/fields/schedule-type/SingleDeliverableScheduleType.tsx
```diff
@@ -19,6 +22,12 @@ export const SingleDeliverableScheduleType = ({
           value={value}
           onChange={onChange}
           isRequired={true}
+          minDate={today}
+          getValidationMessage={() =>
+            value && isBefore(value, today)
+              ? `Date must be on or after ${formatDate(parseISO(today))}.`
+              : ""
+          }
```
tresdonicf, line 30: 🙌

client/src/components/dialog/deliverable/AddDeliverableSlotDialog.tsx
```diff
@@ -103,17 +105,24 @@ export const buildAddDeliverableSlotPayloads = (
...
+  if (data.scheduleType === "Single") {
+    return data.dueDate.length > 0 && !isBefore(data.dueDate, today);
+  }
+  return data.quarterlyDueDates.every(
+    (dueDate) => dueDate.length > 0 && !isBefore(dueDate, today)
+  );
+};
 
-const formIsValid = (data: AddDeliverableSlotFormData): boolean =>
+const formIsValid = (data: AddDeliverableSlotFormData, today: string): boolean =>
```
tresdonicf: NB: I imagine this function could call `getTodayEst` itself instead

client/src/components/input/date/DatePicker.tsx
```diff
@@ -48,16 +64,17 @@ export const DatePicker: React.FC<DatePickerProps> = ({
...
+        ref={inputRef}
         type="date"
         id={name}
         name={name}
         data-testid={name}
         className={`${INPUT_BASE_CLASSES} ${getInputColors(validationMessage ?? "")}`}
         required={isRequired ?? false}
         disabled={isDisabled ?? false}
-        value={value ?? ""}
+        defaultValue={value ?? ""}
```
tresdonicf, line 75: I'm not sure we want `defaultValue` here - could you help me understand if we do? This goes hand in hand with I'm not sure we need the ref stuff anymore, it's no longer used...

client/src/components/input/date/DatePicker.tsx
```diff
@@ -27,8 +28,23 @@ export const DatePicker: React.FC<DatePickerProps> = ({
...
+  // The input is uncontrolled (defaultValue + ref-sync) instead of fully controlled.
+  // Native <input type="date"> fires input events with value="" while the user is mid-typing the
+  // year subfield (the date is briefly incomplete). A controlled input would re-render on every
+  // such event and wipe the in-progress digits. With this ref-sync pattern, React only touches
+  // the DOM when the value prop changes externally.
+  useEffect(() => {
+    if (inputRef.current && inputRef.current.value !== (value ?? "")) {
+      inputRef.current.value = value ?? "";
+    }
+  }, [value]);
```
tresdonicf, line 46: I think we can get rid of all this, it seems to address the issue that was created by using `defaultValue` instead of `value` but if we go back to `value` we shouldn't need it. This component should only need one change as far as I can tell:

`minDate={minDate ?? DEFAULT_MIN_DATE}`, while the business logic (the validation message) is handled outside of it

tresdonicf (dismissed):
I'm a bit confused about using `defaultValue` instead of `value` in the Datepicker, didn't we just change this from using `defaultValue` in #924 ?

client/src/components/input/date/DatePicker.tsx
nunesmlucas: Tried `value={value}` without the ref-sync — year typing breaks in Chrome. While the date is incomplete, native `<input type="date">` fires input events with `e.target.value === ""`, so the parent state goes to `""`, React re-renders, and the controlled write clobbers the digits the user just typed. The `defaultValue` + ref-sync only touches the DOM when `value` changes externally, which preserves partial input.

Added a regression test so we don't revert this again, and applied the `formIsValid` / `getTodayEst` simplification — good call.

PR #1157, opened 2026-04-28
DEMOS-1959-complete-deliverable-mutator

TARehman:
This PR adds the ability to call `completeDeliverable` in the API to move a deliverable from `Under CMS Review` to `Approved`, `Accepted`, or `Received and Filed`. I thought I would need to add a separate TS check for whether there is an active extension but it turns out that the error at the DB level for that properly propagates up to the API level, so we only ended up needing to check that the user was an acceptable type and that we were moving from the right status.


connor-parke-icf (approved):
lgtm, seeing a lot of the same conventions from the previous PRs. Nothing really to say that i havnt noted or discussed already in previous PRs!

PR #1159, opened 2026-04-29
DEMOS-1959-request-resubmission

TARehman:
This PR adds the `requestDeliverableResubmission` mutator to allow requesting resubmission. There's also a few small little cleanup items I rolled in, listed below.

- Tweaks how the `TagName` scalar is implemented with the `NonEmptyStringResolver` to ensure that the API documentation reporting out lists the right type.
- Refactors `validateInputDates.ts` to lower the cognitive complexity by breaking out the various different kind of date validations into separate function calls, encapsulating the looping within.
- Removes two empty tests from `startPhaseByDocument.test.ts` that were leftover remnants of a previous version that did this; the tests have no assertions and upon reading this functionality was moved to the database.
- Removes an unnecessary assertion from `startPhaseByDates.ts`; this type was already `PhaseName` in the source object, making the as statement redundant and unneeded.
- Removes an excess `await` statement in `validatePhaseCompletion.ts`; the outside `await` statement had no affect since the inside was already being awaited.

Tests pass, exercised API locally successfully, SQ has no issues.

### DIFF
--- a/server/src/customScalarResolvers.ts
+++ b/server/src/customScalarResolvers.ts
@@ -93,9 +93,15 @@ export const DateTimeOrLocalDateResolver = new GraphQLScalarType({
   },
 });
 
+// This simple wrapper is necessary to use the logic of NonEmptyString with the TagName alias
+const TagNameResolver = new GraphQLScalarType({
+  ...NonEmptyStringResolver.toConfig(),
+  name: "TagName",
+});
+
 export const customScalarResolvers = {
   DateTime: DateTimeResolver,
   DateTimeOrLocalDate: DateTimeOrLocalDateResolver,
   NonEmptyString: NonEmptyStringResolver,
-  TagName: NonEmptyStringResolver,
+  TagName: TagNameResolver,
 };
--- a/server/src/model/applicationDate/validateInputDates.ts
+++ b/server/src/model/applicationDate/validateInputDates.ts
@@ -10,6 +10,7 @@ import {
   ApplicationDateMap,
   DateTypeValidationChecksRecord,
 } from ".";
+import { EasternTZDate } from "../../dateUtilities";
 
 export function makeEmptyValidations(): DateTypeValidationChecksRecord {
   const result = {} as DateTypeValidationChecksRecord;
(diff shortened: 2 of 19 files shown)


james-becwar-cms (approved):

server/src/model/applicationDate/validateInputDates.ts
```diff
@@ -111,35 +112,61 @@ VALIDATION_CHECKS["Federal Comment Period Start Date"]["offsetChecks"].push({
...
-          dateType,
-          check.dateTypeToCheck,
-          check.dateOffset
-        );
-      }
-    }
+    runExpectedTimestampCheck(dateType, dateValue, checks.expectedTimestamp);
+    runGreaterThanChecks(datesToValidateMap, dateType, checks.greaterThanChecks);
+    runGreaterThanOrEqualChecks(datesToValidateMap, dateType, checks.greaterThanOrEqualChecks);
+    runOffsetChecks(datesToValidateMap, dateType, checks.offsetChecks);
```
connor-parke-icf, line 170: nice, these being more atomic is a good change, especially since we're now checking dates in more places.

server/src/model/applicationPhase/validatePhaseCompletion.ts
```diff
@@ -17,9 +17,8 @@ export async function validatePhaseCompletion(
   const applicationPhaseDocumentTypes = await getApplicationPhaseDocumentTypes(applicationId, tx);
   const applicationPhaseStatuses = await getApplicationPhaseStatuses(applicationId, tx);
   // casting enforced by database schema
-  const applicationClearanceLevel = (await (
-    await getApplication(applicationId)
-  ).clearanceLevelId) as ClearanceLevel;
```
connor-parke-icf, line 22: odd double await, glad thats fixed

server/src/model/deliverable/checkDeliverableInputFunctions.ts
```diff
@@ -17,17 +17,30 @@ export function checkDemonstrationStatus(demonstration: PrismaDemonstration): st
   }
 }
 
+export function checkDeliverableHasStatus(
+  deliverable: PrismaDeliverable,
+  expectedStatuses: DeliverableStatus[]
```
connor-parke-icf, line 22: the change to expectedStatuses being an array i assume is because a given action can be applicable to more than one status

server/src/customScalarResolvers.ts
```diff
@@ -93,9 +93,15 @@ export const DateTimeOrLocalDateResolver = new GraphQLScalarType({
   },
 });
 
+// This simple wrapper is necessary to use the logic of NonEmptyString with the TagName alias
+const TagNameResolver = new GraphQLScalarType({
```
connor-parke-icf, line 97: something about this makes me feel we might not be doing something right with how we're intending to use scalars and their aliases, but if this works it works!

connor-parke-icf (approved):
LGTM, i recognize a lot of these changes from discussions we've had, so nothing really stood out to me

dustbuster (approved):

PR #1164, opened 2026-04-30
Demos 1625 side nav removal - Also includes PR 1163

dustbuster:
State User Deliverables Home page. 

https://github.com/user-attachments/assets/c39a2501-ecd6-4dc1-b646-ff051f704baa

### DIFF
--- a/client/src/layout/PrimaryLayout.tsx
+++ b/client/src/layout/PrimaryLayout.tsx
@@ -3,10 +3,13 @@ import { Footer, Header, ToastContainer, ToastProvider } from "components";
 import { SideNav } from "./SideNav";
 import { useLocation } from "react-router-dom";
 import { DialogProvider } from "components/dialog/DialogContext";
+import { getCurrentUser } from "components/user/UserContext";
+import type { PersonType } from "demos-server";
 
 const ROUTES_WITH_HIDDEN_SIDENAV = ["/demonstrations/", "/deliverables/", "/admin"];
 
-export const shouldHideSideNav = (pathname: string) =>
+export const shouldHideSideNav = (pathname: string, personType?: PersonType) =>
+  personType === "demos-state-user" ||
   ROUTES_WITH_HIDDEN_SIDENAV.some((route) => {
     if (!route.endsWith("/") && pathname === route) return true;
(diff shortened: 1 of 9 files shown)


client/src/pages/deliverables/sections/DeliverableButtons.tsx
```diff
@@ -25,14 +34,24 @@ export const DeliverableButtons = ({
...
+      ) : null}
+      {userPersonType === "demos-cms-user" ? (
+        <SecondaryButton
+          name={REQUEST_RESUBMISSION_BUTTON_NAME}
+          onClick={() => onRequestResubmission(deliverable)}
+          disabled={!canRequestResubmission(deliverable.status)}
+        >
+          Request Resubmission
+        </SecondaryButton>
+      ) : null}
```
tresdonicf, line 54: Sometimes for this sort of complexity I like to create a new component like that encapsulates the logic itself and keeps the top-level clean

client/src/layout/PrimaryLayout.tsx
```diff
@@ -3,10 +3,13 @@ import { Footer, Header, ToastContainer, ToastProvider } from "components";
...
 import { useLocation } from "react-router-dom";
 import { DialogProvider } from "components/dialog/DialogContext";
+import { getCurrentUser } from "components/user/UserContext";
+import type { PersonType } from "demos-server";
 
 const ROUTES_WITH_HIDDEN_SIDENAV = ["/demonstrations/", "/deliverables/", "/admin"];
 
-export const shouldHideSideNav = (pathname: string) =>
+export const shouldHideSideNav = (pathname: string, personType?: PersonType) =>
+  personType === "demos-state-user" ||
```
tresdonicf, line 12: FWIW, this logic does already exist in SideNav - but it's all route based atm I think. We could probably add this to that instead

tresdonicf (approved):
A couple things but overall nothing worth blocking on!

client/src/layout/PrimaryLayout.tsx
dustbuster: okay. I'll take a look. but i need to move on. So i'll make a note!

PR #1165, opened 2026-04-30
DEMOS-1625: updated empty messages - State User Deliverables update

dustbuster:
Resolves defects found in [DEMOS-1625](https://jiraent.cms.gov/browse/DEMOS-1625)

Fetching deliverable for state user via `getCurrentUser.person.roles.demonstrations.deliverables`
Which makes a really long query. but is much better than adding a resolver or totally filtering this on the F/E

- Added a page-specific state-user deliverables query using `currentUser.person.roles.demonstration.deliverables`.
- Kept CMS/admin users on the normal `deliverables` query.
- For state users, filters role assignments to `"State Point of Contact"`.
- Flattens deliverables from assigned demonstrations and dedupes by deliverable `id`.
- State users see the deliverables table directly, with no My/All tabs.
- State-user table uses the state-user column mode.
- Added empty-state copy:
  - All/state-user view: `There are no assigned Deliverables at this time`
  - My Deliverables: `You have no assigned Deliverables at this time`
- Updated `DeliverablesPage` tests for:
  - state-user query shape
  - no tabs for state users
  - state-user empty message
  - filtering out deliverables from non-State Point of Contact roles
  - distinct empty messages for My vs All Deliverables

Screenshot 2026-05-01 at 10 50 56

updated empty row messages: 
Screenshot 2026-04-30 at 15 51 52


client/src/components/table/tables/DeliverableTable.tsx
```diff
@@ -79,6 +93,46 @@ export const DELIVERABLES_PAGE_QUERY = gql`
...
+                }
+              }
+              dueDate
+            }
+          }
+        }
+      }
+    }
+  }
+`;
```
dustbuster, line 134: It's a mouthful of a query.  but the pattern here is JUST for SI is 
`getCurrentUser.person.roles.demonstrations.deliverables`

We can grab them via the user context. This way we don't need to overfetch, this table on PMDA has 12000 records. so filtering this on the F/E could be an issue. 

IF this becomes a problem or schema changes there's many other routes we can take.

client/src/components/table/tables/DeliverableTable.tsx
```diff
@@ -49,6 +49,20 @@ export type DeliverablesQueryResult = {
   deliverables: DeliverableTableRow[];
 };
 
+export type StateUserDeliverablesQueryResult = {
+  currentUser?: {
```
connor-parke-icf, line 53: huh, didnt think about this until now, but Query.currentUser returns User without an exclamation point.
```
  type Query {
    currentUser: User
  }
```
not really related to this ticket, but perhaps in the future we should update this in the schema to not return something nullable; feels wrong to say that fetching the current user can return null and that _not_ being a showstopper. 

That being said, the result of the query should never be undefined. GQL doesnt have a strong concept of undefined for query results: if you ask gql for something, it _will_ give you something back, even if that something is null.

Either way, NB but maybe consider removing the question mark here, but keeping the | null

client/src/components/table/tables/DeliverableTable.tsx
```diff
@@ -79,6 +93,46 @@ export const DELIVERABLES_PAGE_QUERY = gql`
   }
 `;
 
+export const STATE_USER_DELIVERABLES_PAGE_QUERY = gql`
+  query GetStateUserDeliverablesPage {
```
connor-parke-icf, line 97: might be a good place to break out the Deliverable's details from the path to reach the deliverable, by using a fragment. That way its obvious which part of the query has the job of filtering the results and which part of the query has the job of fetching the fields that matter.

I cant remember where i saw it, but it was an opinion piece about fragment usage that said that fragments were less about reusing common shapes and more about breaking down shapes into digestible pieces. Not sure how much i agree with that, but this seems like one of those scenarios they were talking about

client/src/pages/DeliverablesPage.tsx
```diff
@@ -11,6 +13,27 @@ import { useSessionTab } from "hooks/useSessionTab";
...
+  DeliverablesQueryResult & StateUserDeliverablesQueryResult
+>;
+
+const getStateUserDeliverables = (
+  data: DeliverablesPageQueryResult | undefined
+): DeliverablesQueryResult["deliverables"] => {
+  const deliverables =
+    data?.currentUser?.person.roles
+      .filter((roleAssignment) => roleAssignment.role === STATE_POINT_OF_CONTACT_ROLE)
+      .flatMap((roleAssignment) => roleAssignment.demonstration.deliverables) ?? [];
```
connor-parke-icf, line 30: NB: here we might want to separate out the conditions of failure from the empty list default. 

For instance, 
if theres no data, thats definitely an error state -> somethin completely broke we we should probably know about it
if theres no data.currentUser -> we dont have a current user, whatever happens past this is probably an undefined state

then we can just use data.currentUser.person.roles all we want without worry

client/src/pages/DeliverablesPage.tsx
```diff
@@ -43,9 +70,11 @@ export const DeliverablesPage: React.FC = () => {
   const rawPersonType = currentUser?.person.personType;
   const viewMode = rawPersonType as UserType;
   const isStateUser = rawPersonType === "demos-state-user";
-  const { data, loading, error } = useQuery<DeliverablesQueryResult>(DELIVERABLES_PAGE_QUERY);
+  const { data, loading, error } = useQuery<DeliverablesPageQueryResult>(
+    isStateUser ? STATE_USER_DELIVERABLES_PAGE_QUERY : DELIVERABLES_PAGE_QUERY
+  );
```
connor-parke-icf, line 75: im kinda torn on the best way to organize the state user views from the cms user views. Part of me thinks these should be separate components, or maybe a wrapper that points to the same inner component with the query, or query results, being passed in. Certainly not a blocker, just thinking in a general sense which is more ergonomic

connor-parke-icf (approved):
looks good to me! i didnt think about the fact that we might have to dedup the deliverable list via this method where we query through currentUser. I dont _think_ thats a problem, but its slightly more cumbersome than i initially realized; good catch on that! 

No blockers, just foods for thoughts/suggestions

PR #1166, opened 2026-04-30
Demos 1989 row based auth refactor user

connor-parke-icf:
This set of permissions in particular made me pause and evaluate my strategy here, as there is some redundancy built-in to the way we determine permissions and how graphql resolves object by default. In short, there are many places where we dont actually need to check for permissions as the permission filter exactly matches the filtering provided by the parent-child relationship. I.e., the permissions were built in already. However, trying to come up with and find a pattern to determine which routes and edges succesfully stayed within an authorized "realm" and which did not turned the problem into a much greater analysis (as instead of considering Nodes we had to consider Edges, which is far worse-scaling. "what bypasses do we need to allow if the user came from X node? if you can see a demonstration you can see its documents, but what about the documents of a user? the first rule applies, but the gql route to get there does not involve demonstrations"). 

This is all to say, im sticking with my original design where permissions always applied regardless of the GQL. This separates the data-fetching layer from the gql navigation layer thus no need to consider how someone fetches the data; the data is available to them or not regardless of the route. 

Anyway... heres User. This is the most complicated set of permissions so far as there are a lot of places where we need to surface the user on objects tangentially related to demonstrations: Documents, deliverables, deliverable actions (not fully implemented yet, but easy to integrate), events, etc. This led to some ideas of reorganizing which i settled on, but will be included in a followup PR. The main permission rule holds: State users can only see things that are attached to their demonstrations, or things that are attached to things attached to their demonstrations. That and the addition of "can see myself"


server/src/auth/buildAuthorizationFilter.ts
```diff
@@ -17,6 +17,34 @@ export function isStatePointOfContactOnDemonstration(
   };
 }
 
+export function isStatePointOfContactAssociatedToAmendment(
```
connor-parke-icf, line 20: this led into a more overall refactor on how im gona try and cascade these permissions a bit more cleanly, but it blew up enough to be its own PR, so that will come next.

server/src/model/deliverable/validateDeliverableInputs.test.ts
```diff
@@ -22,21 +18,20 @@ import { EasternTZDate } from "../../dateUtilities";
...
+  validateCompleteDeliverableInput,
 } from "./validateDeliverableInputs";
 
 // Mock imports
 vi.mock("../application", () => ({
   getApplication: vi.fn(),
 }));
 
-vi.mock("../user", () => ({
+vi.mock("../user/queries", () => ({
```
connor-parke-icf, line 44: this function doesnt conform to the pattern of naming, and there are some quirks to changing it to use the "select" function that would introduce more changes to the validateDeliverableInputs than i want to include in this PR (mainly null handling). So just making sure the name doesnt collide with the other getUser for now.

server/src/model/user/userData.ts
```diff
@@ -0,0 +1,96 @@
...
+  PermissionFilters,
+  ContextUser,
+} from "../../auth";
+import { selectUser, selectManyUsers } from "./queries";
+import { PrismaTransactionClient } from "../../prismaClient";
+import {
+  isStatePointOfContactAssociatedToApplication,
+} from "../../auth/buildAuthorizationFilter";
+
+const getPermissionFilters = (userId: string) =>
```
connor-parke-icf, line 14: currently not a fan of how this is done. but my attempt to change it led to a bigger scope increase i was looking for, so this will be updated in the next aforementioned PR

server/src/seeder.ts
```diff
@@ -212,6 +212,7 @@ async function seedDeliverables(actionUserId: string) {
   const context = {
     user: {
       id: actionUserId,
+      personTypeId: "demos-admin",
```
connor-parke-icf, line 215: seeding the deliverables needed a person type since we do that through the API functions, so added one here

server/src/auth/buildAuthorizationFilter.ts
dustbuster: Would it not be better to make this vertical? like `isStatePointOfContact("Amendment", $userId)`
That seems less. Specific. Maybe this what you mean by "come next"

connor-parke-icf: more or less, yeah; trying to see if i can make the permissions cascade from demonstration block-by-block rather than a straight-shot from the current table to demonstration. Users have Documents, documents have applications, applications have either extensions, amendments, or demonstrations; amendments and extensions also have demonstrations. So im looking to frame these permissions as "i have access to this thing, which has access to this thing, which has access to demonstrations" in a way that all of those paths dont need to be fully specified and the permission can rely on the permission above it. This only works if the permissions model is a Directed Acyclic Graph, which is currently causing some problems for the user/person relation because that opens the door to a cycle. Still workin through this though

dustbuster (approved):
Looks good to me.

PR #1167, opened 2026-04-30
DEMOS-1956-sort-order-for-deliverable-actions

TARehman:
Tiny fix; ensures that deliverable actions are ordered by timestamp when returned. Almost always this is the desired ordering because we get a single deliverable at a time.


connor-parke-icf (approved):
no notes, lgtm

PR #1169, opened 2026-05-01
DEMOS-2052: Remove sonarqube from pipelines

cms-jesse:
This PR removes SonarQube from the pipelines and replaces it with Snyk Code. The `snyk code` test will only fail a pipeline for a High finding (critical is not a status on snyk code)


client/change.txt
```diff
@@ -0,0 +1 @@
+this is a change. Do not merge if this file is present
```
connor-parke-icf: 👀

pipelines/Jenkinsfile.lambdas
```diff
@@ -1,184 +1,162 @@
 pipeline {
   agent {
     kubernetes {
-      yaml kubeBlock(containerNames:["node", "scanner","snyk"], opts: [
-        NODE_IMAGE: "artifactory.cloud.cms.gov/docker/node:24-alpine"
-      ])
-    }
+      yaml kubeBlock(containerNames:['node', 'scanner', 'snyk'], opts: [
```
connor-parke-icf: NB: is this change intentional? looks like identical lines just spacing differences

connor-parke-icf (approved):
tbh im not familiar with most of this. Im curious about the change.txt pattern, but i assume that means "if this is included from this point on, do not merge". I also see some of what look like just formatting changes which i assume are of no real concern. Either way, happy to see SQ go!

client/change.txt
JBToula: oh, shoot, thanks. Yeah I need to remove those

pipelines/Jenkinsfile.lambdas
cms-jesse: My auto-formatter changed locally, but the changes are fine

PR #1170, opened 2026-05-01
(2/2) DEMOS-1813: CMS / Admin users only see private comments

tresdonicf:
This is the last piece for DEMOS-1813 to create the distinction for private comments and do visibility filtering per user type. For now stubbed out some functions to get the public and private comments, trying to create a nice structure for the integration ticket that will follow when the Comments API is ready!

https://github.com/user-attachments/assets/9e0d10e4-7d35-4b07-9b9c-47bce93ad3de


client/src/pages/deliverables/sections/comment_box/getComments.ts
```diff
@@ -0,0 +1,26 @@
...
+    userFullName: "John Smith",
+    timestamp: new Date("2026-04-02T09:00:00"),
+    commentVisibility: "private",
+  },
+];
+
+export const getComments = (personType: PersonType): CommentBoxComment[] => {
+  const isStateUser = personType === "demos-state-user";
+  return [...getPublicComments(), ...(isStateUser ? [] : getPrivateComments())];
+};
```
tresdonicf, line 26: Makes sure we only dispatch the fetches for the comments as needed based on the user type (though no replacement for server-side permissions)

dustbuster (approved):
Looks fine to me!

client/src/pages/deliverables/sections/comment_box/CommentBox.tsx
```diff
@@ -29,7 +36,9 @@ const CommentBoxHeader = ({ onCollapse }: { onCollapse: () => void }) => (
 export const CommentBox = () => {
   const { currentUser } = getCurrentUser();
   const [isCollapsed, setIsCollapsed] = useState(false);
-  const [comments, setComments] = useState<CommentBoxComment[]>([]);
+  const [comments, setComments] = useState<CommentBoxComment[]>(() =>
```
connor-parke-icf, line 39: i always forget that one can put a function inside the useState initilization. Super useful when you need to map something.  

Im not super big on defaulting the result of currentUser, i feel like if there is no current user it should error, but defaulting to the most restricted user seems fine here

connor-parke-icf (commented):
This looks good to me! i like the framework and general flow, with the comment boxes being the same just with a commentVisibility flag separating them. We will probably have to restructure the way we check for the individual comments' visibility, since theres not gona be a "commentVisibility" field on the comment itself (at least, from the API) but thats a bridge we can absolutely cross later with integration.

PR #1171, opened 2026-05-01
Demos 1989 general permission map refactor

connor-parke-icf:
For this PR i wanted to reframe the relationships between the permissions of each type. Now, largely each type only has to care about any type-specific permissions (atm, just "View My User" and "View My Person"), and beyond that its relationship to a demonstration is built from its relationship with its immediate neighbors. 

While i was doing this, i went ahead and fixed up the person Permissions, which were previously defined to be unrestricted. After reviewing though, i realized that we could have actual restrictions here based on the permissions of their immediate neighbors (of which, there are relatively few; most connections go through User)


server/src/adapters/s3/LocalS3Adapter.test.ts
```diff
@@ -2,6 +2,7 @@ import { describe, it, expect, vi, beforeEach } from "vitest";
 import { PrismaTransactionClient } from "../../prismaClient";
 import { UploadDocumentInput } from "../../types";
 import { createLocalS3Adapter } from "./LocalS3Adapter";
+import { Document as PrismaDocument } from "@prisma/client";
```
connor-parke-icf, line 5: i adjusted some of the linter settings to be a little bit less permissive on our test files, which is why some of these seemingly unrelated edits exist.

server/src/auth/buildAuthorizationFilter.ts
```diff
@@ -1,50 +1,8 @@
-import { Prisma } from "@prisma/client";
 import { Permission } from "../types";
 import { ContextUser } from "./userContext";
 
 export type PermissionFilters<WhereClause> = Partial<Record<Permission, WhereClause>>;
 
-export function isStatePointOfContactOnDemonstration(
```
connor-parke-icf, line 7: moved these functions to the "Data" files theyre needed in

server/src/model/amendment/amendmentData.ts
```diff
@@ -1,12 +1,14 @@
...
+import { buildAuthorizationFilter, PermissionFilters, ContextUser } from "../../auth";
 import { selectAmendment, selectManyAmendments } from "./queries";
 import { PrismaTransactionClient } from "../../prismaClient";
+import { isAStatePointOfContactAssociatedWithDemonstration } from "../demonstration/demonstrationData";
+
+export const isAStatePointOfContactAssociatedWithAmendment = (
+  userId: string
+): Prisma.AmendmentWhereInput => ({
+  demonstration: isAStatePointOfContactAssociatedWithDemonstration(userId),
+});
```
connor-parke-icf, line 11: main idea with this refactor is to be a little more explicit in how gaining permissions on one object may grant children objects. This depends on the logical flow of permissions acting like a Directed Acyclic Graph, which is possible because "all roads lead to the demonstration".  That is to say, if an Amendment is visible because its attached to a visible Demonstration, the Amendment's documents also are.

server/src/model/application/applicationData.ts
```diff
@@ -0,0 +1,20 @@
+import { Prisma } from "@prisma/client";
+import { isAStatePointOfContactAssociatedWithAmendment } from "../amendment/amendmentData";
+import { isAStatePointOfContactAssociatedWithDemonstration } from "../demonstration/demonstrationData";
+import { isAStatePointOfContactAssociatedWithExtension } from "../extension/extensionData";
+
+export const isAStatePointOfContactAssociatedWithApplication = (
```
connor-parke-icf, line 6: though i havnt yet tackled Application in isolation, this was a useful struct that i found myself making in a few places.

server/src/model/deliverable/deliverableData.ts
```diff
@@ -0,0 +1,8 @@
+import { Prisma } from "@prisma/client";
+import { isAStatePointOfContactAssociatedWithDemonstration } from "../demonstration/demonstrationData";
+
+export const isAStatePointOfContactAssociatedWithDeliverable = (
```
connor-parke-icf, line 4: same here, i havnt yet gotten to Deliverables, but we need to be able to see the cmsOwner of Deliverables that are attached to Demonstrations, so following this pattern, the User just need to include this permission set in its query.

server/src/local-server.test.ts
```diff
@@ -52,7 +52,7 @@ describe("local-server startup", () => {
 
   it("initializes ApolloArmor with GraphQLArmorConfig and starts the server", async () => {
     // Import module after mocks are active
-    const mod = await import("./local-server.ts");
```
connor-parke-icf, line 55: no longer ignoring the test files in the linting rules meant that the .ts had to go.

server/eslint.config.js
```diff
@@ -1,32 +1,34 @@
...
 
 const eslintConfig = tseslint.config(
   {
-    ignores: [
-      '**/dev/*',
-      '**/dist/*',
-      '**/*.test.ts',
-      'tsconfig.json',
-    ]
+    ignores: ["**/dev/*", "**/dist/*", "tsconfig.json"],
```
connor-parke-icf, line 7: no longer ignoring .test.ts files

server/eslint.config.js
```diff
@@ -1,32 +1,34 @@
...
-    }
+    },
   },
+  {
+    files: ["**/*.{ts,tsx}"],
+    rules: {
+      "@typescript-eslint/no-explicit-any": "off",
+      "@typescript-eslint/no-unused-vars": "off",
+      "no-useless-escape": "off",
+    },
```
connor-parke-icf, line 30: but in those .ts files, i relaxed some of the more common rules so we didnt have a billion things to fix. we can probably taper back these relaxations in future prs, but at least now the linter can tell us if we've defined something strangely in test files.

server/src/model/amendment/amendmentData.ts
dustbuster: I learned what a DAG is today!

server/src/model/applicationTagAssignment/applicationTagAssignmentData.ts
```diff
@@ -21,25 +23,8 @@ const getPermissionFilters = (userId: string) =>
...
-          {
-            extension: {
-              demonstration: isStatePointOfContactOnDemonstration(userId),
-            },
-          },
-        ],
-      },
-    },
+    "View ApplicationTagAssignments on Assigned Demonstrations":
+      isAStatePointOfContactAssociatedWithApplicationTagAssignment(userId),
```
dustbuster, line 27: Well. that's a heckuva function name!

server/src/model/applicationTagSuggestion/applicationTagSuggestionData.ts
```diff
@@ -1,15 +1,16 @@
...
-} from "../../auth";
+import { buildAuthorizationFilter, PermissionFilters, ContextUser } from "../../auth";
 import { selectApplicationTagSuggestion, selectManyApplicationTagSuggestions } from "./queries";
 import { PrismaTransactionClient } from "../../prismaClient";
+import { isAStatePointOfContactAssociatedWithApplication } from "../application/applicationData";
+
+export const isAStatePointOfContactAssociatedWithApplicationTagSuggestion = (
+  userId: string
+): Prisma.ApplicationTagSuggestionWhereInput => ({
+  application: isAStatePointOfContactAssociatedWithApplication(userId),
```
dustbuster, line 10: isAStatePOCWithApp? idk. I think we got killed with abbreviation with PMDA so i get the spelling it out.

server/src/model/deliverable/deliverableData.ts
dustbuster: Would it not be more useful to go veritical a bit. 
`isAStatePointOfContactAssociatedWith("demonstration", $userid)`

And yes. i used the sigil. lol

dustbuster (approved):
Is there any setup?

server/src/model/amendment/amendmentData.ts
```diff
@@ -17,9 +19,8 @@ const getPermissionFilters = (userId: string) =>
         },
       },
     },
-    "View Amendments on Assigned Demonstrations": {
-      demonstration: isStatePointOfContactOnDemonstration(userId),
-    },
+    "View Amendments on Assigned Demonstrations":
+      isAStatePointOfContactAssociatedWithAmendment(userId),
```
TARehman, line 23: I'm going to be honest; some of this seems like it might be a bit more abstraction and structure than we need for an app of this complexity level. It's also a little hard to reason about because you need to think through it as a chain of Prisma where clauses, where you are feeding one where clause into another where clause. I think it might be a good idea to rename these functions to `whereIsA...` instead of just `isA`. Because the function isn't really returning whether or not something is that. It's returning a where clause that filters for that. That might seem like a bit of a nuanced distinction, but I think it would make it a lot easier to read what is going on.

Thinking about it more, I think the core issue is that it's really difficult to understand the final state of a filter because the construction of them is split into so many different constructors. I don't know how you fix that necessarily, but I think its at the root of the challenge.

server/src/model/applicationTagAssignment/applicationTagAssignmentData.ts
TARehman: Echoing Dustin's point here...I think we may not be getting much juice from the squeeze of abstracting in this fashion in some cases. All this has done is literally moved `{ application: isAStatePointOfContactAssociatedWithApplication(userId) }` to a different function call _inside this file_. So it doesn't even support like, mocking for test cases. My concern here is mostly about reasonability and the extra mental steps needed to trace the path of code.

TARehman (approved):
In principle, this all looks like its correct. My "gut instinct" here, which might be wrong, is that we've maybe stepped one step past what level of abstraction is getting us benefits. I don't think that blocks it going forward, but I do wonder if some of the "last mile" abstractions are needed (e.g. wrapper functions that just do `{ application: thing() }` so that we don't have to just write the value that way).

I think reasonability is going to be an issue, but I don't necessarily know how we should address it.

server/src/model/applicationTagAssignment/applicationTagAssignmentData.ts
connor-parke-icf: yeah, here its particularly redundant since the applicationTagAssignment is more or less a leaf node, since ApplicationTags themselves wont have any restrictions on them. But, wanted to resolve the pattern fully. I think adjusting the naming convention would smooth over some of the awkwardness

server/src/model/amendment/amendmentData.ts
connor-parke-icf: yeah, i went back and forth between how to go about structuring this, and i agree that the naming convention isnt doing us any favors at the moment. I will likely make another pass at this to adjust in a separate PR. As far as the dynamic building of the filters, i looked to see if there was a more centralized way we could build them out so all of the appropriate restrictions exist in one file, but ultimately i think that pulled too much away from the particular models. I _do_ think that level of abstraction is appropriate here, just could be conveyed a bit clearer; the abstraction really shines for models like User/Person where a bunch of different relationships could grant access (thinking ahead to deliverable action).

PR #1182, opened 2026-05-04
DEMOS-1997 readonly fields are recomputed when marking complete/incom…

dustbuster:
Making a PR for @JeffreyM-GA


dustbuster (commented):
Looks good! but i can't approve it.

JeffreyM-GA:
This fixes an issue where RO fields were not being recomputed when the demonstration was marked incomplete.

This meant that if a user completed some fields, marked the section complete, and then marked it incomplete, they could modify the fields they modified earlier, until they refreshed the page. This behavior was not intended.

After this change, once a user marks the section complete, the values are locked in and cannot be changed.

image

PR #1183, opened 2026-05-04
DEMOS-1814 add reports page

dustbuster:
Making a PR for @JeffreyM-GA


dustbuster (commented):
Good!

JeffreyM-GA:
This adds the reports page, with a table of reports available for download and buttons that don't do anything
image

PR #1185, opened 2026-05-04
DEMOS-1961-extension-request

TARehman:
This implements part of DEMOS-1961, the API for deliverable extension request management. This involves setting up some of the broader extension management items as well. The PR does the following.

- Adds new validation functions for handling relevant cases for extensions (checking that a new due date is strictly greater than the old one, checking if an active extension exists before allowing another to be created.
- Adds some of the queries necessary for manipulating the deliverable extension table.
- Add new request function to be used by the resolvers, which enters an action and then logs the extension request as needed.
- Exposes the extension reason codes as static constraints in our normal fashion.
- Small unrelated refactor which changes the imports throughout the deliverable system to come from `./auth` rather than `./auth/auth.util`.
- Modifies the DeliverableExtension Prisma model so the UUID `id` field is generated automatically. Note this is a Prisma-only change so no migration is needed.


server/src/model/deliverable/requestDeliverableExtension.ts
```diff
@@ -0,0 +1,53 @@
...
+import { prisma } from "../../prismaClient";
+import { insertDeliverableAction } from "../deliverableAction/queries";
+import { insertDeliverableExtension } from "../deliverableExtension/queries";
+
+export async function requestDeliverableExtension(
+  deliverableId: string,
+  input: RequestDeliverableExtensionInput,
+  context: GraphQLContext
+): Promise<PrismaDeliverable> {
+  validateUserPersonTypeAllowed(context, "requestDeliverableExtension", ["demos-state-user"]);
```
james-becwar-cms, line 19: Nice

james-becwar-cms (approved):

PR #1194, opened 2026-05-06
Update Synk Exception for ApolloProtobuff issue.

james-becwar-cms:
Add a year for the Synk Exception Since Apollo says the protobuff issue isn't real and they won't patch it.


TARehman (approved):
Approved based on team conversation. Apollo says this is not an issue and they don't intend to patch immediately.

PR #1196, opened 2026-05-06
DEMOS-1961-add-deliverable-extension-list-to-api

TARehman:
This PR makes the list of extensions queryable as a subkey of `deliverable`. It also makes a few other small changes. The file change count is higher than the actual changes for reasons I'll explain below.

- Updates the data model to add a partial unique index on the `deliverable_action` table so that for every extension request, we can have only one record of their request, approval, denial, and so forth. This is done outside of Prisma as it is not supported in Prisma yet. This migration also adds a single non-empty string check to bring the data model into alignment with the documentation.
- The import of the `DeepPartial` utility was moved up in the imports for about a dozen test files. For most of these, there's no meaningful change otherwise.
- Adds a proper static constraint / scalar for `DeliverableExtensionStatus` and exports it appropriately.
- Updates the seeder to seed a series of deliverable actions for one of the created deliverables.
- Creates a resolver and schema for `DeliverableExtension` and exports them appropriately, with tests for functions.
- Annotates the `s3Path` field in `document` as `NonEmptyString` which matches the DB design.
- Rewrites the `selectDeliverableExtension` function to allow for querying of things we know should be unique, but which are not unique in Prisma (due to the DB constraints being more complex than Prisma can understand). In this case, we know that there can never be more than one "Requested" extension at a time for a deliverable, meaning we might want to have functionality akin to `findUniqueOrThrow`.
- Adds a `selectDeliverableAction` function with similar design to the above, with similar justifications.

Tests passing, working locally in devcontainer, SQ passing with no issues.


nunesmlucas (approved):

server/src/model/deliverableAction/queries/selectDeliverableAction.ts
```diff
@@ -0,0 +1,29 @@
+import { DeliverableAction as PrismaDeliverableAction, Prisma } from "@prisma/client";
+import { prisma, PrismaTransactionClient } from "../../../prismaClient";
+
+export async function selectDeliverableAction(
+  where: Prisma.DeliverableActionWhereInput,
+  shouldAlwaysReturn: true,
```
connor-parke-icf, line 6: not huge into parameterize the result return behavior, but also understand the need for a function like this. I feel like it may be the caller's job to validate the result rather than this query method, but its not a big deal. Half tempted to see if we could make a Prisma extension with a query of this behavior.

server/src/model/deliverableExtension/deliverableExtensionResolvers.ts
```diff
@@ -0,0 +1,42 @@
...
+      const result = await selectDeliverableAction(
+        {
+          deliverableId: parent.deliverableId,
+          activeExtensionId: parent.id,
+          actionTypeId: "Requested Extension",
+        },
+        true
+      );
+      // Note is guaranteed to exist by database in this case
+      return result.note!;
```
connor-parke-icf, line 27: i think this is an appropriate place to use the exclamation point, since that which enforces it is in the check constraints. Wonder if itd be better to have an explicit error though instead, in case something about those check constraints ever change

server/src/model/document/documentSchema.ts
```diff
@@ -14,7 +14,7 @@ export const documentSchema = gql`
     id: ID!
     name: NonEmptyString!
     description: String
-    s3Path: String!
+    s3Path: NonEmptyString!
```
connor-parke-icf, line 17: nice catch

connor-parke-icf (approved):
LGTM! There are some spots i think may could be reworked, but itd ultimately lead to a bigger refactor than i think is worth within the scope of this ticket. Prisma restrictions are making it feel like we have to wrap prisma calls in our own custom ORM layer.  Left a few comments but nothing to slowdown for

server/src/model/document/documentSchema.ts
TARehman: It was part of me having a small meltdown trying to figure out whether all my uses of NonEmptyString throughout the rest of deliverables work. Since they all come from `deliverable_action` more or less and that has conditionally checks on that `note` field I think it works out.

server/src/model/deliverableExtension/deliverableExtensionResolvers.ts
TARehman: That feels overly defensive to me. We're assuming the database works, I think that's reasonable.

server/src/model/deliverableAction/queries/selectDeliverableAction.ts
TARehman: To me, its really just either we have one function that has an argument (and with the type signatures we get the benefit of knowing whether something can be null or not), or we'd have two functions that are _almost_ identical. 🤷🏻

PR #1198, opened 2026-05-06
DEMOS-2069 replace 'RequireAdmin' with 'RequireRole'

JeffreyM-GA:
Replaces old POC auth structure with a RequireRole component


connor-parke-icf (approved):
not much of anything to say here other than at some point we may want to look at referencing a specific Permission rather than a Role for determining what stuff is available for the frontend to render, but we're not there yet with the permissions in general. Also, using role for the route may actually be more appropriate since its less about what a user can see and more about what page should the user be brought to. so lgtm! i like the auth being parameterized rather than hardcoded as admin

JeffreyM-GA:
@connor-parke-icf Maybe either/both eventually? We'll figure it out when we get there I guess

PR #1200, opened 2026-05-06
DEMOS-1961-deny-deliverable-extension

TARehman:
This PR primarily adds the `denyDeliverableRequest` mutator and should hopefully be the last one of the DEMOS-1961 tickets.

It also adds additional records to the seeder to support denied requests, and tweaks what is returned in the extensionRequests key to include the denial reason for those that are denied.


connor-parke-icf (approved):
nothing in here stands out to me, and appears aligned with the existing patterns.  This area of the code is a little more particular to parse out and scan through, but it all seems fine to me.

PR #1204, opened 2026-05-07
DEMOS-2088: Increase cpu in node container for client pipeline

cms-jesse:
This change increases the CPU available to the Client pipeline which allows more workers for vitest


TARehman (approved):
We need to go faster. 🚀 Looks okay to me, excited to hopefully see some speedup in the pipelines.

james-becwar-cms (approved):
Thank you!

PR #1205, opened 2026-05-07
DEMOS-2078: adjusted demo delvrbles to match all/my tables (mostly)

dustbuster:
Corrected items with demo deliverable table. 

Screenshot 2026-05-07 at 15 18 35

### DIFF
--- a/client/src/pages/DemonstrationDetail/deliverables/DeliverablesTab.tsx
+++ b/client/src/pages/DemonstrationDetail/deliverables/DeliverablesTab.tsx
@@ -57,8 +57,8 @@ export const DeliverablesTab = ({
   const deliverables = isStateUser
     ? getStateUserDemonstrationDeliverables(data, parentDemonstration.id)
     : (data?.deliverables?.filter(
-        (deliverable) => deliverable.demonstration.id === parentDemonstration.id
-      ) ?? []);
+      (deliverable) => deliverable.demonstration.id === parentDemonstration.id
+    ) ?? []);
 
   return (
     <div className="flex flex-col">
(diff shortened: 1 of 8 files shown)


james-becwar-cms (approved):

client/src/pages/DemonstrationDetail/deliverables/DeliverablesTab.tsx
```diff
@@ -24,11 +49,16 @@ export const DeliverablesTab = ({
...
-  const { data, loading, error } = useQuery<DeliverablesQueryResult>(DELIVERABLES_PAGE_QUERY);
-  const deliverables = data?.deliverables.filter(
-    (deliverable) => deliverable.demonstration.id === parentDemonstration.id
-  ) ?? [];
+  const { data, loading, error } = useQuery<DeliverablesTabQueryResult>(
+    isStateUser ? STATE_USER_DELIVERABLES_PAGE_QUERY : DELIVERABLES_PAGE_QUERY
+  );
+  const deliverables = isStateUser
+    ? getStateUserDemonstrationDeliverables(data, parentDemonstration.id)
+    : (data?.deliverables?.filter(
```
connor-parke-icf, line 59: it feels like we may be repeating logic here when we check for isStateUser both to provide the query and where we determine how best to filter. Might be a good opportinuty to bundle both the fetching and the mapping to a hook for each case. 

Also, the ? after deliverables here may not be needed, since if theres a data there should also be a deliverables key on it. In general i think we should handle the case when data is not yet defined separately from how we pull out the actual fields from it, just so it easier to see how a query might be failing better.

connor-parke-icf (commented):
LGTM! i get a little nervous with defaulting data that we either know _should_ exist or could use the error it might provide as signal that something deeper is going wrong, but i dont think its really blocking.  I think we also might can consolidate some of the user type checks such that we're not repeating them within the same component, but again, not blocking

client/src/pages/DemonstrationDetail/deliverables/DeliverablesTab.tsx
dustbuster: Dang it. Already merged. Lemme just check this.

PR #1206, opened 2026-05-07
DEMOS-2015-add-deletion-path-for-deliverables

TARehman:
This PR adds the basic ability to soft-delete deliverables by adding a new deliverable status and action for deletion. The actual implementation of deletion will be done in a separate PR; this one focuses on refining some of the checks and adding appropriate records to the database.

### DIFF
--- a/server/src/model/deliverable/deliverableResolvers.ts
+++ b/server/src/model/deliverable/deliverableResolvers.ts
@@ -40,6 +40,7 @@ import { getManyDocuments } from "../document";
 import { getFormattedDeliverableActions } from "../deliverableAction";
 import { getManyDeliverableDemonstrationTypes } from "../deliverableDemonstrationType";
 import { selectManyDeliverableExtensions } from "../deliverableExtension/queries";
+import { DELETED_DELIVERABLE_STATUS } from "../../constants";
 
 export async function resolveDeliverable(
   parent: PrismaDocument,
@@ -52,7 +53,7 @@ export async function resolveDeliverable(
 
   if (parentType === Prisma.ModelName.Document) {
     if (parent.deliverableId) {
-      filter = { id: parent.deliverableId };
+      filter = { id: parent.deliverableId, NOT: { statusId: DELETED_DELIVERABLE_STATUS } };
     } else {
       filter = null;
     }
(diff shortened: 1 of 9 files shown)


james-becwar-cms (approved):

server/src/model/deliverable/deliverableResolvers.ts
```diff
@@ -92,7 +98,7 @@ export async function resolveManyDeliverables(
 }
 
 export async function queryDeliverables(): Promise<PrismaDeliverable[]> {
-  return await getManyDeliverables();
+  return await getManyDeliverables({ NOT: { statusId: DELETED_DELIVERABLE_STATUS } });
```
connor-parke-icf, line 101: theres kinda an asymmetry here with how, for singular records we filter after querying, but for many records we can keep the filter within the getMany call. Discussed, but it still feels a little off to me. i dont know how it might be done differently though, given the constraints.

server/src/model/deliverable/validateDeliverableInputs.test.ts
```diff
@@ -627,15 +634,31 @@ describe("validateDeliverableInputs", () => {
...
 
+    it("should call the check functions", async () => {
+      const testInput: Partial<PrismaDeliverable> = {
+        id: "86e6a9f2-ea55-40de-a802-507d5b2cd852",
+        statusId: "Upcoming",
+        cmsOwnerUserId: "7d8fdea5-ca19-42e5-af50-98836b6d47db",
+      };
+
+      await validateSubmitDeliverableInput(testInput as PrismaDeliverable, mockTransaction);
+      expect(checkDeliverableHasStatus).toHaveBeenCalledExactlyOnceWith(
```
connor-parke-icf, line 645: ah i see, so the check for final statuses was replaced with the check for any status; makes sense

connor-parke-icf (approved):
LGTM, i like the generalization of the status checker

server/src/model/deliverable/deliverableResolvers.ts
TARehman: Yeah, I agree that it feels a little inconsistent. The ultimate generalization is probably something like four functions:

- get a single thing and throw if it's missing
- get a single thing and don't throw if it's missing
- get multiple things and don't throw if it's missing
- get multiple things and throw if it's missing

Or the argument option where we say whether you should throw or not.

server/src/model/deliverable/validateDeliverableInputs.test.ts
TARehman: Yeah I realized after the fact that I could have just used the general status check from the start, but didn't think to go back and refactor until now.

PR #1211, opened 2026-05-08
DEMOS-1736-add-comment-queries

TARehman:
Initial work for DEMOS-1736 which is kind of intertwined with DEMOS-1958. This is adding some basic types and queries to start returning comments in the deliverable API. This is related to 1958 because of restrictions applied to deletion when something already has comments.

- Makes a small update to make the comment field non-empty string, updates the DB model and enforces this with check constraints.
- Defines a single comment type; this type is basically identical for both private and public comments, so it is living in `publicComment` for now. Note that from a GQL perspective there is no such thing as a `PrivateComment` or a `PublicComment`; they are all just `DeliverableComment`.
- Tweaks the `resolveDeliverable` function so it will work with comments. Note that this is a temporary fix; that entire structure will likely be refactored to follow @connor-parke-icf's patterns once the work is more or less done on deliverables. Note also that the incoming type is not a 1:1 match with the name of the Prisma model, since both `PublicComment` and `PrivateComment` map to `DeliverableComment`.
- Adds some comment-related queries.
- Tests throughout.

Passes SQ, passes tests, and I remembered to update the client this time to 🤞🏻 hopefully avoid causing another build error! 😁


james-becwar-cms (approved):

server/src/model/publicComment/publicCommentSchema.ts
```diff
@@ -0,0 +1,22 @@
+import { gql } from "graphql-tag";
+import { Deliverable, NonEmptyString, User } from "../../types";
+
+export const publicCommentSchema = gql`
+  type DeliverableComment {
+    id: ID!
+    deliverable: Deliverable!
```
connor-parke-icf, line 7: we might not actually need to resolve the deliverable from the comment at the moment, unless we have use cases where we want to find all of a users' comments, and the deliverables theyre on.  Might simplify things to eliminate that backlink

server/src/model/publicComment/publicCommentResolvers.test.ts
```diff
@@ -0,0 +1,72 @@
...
+    user: {
+      id: "940eb632-e307-439f-97ff-62dc42abbecd",
+    },
+  };
+
+  beforeEach(() => {
+    vi.resetAllMocks();
+  });
+
+  describe("DeliverableComment.authorUser", () => {
```
connor-parke-icf, line 47: though the actual resolver behavior is tested elsewhere, i dont think we are checking that DeliverableComment.deliverable points toward the resolveDeliverable function. Not a huge deal, but may be good to have a test that says "defers to resolveDeliverable" (assuming we want to keep that backlink)

connor-parke-icf (approved):
LGTM! no blockers, but a couple of comments. i see what you were saying about the old pattern meeting the new, with the info-param-based resolver structure of deliverable vs the arrow functions of deliverableComments. 
Im sure we will have a usecase of resolving the deliverable from the comment in the future, but for now it might reduce surface area if that "backlink" wasnt there. Its already implemented though, so not sure if it matters or not

PR #1212, opened 2026-05-08
task/cleanup-small-sq-items

TARehman:
Two very small cleanup items for SQ gripes that I was able to address while finishing up for the day. There's one left open for now that deals with the token decoder and I think it might need more brainpower than I have to spare on a Friday afternoon.

Side note: SQ told me to fix one thing with a different pattern, then promptly complained about that pattern and wanted it changed again. I just think it's funny. 🤣

### DIFF
--- a/server/src/model/applicationPhase/declareCompletenessPhaseIncomplete.ts
+++ b/server/src/model/applicationPhase/declareCompletenessPhaseIncomplete.ts
@@ -27,7 +27,7 @@ export async function declareCompletenessPhaseIncomplete(
       await updatePhaseStatus(applicationId, "Completeness", "Incomplete", tx);
       await updatePhaseStatus(applicationId, "Application Intake", "Started", tx);
 
-      const datesToDelete: DateType[] = [
+      const datesToDelete = new Set<DateType>([
         "Completeness Start Date",
         "Completeness Completion Date",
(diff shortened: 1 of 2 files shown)


server/src/model/applicationPhase/declareCompletenessPhaseIncomplete.ts
```diff
@@ -27,7 +27,7 @@ export async function declareCompletenessPhaseIncomplete(
       await updatePhaseStatus(applicationId, "Completeness", "Incomplete", tx);
       await updatePhaseStatus(applicationId, "Application Intake", "Started", tx);
 
-      const datesToDelete: DateType[] = [
+      const datesToDelete = new Set<DateType>([
```
connor-parke-icf, line 30: Interesting that this was highlighted by SQ, what was the warning message for this? i wouldnt think it would know whether or not to use an array or a set here.

connor-parke-icf (approved):

server/src/model/applicationPhase/declareCompletenessPhaseIncomplete.ts
TARehman: It was basically like "you should use Set in this case for what you're testing", I was surprised it found it as well.

PR #1216, opened 2026-05-11
DEMOS-1958-delete-deliverable

TARehman:
This PR adds support to delete deliverables via the `deleteDeliverable` API mutator.

The ability to soft-delete deliverables adds a new kind of issue we're not used to having usually, namely, the ability to run a query and get back a record that "shouldn't" exist. Normally, we just use `findUniqueOrThrow` and then the application will just throw if given an ID that does not exist. However, in the case of a soft-delete, if you default to hiding that at a different point in the query chain, it means you can get odd inconsistent scenarios (for instance, trying to delete a deliverable that has already been deleted will fail because the `statusId` is in a disallowed state, not because the deliverable doesn't exist).

Now, there's really only one place where this filtering matters, and that is in how we handle querying inside the `resolveDeliverable` function, because this needs to gracefully return `null` so that we can allow for `documents` to have a null deliverable, for instance. In other words, _most of the time_ we want to exclude the deleted deliverables, but it can occasionally be helpful to gather them. I decided to modify `getDeliverable` and `getManyDeliverables` to take an optional argument to include deleted records, so that we can, when needed, get the deleted ones, but otherwise, it functions mostly like it did before.

I also noticed that I was passing an entirely extraneous argument to the `insertDeliverableAction` function, the action time. It was always taking the system time when doing that insert, but I was still passing it along - probably a remnant of how I had been doing things before. I removed this, which allowed removing a bunch of test code that never mattered to begin with dealing with system timers.

On top of all this, I just implemented the regular checking, namely looking for public and private comments and for documents, and preventing deletion of a deliverable in those cases.

End result is a lot of small file changes, but hopefully not too much to review because most of it is just the test or call changes for small edits to a function signature.

### DIFF
--- a/server/src/model/deliverable/deliverableResolvers.ts
+++ b/server/src/model/deliverable/deliverableResolvers.ts
@@ -94,10 +94,10 @@ export async function resolveManyDeliverables(
 
   switch (parentType) {
     case Prisma.ModelName.Demonstration:
-      filter = { demonstrationId: parent.id, NOT: { statusId: DELETED_DELIVERABLE_STATUS } };
+      filter = { demonstrationId: parent.id };
       break;
     case Prisma.ModelName.User:
-      filter = { cmsOwnerUserId: parent.id, NOT: { statusId: DELETED_DELIVERABLE_STATUS } };
+      filter = { cmsOwnerUserId: parent.id };
       break;
     default:
       throw new Error(`Unsupported parent type: ${parentType}`);
@@ -108,7 +108,7 @@ export async function resolveManyDeliverables(
 }
 
 export async function queryDeliverables(): Promise<PrismaDeliverable[]> {
-  return await getManyDeliverables({ NOT: { statusId: DELETED_DELIVERABLE_STATUS } });
+  return await getManyDeliverables();
 }
 
 export function resolveDeliverableType(parent: PrismaDeliverable): DeliverableType {
--- a/server/src/model/deliverable/deliverableResolvers.ts
+++ b/server/src/model/deliverable/deliverableResolvers.ts
@@ -74,7 +74,7 @@ export async function resolveDeliverable(
   if (filter === null) {
     return null;
   }
-  const result = await getDeliverable(filter);
+  const result = await getDeliverable(filter, { includeDeleted: true });
   // We add the filter here to handle soft-deleted records
   // Do it here instead of in Prisma filter to avoid throwing when returning no records
   if (result.statusId === DELETED_DELIVERABLE_STATUS) {
--- a/server/src/model/deliverable/deliverableResolvers.ts
+++ b/server/src/model/deliverable/deliverableResolvers.ts
@@ -14,6 +14,7 @@ import {
   approveDeliverableExtension,
   completeDeliverable,
   createDeliverable,
+  deleteDeliverable,
   denyDeliverableExtension,
   getDeliverable,
   getManyDeliverables,
@@ -194,6 +195,9 @@ export const deliverableResolvers = {
       context: GraphQLContext
     ) => {
       return await denyDeliverableExtension(args.deliverableId, args.input, context);
+    },
+    deleteDeliverable: async (parent: unknown, args: { id: string }, context: GraphQLContext) => {
+      return await deleteDeliverable(args.id, context);
     },
   },
 
--- a/server/src/model/deliverable/queries/getDeliverable.ts
+++ b/server/src/model/deliverable/queries/getDeliverable.ts
@@ -1,12 +1,23 @@
 import { Prisma, Deliverable as PrismaDeliverable } from "@prisma/client";
 import { prisma, PrismaTransactionClient } from "../../../prismaClient";
+import { DELETED_DELIVERABLE_STATUS } from "../../../constants";
 
 export async function getDeliverable(
   filter: Prisma.DeliverableWhereUniqueInput,
-  tx?: PrismaTransactionClient
+  options: {
+    includeDeleted?: boolean;
+    tx?: PrismaTransactionClient;
+  } = {}
 ): Promise<PrismaDeliverable> {
-  const prismaClient = tx ?? prisma();
+  const prismaClient = options.tx ?? prisma();
+  let finalFilter = filter;
+  if (!options.includeDeleted) {
+    finalFilter = {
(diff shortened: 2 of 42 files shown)


server/src/model/deliverable/queries/getDeliverable.ts
```diff
@@ -1,12 +1,23 @@
...
 import { prisma, PrismaTransactionClient } from "../../../prismaClient";
+import { DELETED_DELIVERABLE_STATUS } from "../../../constants";
 
 export async function getDeliverable(
   filter: Prisma.DeliverableWhereUniqueInput,
-  tx?: PrismaTransactionClient
+  options: {
+    includeDeleted?: boolean;
+    tx?: PrismaTransactionClient;
+  } = {}
```
connor-parke-icf, line 10: i think itd rather see these flattened (3 independent params), cause this makes it seem like whether or not to use a transaction is the same "kind" of option as whether or not to includeDeleted, when using a transaction seems to me more meta than including delete. I suspect this is also because of annoyances with positional args, vs named args. Either way nbd; i do like the separation between these options and the main filter.

server/src/model/deliverable/queries/getDeliverable.ts
```diff
@@ -1,12 +1,23 @@
...
+  let finalFilter = filter;
+  if (!options.includeDeleted) {
+    finalFilter = {
+      ...filter,
+      NOT: { statusId: DELETED_DELIVERABLE_STATUS },
+    };
+  }
   return await prismaClient.deliverable.findUniqueOrThrow({
-    where: { ...filter },
+    where: { ...finalFilter },
```
connor-parke-icf, line 21: at some point i wana figure out why prisma is ok with this filter, since it includes things that arent part of the unique key. not sure if this is kinda an oversight with prisma or expected behavior

server/src/model/deliverable/deliverableResolvers.ts
```diff
@@ -195,6 +196,9 @@ export const deliverableResolvers = {
     ) => {
       return await denyDeliverableExtension(args.deliverableId, args.input, context);
     },
+    deleteDeliverable: async (parent: unknown, args: { id: string }, context: GraphQLContext) => {
+      return await deleteDeliverable(args.id, context);
```
connor-parke-icf, line 200: NB: generally ive been trying to pull the user off of the context here in the resolver instead of passing the entire context so that the called functions dont need to carry around anything else that might be attached to the context. Not that it matters though, the only thing on the context is the user at the moment.

server/src/model/deliverable/validateDeliverableInputs.test.ts
```diff
@@ -1422,4 +1427,129 @@ describe("validateDeliverableInputs", () => {
...
+          "One or more validation checks for deleteDeliverable have failed."
+        );
+        expect(error.extensions.code).toBe("DELETE_DELIVERABLE_VALIDATION_FAILED");
+        expect(error.extensions.originalMessages).toStrictEqual([
+          "The check for no comments has failed!",
+        ]);
+      }
+    });
+
+    it("should combine all errors into one bject", async () => {
```
connor-parke-icf: typo

server/src/model/deliverable/validateDeliverableInputs.ts
```diff
@@ -239,3 +241,18 @@ export function validateDenyDeliverableExtensionInput(
     "DENY_DELIVERABLE_EXTENSION_VALIDATION_FAILED"
   );
 }
+
+export async function validateDeleteDeliverableInput(
+  deliverable: PrismaDeliverable,
+  tx: PrismaTransactionClient
+): Promise<void> {
+  const errors: (string | undefined)[] = [];
```
connor-parke-icf, line 249: hmm, this type throws me off a little bit; if the value is undefined why keep it in the array at all?

connor-parke-icf (approved):
looks good to me. half of the files changed here are the same tiny change, so wasnt too bad. Looks fine otherwise; we're gona have to play around with the showDeleted flag to see how ergonomic it is, but until we have another soft delete elsewhere its hard to form any opinion on it. nice work

server/src/model/deliverable/queries/getDeliverable.ts
TARehman: You can't do it without being in an options argument unless you are willing to make one of them required but default to whatever, which itself makes the input patterns weird. You have to be able to do:

- transaction present, includeDeleted true
- transaction absent, includeDeleted true
- transaction present, includeDeleted false
- transaction absent, includeDeleted false

The best pattern seems to be putting them in an object, AFAICT.

server/src/model/deliverable/queries/getDeliverable.ts
TARehman: I think its because the filter still is guaranteed to return a unique item. `findUnique` itself just needs you to use a filter that it knows has a unique result, and since the base of the filter is already unique, adding additional inputs can't change that. It's not a separate type for `findUniqueOrThrow` arguments, after all, so the fact that you can scope it even _more_ restrictively isn't really an error. We might be using `findUnique` instead, after all.

server/src/model/deliverable/deliverableResolvers.ts
TARehman: That feels like a problem for future us. 😛

server/src/model/deliverable/validateDeliverableInputs.test.ts
TARehman: What do you mean? It's actually the bject, short for...nah you're right.

```suggestion
    it("should combine all errors into one object", async () => {
```

server/src/model/deliverable/validateDeliverableInputs.ts
TARehman: We don't per se. The issue is that we call all the `checkWhatever` functions which either return a string, or don't return anything, which adds `undefined` to the array. Then, later, we strip the undefined ones out of the array and if it has any messages, we throw the error. To avoid adding even an undefined item to the array I think I'd have to end up with something like `if (checkTheThing(things) !== undefined) { errors.push(checkTheThing(things) }` which feels exceedingly verbose versus just filtering the undefined items out after I do all my pushes.

PR #1221, opened 2026-05-11
DEMOS-1919 remove select column from demos table

JeffreyM-GA:
image

Removes the select column from the demonstrations table


connor-parke-icf (approved):
i have nothing to say, but feels wrong reviewing without at least some kinda comment.  nice work lol

PR #1222, opened 2026-05-12
Demos 1733 deliverable api fixes

connor-parke-icf:
This blew up more than i was looking for, but changes to the API required followup changes to the frontend.  But heres the main ideas:

BACKEND:
- uploadDocument was split into 3 separate mutators.  Because uploading to a deliverable requires extra fields, which were already in conflict with uploading to a Phase, i went ahead and separated the three uploading instances out to keep the GQL inputs strict. 
  UploadDocumentToApplication - nothing extra
  UploadDocumentToPhase - phaseId required
  UploadDocumentToDeliverable - deliverableId and isCMSUploadedFile required
- With the changing inputs, some of the helpers that relied on a single input type were changed to reflect the Prisma type instead. With this, UploadDocumentInput types were largely replaced with PrismaDocumentPendingUpload 
- It dawned on me that the upload mutators were operations on DocumentPendingUpload rather than documents, and the DocumentUploadResult type could be entirely replaced with a DocumentPendingUpload type.
- The role of the mutation resolvers no longer has to care about returning the presignedUploadUrl; that can be resolved through typical gql field resolution similarly to the presignedDownloadUrl

FRONTEND:
- because the api was split into three separate callsites, the frontend dialogs needed to be adjusted to call the correct one. To do this, i started ripping out parts of the upload process that werent about creating a DocumentPendingUpload into separate hooks, leaving the dialog handleUpload a little thinner (i think it could be yet thinner, but might be best to relegate that to the deliverable dialog ticket). Now the s3 upload, the polling, and the uploadMutation are performed in separate hooks.

Git has trouble identifying large chunks that were simply relocated, but the innards of each part are largely the same as before. Creating a new dialog for the deliverable upload _should_ slot in a bit easier with the other two, however i suspect there will be an issue with adding a new field for whether or not its a cms-attached file (which may just be defaulted to the current user's role until post-mvp. not sure).


.devcontainer/devcontainer-lock.json
connor-parke-icf, line 1: some new update to devcontainer made it start to generate this file. I ended up just putting it here, since it was on my branch. if desired i can remove it and make a separate PR for it. See slack thread for context

client/src/components/dialog/document/AddDocumentToApplicationDialog.test.tsx
```diff
@@ -350,57 +346,12 @@ describe("virus scan polling", () => {
     expect(mockLazyQueryFn).toHaveBeenCalledTimes(VIRUS_SCAN_MAX_ATTEMPTS);
   });
 
-  it("skips virus scan polling for localhost uploads", async () => {
```
connor-parke-icf, line 353: this check was moved into the new polling hook to simplify the main component

client/src/components/dialog/document/AddDocumentToApplicationDialog.tsx
connor-parke-icf, line 1: i wish GIT had a better way of detecting whether or not a file was changed and renamed vs deleted/created. 

Main changes here were to pull the parts that were universal to all kinds of documents into their own hooks, leaving this component for configuring which migration needed to be sent tot he server for kicking off the upload process. That way we could configure a component for applications, phases, and deliverables without bloating up everything with optional props.

client/src/components/dialog/document/AddDocumentToApplicationDialog.tsx
```diff
@@ -0,0 +1,105 @@
...
+  const handleDocumentUploadSucceeded = async (
+    payload: UploadDocumentToApplicationInput
+  ): Promise<void> => {
+    onDocumentUploadSucceeded?.(payload);
+    if (refetchQueries) {
+      await client.refetchQueries({ include: refetchQueries });
+    }
+  };
+
+  const handleUpload = async (
```
connor-parke-icf, line 63: everything mostly stayed the same shape, just exported. The function was the most impacted. Now it uses hooks to manage the creation of the documentPendingUpload, the uploading of the file to s3, and the polling for whether or not the file successfully uploaded.  I think there is still more that can be done here, but will defer that to the deliverable document upload, since the scope of this ticket has already blown up.

client/src/components/dialog/document/AddDocumentToPhaseDialog.tsx
```diff
@@ -0,0 +1,107 @@
...
+  onClose: () => void;
+  applicationId: string;
+  documentTypeSubset?: DocumentType[];
+  titleOverride?: string;
+  refetchQueries?: string[];
+  onDocumentUploadSucceeded?: (payload?: UploadDocumentToPhaseInput) => void;
+  phaseName: PhaseName;
+}
+
+export const AddDocumentToPhaseDialog: React.FC<AddDocumentToPhaseDialogProps> = ({
```
connor-parke-icf, line 42: wouldve loved to get these components simpler such that they really only configure different gql mutations and types, but this is as far as i was able to get. Nonetheless, deliverables should be easier to implement now that the extra s3 and polling behavior was moved out.

server/src/adapters/s3/AwsS3Adapter.ts
```diff
@@ -105,19 +104,18 @@ export function createAWSS3Adapter(): S3Adapter {
...
-    }> {
-      const documentPendingUpload = await createDocumentPendingUpload(tx, input, userId);
-      const presignedURL = await this.getPresignedUploadUrl(documentPendingUpload.id);
-      return {
-        presignedURL,
-        documentId: documentPendingUpload.id,
-      };
+      documentData: Prisma.DocumentPendingUploadCreateArgs["data"],
+      tx?: PrismaTransactionClient
+    ): Promise<PrismaDocumentPendingUpload> {
```
connor-parke-icf, line 109: when going through the api changes, i made some realizations. 
1) we dont need to couple the fetching of the upload url with the mutation entry point, and we could treat it like any other field (similarly to how the downloadUrl is fetched). That meant that the upload function in the aws adaptor's only job is only to create the document pending upload in the database - thats is.  The local adapter just needs to create the documentPendingUpload then also the Document.  But no aws connection is actually needed in this function
2) the upload operations are more of operations on the documentPendingUpload than a document. With that, i could separate out these mutations into a new DocumentPendingUpload schema, that looks similar to a regular Document. Now the upload operations simply return a DocumentPendingUpload, the the default behavior for field resolution can take effect.

server/src/model/applicationPhase/startPhase.test.ts
```diff
@@ -19,9 +19,7 @@ vi.mock("../applicationDate", () => ({
 describe("startPhaseByDocument", () => {
   const mockTransaction = "mockTransaction" as any;
   const testApplicationId = "app-123-456";
-  const mockDocument: Pick<UploadDocumentInput, "phaseName"> = {
-    phaseName: "Concept",
-  };
+  const mockPhaseName: PhaseName = "Concept";
```
connor-parke-icf, line 22: this function used to be dependent on the input type, which was now changing to be three separate types. Also, it was only using one field on the object, so went ahead and whittled down the param to be only that which we cared about here. Because the function no longer involves Documents, it was renamed accordingly.

server/src/model/applicationPhase/startPhase.test.ts
```diff
@@ -104,19 +102,4 @@ describe("startPhaseByDocument", () => {
     expect(createPhaseStartDate).toHaveBeenCalledExactlyOnceWith("Concept", mockEasternNow);
     expect(result).toBeNull();
   });
-
-  it("should return null immediately when phaseName is undefined", async () => {
```
connor-parke-icf, line 108: phaseName is now a required param, thus this test is invalid.

server/src/model/document/queries/updateDocument.ts
```diff
@@ -13,9 +13,6 @@ export async function updateDocument(
       name: input.name,
       description: input.description,
       documentTypeId: input.documentType,
-      applicationId: input.applicationId,
-      deliverableId: input.deliverableId,
-      phaseId: input.phaseName,
```
connor-parke-icf, line 18: while i was at it i trimmed down the update mutator, since we are never moving a document between applications, deliverables, or phases; only modifying the concrete fields

server/src/model/document/documentResolvers.test.ts
```diff
@@ -219,103 +213,6 @@ describe("documentResolvers", () => {
     });
   });
 
-  describe("uploadDocument", () => {
```
connor-parke-icf, line 222: these functions were moved into the documentPendingUpload resolver

server/src/model/document/documentSchema.ts
```diff
@@ -26,31 +26,13 @@ export const documentSchema = gql`
...
     documentType: DocumentType
-    applicationId: ID
-    phaseName: PhaseName
-    deliverableId: ID
-  }
-
-  type UploadDocumentResponse {
-    presignedURL: String!
-    documentId: ID!
   }
```
connor-parke-icf, line 33: UploadDocumentResponse was replaced with a fleshed out documentPendingUpload object.

server/src/model/documentPendingUpload/queries/createDocumentPendingUpload.test.ts
connor-parke-icf, line 1: this was somewhat a tough call, but documents are special: the only place we create a documentPendingUpload is through the s3Adaptors, which are already fairly outside the main patterns we are using so far. So rather than mix them, im letting the documents continue to be special and do all the handling in the s3 adapters.

server/src/model/documentPendingUpload/documentPendingUploadResolvers.ts
```diff
@@ -0,0 +1,106 @@
...
+import { getS3Adapter } from "../../adapters";
+import { getApplication, PrismaApplication } from "../application";
+import { getUser } from "../user";
+import { getDeliverable, resolveDeliverable } from "../deliverable";
+import { updateAssociatedPhase } from "./updateAssociatedPhase";
+
+export async function resolveApplication(
+  parent: PrismaDocumentPendingUpload
+): Promise<PrismaApplication> {
+  return await getApplication(parent.applicationId);
```
connor-parke-icf, line 22: until Application permissions are tackled, gona stick to the pattern we have elsewhere lest this PR blows up further.

server/src/model/documentPendingUpload/documentPendingUploadResolvers.ts
```diff
@@ -0,0 +1,106 @@
...
+}
+
+export const documentPendingUploadResolvers = {
+  Mutation: {
+    uploadDocumentToApplication: async (
```
connor-parke-icf, line 27: with this pattern, the only job of the mutator is to create the DocumentPendingUpload. It only has to go through the s3 adapter so that it can optionally create a Document alongside if on local. 

The presignedUploadUrl field now handles the fetching of the upload url, which is handled through typical resolution of subfields.

server/src/model/documentPendingUpload/documentPendingUploadSchema.ts
```diff
@@ -0,0 +1,100 @@
...
+  id: string;
+  name: NonEmptyString;
+  description?: string;
+  owner: User;
+  documentType: DocumentType;
+  application: Application;
+  phaseName?: PhaseName;
+  deliverable?: Deliverable;
+  createdAt: Date;
+  updatedAt: Date;
```
connor-parke-icf, line 74: on the fence whether we should trim down these fields or not. Theyre basically the same fields that are on Document just with a couple of changes (mainly, no s3path, no deliverableActions, and presignedUploadUrl instead of downloadUrl). i could be easily convinced to trim this down.

server/src/model/documentPendingUpload/updateAssociatedPhase.ts
```diff
@@ -0,0 +1,28 @@
+import { PhaseName } from "../../constants";
+import { getEasternNow } from "../../dateUtilities";
+import { PrismaTransactionClient } from "../../prismaClient";
+import { validateAndUpdateDates } from "../applicationDate";
+import { ApplicationDateInput } from "../applicationDate/applicationDateSchema";
+import { startPhase } from "../applicationPhase";
+
+export const updateAssociatedPhase = async (
```
connor-parke-icf, line 8: popped this out into its own file to keep the resolver more focused on routing to the appropriate places than doing much hard logic.

.devcontainer/devcontainer-lock.json
james-becwar-cms: Strange, I thought I merged this file already...

client/src/components/dialog/document/AddDocumentToApplicationDialog.tsx
james-becwar-cms: In the future, use `git mv` and it will see the change.

james-becwar-cms (approved):
Lot of files changed, but it was an easy read.  Nice job.

server/src/model/applicationPhase/startPhase.test.ts
```diff
@@ -19,9 +19,7 @@ vi.mock("../applicationDate", () => ({
 describe("startPhaseByDocument", () => {
```
TARehman, line 19: Nit: should rename this test line.

server/src/model/documentPendingUpload/updateAssociatedPhase.ts
TARehman: I might have kept this as a function vs a const but it really doesn't affect much.

server/src/model/documentPendingUpload/updateAssociatedPhase.ts
```diff
@@ -0,0 +1,28 @@
...
+  tx: PrismaTransactionClient,
+  applicationId: string,
+  phaseName: PhaseName
+) => {
+  const easternNow = getEasternNow();
+  const phaseStartDate = await startPhase(tx, applicationId, phaseName, easternNow);
+
+  const datesToUpdate: ApplicationDateInput[] = [];
+
+  if (phaseStartDate) {
```
TARehman, line 18: Nit: This probably doesn't need to be quite as verbose with a separate array of dates to update, etc. Could just be:

```typescript
if (phaseStartDate) {
  await validateAndUpdateDates(
    {
      applicationId: applicationId,
      applicationDates: [phaseStartDate]
    }
  )
}
```

server/src/model/documentPendingUpload/documentPendingUploadSchema.ts
TARehman: Not sure what the benefit or cost would be, to be honest?

server/src/model/documentPendingUpload/documentPendingUploadResolvers.ts
```diff
@@ -0,0 +1,106 @@
...
+  },
+
+  DocumentPendingUpload: {
+    owner: (parent: PrismaDocumentPendingUpload, args: unknown, context: GraphQLContext) =>
+      getUser({ id: parent.ownerUserId }, context.user),
+    documentType: (parent: PrismaDocumentPendingUpload) => parent.documentTypeId as DocumentType,
+    presignedUploadUrl: async (parent: PrismaDocumentPendingUpload) =>
+      await getS3Adapter().getPresignedUploadUrl(parent.id),
+    application: resolveApplication,
+    deliverable: resolveDeliverable,
```
TARehman, line 103: Actually realized this might be important: we need to change the resolveDeliverable function to accept this as a parent or its not going to resolve correctly I think.

PR #1223, opened 2026-05-12
DEMOS-2025-fix-cms-owner-edit-on-deliverable

TARehman:
This is a small PR which fixes a bug identified by @PatyshagulGurbangeldiyeva in the `updateDeliverable` mutator. Before, I was only changing the user ID of the CMS owner. However, if the new user was a different person type than the old user, this caused a foreign key error because you have to update both fields at the same time. This PR fixes that by properly updating both the user and the person type.

### DIFF
--- a/server/src/model/deliverable/updateDeliverable.ts
+++ b/server/src/model/deliverable/updateDeliverable.ts
@@ -1,5 +1,5 @@
 import { Deliverable as PrismaDeliverable } from "@prisma/client";
-import { UpdateDeliverableInput } from "../../types";
+import { PersonType, UpdateDeliverableInput } from "../../types";
 import { GraphQLContext } from "../../auth";
 import {
   editDeliverable,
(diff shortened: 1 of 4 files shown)


james-becwar-cms (approved):

server/src/model/deliverable/updateDeliverable.ts
```diff
@@ -27,9 +28,18 @@ export async function updateDeliverable(
...
     const editInput: EditDeliverableInput = {};
-    if (parsedInput.name) editInput.name = parsedInput.name;
-    if (parsedInput.cmsOwnerUserId) editInput.cmsOwnerUserId = parsedInput.cmsOwnerUserId;
+    if (parsedInput.name) {
+      editInput.name = parsedInput.name;
+    }
+    if (parsedInput.cmsOwnerUserId) {
+      const cmsOwner = await getUser({ id: parsedInput.cmsOwnerUserId }, tx);
+      editInput.cmsOwner = {
+        cmsOwnerUserId: cmsOwner.id,
```
connor-parke-icf, line 39: can getUser return null? if so, might want to explicitly throw an error

TARehman: It's findUniqueOrThrow - however, what we really should do is not have getUser be the name of the query and also the function above it.

PR #1224, opened 2026-05-12
DEMOS-1788 Deliverables > History tab cleanup

nunesmlucas:
# Deliverables > History tab cleanup

## Summary

History rows are immutable and ordered chronologically by default, so the row-selection checkbox column, keyword search, and Filter By dropdown are no longer needed. Removed those and tightened the layout above the table.

## Changes

All in [pages/deliverables/sections/HistoryTab.tsx](src/pages/deliverables/sections/HistoryTab.tsx):

- Dropped the checkbox column (`createSelectColumnDef`) — no bulk-action affordance for history rows.
- Stopped passing the `keywordSearch` and `columnFilter` slots to the shared `Table`.
- Added `hideSearchAndActions` to the `Table` to remove the empty `mb-[24px]` wrapper that the shared component otherwise reserves for the search/filter row.
- Removed `noResultsFoundMessage` (only fired when a search/filter narrowed results to zero; both are gone). `emptyRowsMessage="No history available."` still covers the empty case.
- Removed now-unused imports: `ColumnFilter`, `KeywordSearch`, `createSelectColumnDef`.

The chronological sort (`{ sorting: [{ id: "date", desc: true }] }`) and pagination are unchanged.

## Impact

State Files and CMS Files tabs are unaffected — they use `DeliverableFileTable`, not the shared `Table`, and `hideSearchAndActions` is a per-instance prop.

image

### DIFF
--- a/client/src/pages/deliverables/sections/FileAndHistoryTabs.tsx
+++ b/client/src/pages/deliverables/sections/FileAndHistoryTabs.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo } from "react";
+import React from "react";
 
 import { HorizontalSectionTabs, Tab } from "layout/Tabs";
 
@@ -49,10 +49,7 @@ export const FileAndHistoryTabs: React.FC<{
 }> = ({ deliverable }) => {
   const stateFiles = deliverable.stateDocuments;
   const cmsFiles = deliverable.cmsDocuments;
-  const historyRows = useMemo<DeliverableHistoryRow[]>(
-    () => deliverable.deliverableActions.map(toHistoryRow),
-    [deliverable.deliverableActions]
-  );
+  const historyRows: DeliverableHistoryRow[] = deliverable.deliverableActions.map(toHistoryRow);
 
   const { showRequestResubmissionDeliverableDialog, showCompleteReviewDeliverableDialog } =
     useDialog();
--- a/client/src/pages/deliverables/sections/FileAndHistoryTabs.tsx
+++ b/client/src/pages/deliverables/sections/FileAndHistoryTabs.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useMemo } from "react";
 
 import { HorizontalSectionTabs, Tab } from "layout/Tabs";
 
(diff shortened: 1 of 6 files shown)


client/src/pages/deliverables/sections/FileAndHistoryTabs.tsx
```diff
@@ -34,13 +44,15 @@ const TABS = {
...
-
 export const FileAndHistoryTabs: React.FC<{
   deliverable: DeliverableDetailsManagementDeliverable;
 }> = ({ deliverable }) => {
   const stateFiles = deliverable.stateDocuments;
   const cmsFiles = deliverable.cmsDocuments;
+  const historyRows = useMemo<DeliverableHistoryRow[]>(
+    () => deliverable.deliverableActions.map(toHistoryRow),
+    [deliverable.deliverableActions]
+  );
```
dustbuster: Don't think you need useMemo here. 

```
const historyRows: DeliverableHistoryRow[] =
  deliverable.deliverableActions.map(toHistoryRow);
```

I think you can just do that. ^^

dustbuster (approved):
I don't think you need useMemo, but not gonna block on that either.

client/src/pages/deliverables/sections/FileAndHistoryTabs.tsx
connor-parke-icf: agreed, if it was a larger operation that was super costly to render each time then yeah, but as it is i imagine like 10-20 or so actions on a deliverable and the map is just a map. That being said, i _think_ it will update appropriately on change of deliverable.deliverableActions, so i imagine it wont cause any issues or anything. more of a readability thing.  doesnt seem blocking to me.

client/src/pages/deliverables/sections/HistoryTab.tsx
```diff
@@ -20,10 +18,14 @@ export type DeliverableHistoryRow = {
 
 const INITIAL_TABLE_STATE = { sorting: [{ id: "date", desc: true }] };
 
+// Details may include embedded "\n" (e.g. extension request reasons); preserve those line breaks.
+function detailsCell(context: CellContext<DeliverableHistoryRow, string>) {
+  return <div className="whitespace-pre-line">{highlightCell(context)}</div>;
+}
+
```
connor-parke-icf, line 25: ive been thinking about this highlight cell thing for a while; feels like we keep needing tiny workarounds for weird interactions. Would be nice to revamp it at some point.

james-becwar-cms (approved):

connor-parke-icf (approved):
LGTM! love seeing the frontend and the backend slowly sync up on fields and datashapes and whatnot. just nitpicks and no blockers, nice!

PR #1229, opened 2026-05-13
DEMOS-1572 — CMS/Admin: Review Extension Requests

nunesmlucas:
# DEMOS-1572 — CMS/Admin: Review Extension Requests

CMS and Admin users can now Approve, Approve with New Date, or Deny extension
requests submitted by State users via a banner + modal on the deliverable
detail page. State users do not see the banner.

### Role guard

As a State user, the banner does **not** appear.

image
image
image
image

### DIFF
--- a/client/src/components/dialog/deliverable/ReviewExtensionDeliverableDialog.test.tsx
+++ b/client/src/components/dialog/deliverable/ReviewExtensionDeliverableDialog.test.tsx
@@ -1,0 +1,352 @@
+import React from "react";
+import { render, screen, waitFor, fireEvent } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { describe, it, expect, vi, beforeEach } from "vitest";
+
+import {
+  INITIAL_FORM_DATA,
+  REVIEW_EXTENSION_DETAILS_FIELD_NAME,
+  REVIEW_EXTENSION_DIALOG_TITLE,
+  REVIEW_EXTENSION_EXPIRED_NOTICE_NAME,
+  REVIEW_EXTENSION_NEW_DATE_FIELD_NAME,
+  REVIEW_EXTENSION_STATUS_FIELD_NAME,
+  REVIEW_EXTENSION_SUBMIT_BUTTON_NAME,
+  ReviewExtensionDeliverableDialog,
+  ReviewExtensionDeliverableDialogDeliverable,
+  STATE_REQUESTED_DATE_EXPIRED_MESSAGE,
+  formHasChanges,
+  formIsValid,
+  getNewDateValidationMessage,
+  isStateRequestedDateExpired,
+} from "./ReviewExtensionDeliverableDialog";
+import { DIALOG_CANCEL_BUTTON_NAME } from "components/dialog/BaseDialog";
+import { TestProvider } from "test-utils/TestProvider";
+import { DELIVERABLE_DETAILS_QUERY } from "pages/deliverables/DeliverableDetailsManagementPage";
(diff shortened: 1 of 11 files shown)


client/src/pages/deliverables/DeliverableDetailsManagementPage.tsx
```diff
@@ -193,6 +214,34 @@ export const DeliverableDetailsManagementPage: React.FC<{
...
+      extensionRequest: {
+        id,
+        reasonCode,
+        reasonDetails,
+        initialDueDateAtRequest,
+        originalDateRequested,
+      },
+    });
+  };
+
```
nunesmlucas, line 244: Replaced the sort-and-pick-first pattern with `.find()` for the open extension and an index lookup for the latest "Requested Extension" action. The deliverable can only have one open extension at a time (enforced by `canRequestExtension`), and the server returns actions ordered by `actionTimestamp` ascending, so the last entry is the most recent — no client-side sort needed.

client/src/components/dialog/deliverable/ReviewExtensionDeliverableDialog.test.tsx
```diff
@@ -0,0 +1,352 @@
...
+}));
+
+vi.mock("@apollo/client", async () => {
+  const actual = await vi.importActual("@apollo/client");
+  return {
+    ...actual,
+    useMutation: vi.fn((document: { definitions: { name?: { value: string } }[] }) => {
+      const opName = document.definitions[0]?.name?.value ?? "";
+      if (opName === "ApproveDeliverableExtension") return [mockApproveMutation];
+      if (opName === "DenyDeliverableExtension") return [mockDenyMutation];
```
connor-parke-icf, line 46: interesting, so this is introspecting the GQL to return the appropriate test function? ive never seen this before, neat! might be best to explicitly list out the in a maybe a mockedProvider maybe, but seems fine enough as it

client/src/components/dialog/deliverable/ReviewExtensionDeliverableDialog.tsx
```diff
@@ -0,0 +1,279 @@
...
+              getValidationMessage={() =>
+                attemptedSubmit && formData.newDate === ""
+                  ? "New Date is required."
+                  : newDateError
+              }
+            />
+          )}
+        </div>
+
+        {formData.decision === "Denied" && (
```
connor-parke-icf, line 263: might should make a type to hold these option strings, so we can use something like `"Denied" satisfies ExtensionReviewDecision` or something.  Not a huge deal, assuming there are only two options and its pretty apparent what they are

client/src/pages/deliverables/sections/ExtensionRequestedNotice.tsx
```diff
@@ -0,0 +1,36 @@
+import React from "react";
+import { Button } from "components/button";
+import { WarningIcon } from "components/icons";
+
+export const REVIEW_EXTENSION_REQUEST_BUTTON_NAME = "button-review-extension-request";
+export const EXTENSION_REQUESTED_NOTICE_NAME = "extension-requested-notice";
+
+export const ExtensionRequestedNotice: React.FC<{
```
connor-parke-icf, line 8: we have a notice component, but i dont believe it has a button. Maybe we could give that notice an option button parameter or something? either way, nbd; might even be better to keep it here until we have another instance of this to work with

connor-parke-icf (approved):
LGTM, couple of nitpicks but i dont think anything blocking. Nice work

PR #1230, opened 2026-05-13
DEMOS-1736-remove-getuser-query

TARehman:
Per discussion with @connor-parke-icf. This PR removes the query named `getUser` (hereafter `query getUser`. It was causing a lot of confusion between that and the higher-level `getUser` function that used `selectUser` under the hood. Instead, this replaces `query getUser` with `selectUser`. An issue was that `query getUser` would throw if nothing was returned, and that signature was used throughout the codebase. To avoid issues, I did what we've done before and added a `shouldAlwaysReturn` option on `selectUser`, which allows us to change the function signature based on whether or not we want to throw if nothing is returned.

This also tweaks `getUser` and `getManyUsers` to create a transaction of their own if one is not passed, rather than relying on the next step in the function chain doing it, on the basis that we want to generate the transaction at the highest level if one is not provided.


james-becwar-cms (approved):

server/src/model/user/queries/selectUser.ts
```diff
@@ -3,8 +3,27 @@ import { prisma, PrismaTransactionClient } from "../../../prismaClient";
...
   where: Prisma.UserWhereInput,
+  shouldAlwaysReturn: true,
+  tx?: PrismaTransactionClient
+): Promise<PrismaUser>;
+
+export async function selectUser(
+  where: Prisma.UserWhereInput,
+  shouldAlwaysReturn: false,
+  tx?: PrismaTransactionClient
+): Promise<PrismaUser | null>;
```
connor-parke-icf, line 14: I need to familiarize myself a bit more with how this works. I keep thinking of this as a runtime construct, and i cannot tell if that's true or not. either way, its a pattern we've used in a number of places already

server/src/model/user/userData.ts
```diff
@@ -51,6 +50,7 @@ export async function getUser(
   user: ContextUser,
   tx?: PrismaTransactionClient
 ): Promise<PrismaUser | null> {
+  const prismaClient = tx ?? prisma();
```
connor-parke-icf, line 53: still dont know how i feel about copying the default transaction/client handling further up in the stack. part of me wants the transaction to be all the way up at the resolver level, and part of me wants it to be pushed down at the query level.

connor-parke-icf (approved):
LGTM. Some aspect of this im a little shakey on, but nothing blocking i dont think. I feel like maybe we should have our own separate "find one or throw" method instead of overloading the current select function with a param, or maybe we should just handle the nullability in the caller method, but nothing here seems like it would cause problems or otherwise be difficult to work around.

server/src/model/user/queries/selectUser.ts
TARehman: Like most TS things its kind of a blend. So if you pass the `shouldAlwaysReturn` value as `true`, you _are_ making it throw if something comes back null. You're also telling TS that the behavior of the function will be that.

server/src/model/user/userData.ts
TARehman: Every function that takes an optional `tx` should create it if it doesn't get it. You should never have to reason about where you are in the call chain to determine what transaction you're using. You either made one or you used the one that was passed in. This is trivially obvious when you have a function that does anything more than one thing with a `tx` (e.g. if you have a function that calls two children, both of which take a `tx`, then you almost certainly want those two things to both happen in the same transaction). The best way to ensure that this behavior is consistent is to always initialize an optional transaction if it is not passed in, and then use that throughout the function. It just "doesn't matter" in this case because this function is a pure passthrough - however, as soon as we add even one other function in here you'd want to have the client defined, so the safest thing to do is have it like this.

PR #1232, opened 2026-05-13
DEMOS-2110: removed deleted status from table filters

dustbuster:
Remove deleted status from deliverable statuses in the filters. 

This will be the end statuses, but for now we can just nuke the delete. 
image

Screenshot 2026-05-13 at 16 21 45

### DIFF
--- a/client/src/components/table/columns/DeliverableColumns.tsx
+++ b/client/src/components/table/columns/DeliverableColumns.tsx
@@ -62,7 +62,9 @@ export function DeliverableColumns({
       filterConfig: {
         filterType: "select",
         // Extension Requested options not ready yet.
-        options: DELIVERABLE_STATUSES.map((status) => ({ label: status, value: status })),
+        options: DELIVERABLE_STATUSES
+          .filter((status) => status !== "Deleted")
+          .map((status) => ({ label: status, value: status })),
       },
     },
   });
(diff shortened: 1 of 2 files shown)


client/src/components/table/columns/DeliverableColumns.tsx
```diff
@@ -62,7 +62,9 @@ export function DeliverableColumns({
       filterConfig: {
         filterType: "select",
         // Extension Requested options not ready yet.
-        options: DELIVERABLE_STATUSES.map((status) => ({ label: status, value: status })),
+        options: DELIVERABLE_STATUSES
+          .filter((status) => status !== "Deleted")
```
connor-parke-icf, line 66: should probable use `"Deleted" satisfies DeliverableStatus` here, in case that status ever changes or we want to see where those statuses are used

connor-parke-icf (approved):

client/src/components/table/columns/DeliverableColumns.tsx
connor-parke-icf: nevermind, discussed. a) i think the check here is already typesafe cause DELIVERABLE_STATUSES is typed as DeliverableStatus, b) these statuses are going to be replaced with the frontend-friendly statuses soon anyway

dustbuster:
Just a note. This is gonna be the end statuses. 

image

PR #1233, opened 2026-05-13
DEMOS-1736-create-public-comment

TARehman:
This PR is adding the ability to create public comments to the API. There are also some small clean-up items throughout.

- Main cleanup: throughout a number of the deliverable tests, I was using `testUserContext` instead of `testContext`, and I realized this would be more clear if I specifically named it just `testContext`. So there's a number of files where the only change is a pure find and replace on that.
- Adds a dirt-simple implementation of a limiter so that if a state user tries to query public comments, the API will throw. This will be replaced with whatever directive-based system we use for field permissions so this is intentionally a simple and shallow implementation.
- Adds a new query that checks whether the user is a state POC on the demonstration related to the deliverable; this is used to implement the restriction that state users may only comment on comments where they are POC on the deliverable.
- Adds functions and queries to create public comments, validate that the user is permitted to do that, and insert the actual DB records, as well as tests. Also sets the default value of the ID on public comments to be UUIDs via the Prisma file (note no new migration is required, this is done client-side).


server/src/model/publicComment/queries/insertPublicComment.ts
```diff
@@ -0,0 +1,19 @@
+import { PublicComment as PrismaPublicComment } from "@prisma/client";
+import { prisma, PrismaTransactionClient } from "../../../prismaClient";
+import { NonEmptyString } from "../../../types";
+
+export type InsertPublicCommentInput = {
```
connor-parke-icf, line 5: there is a prisma type we might be able to utilize here instead of rolling our own, similar to document:
```
    async uploadDocument(
      documentData: Prisma.DocumentPendingUploadCreateArgs["data"],
      tx?: PrismaTransactionClient
    ): Promise<PrismaDocumentPendingUpload> {},
```
`Prisma.PublicCommentCreateArgs["data"]`

server/src/model/publicComment/publicComment.prisma
```diff
@@ -1,5 +1,5 @@
 model PublicComment {
-  id            String   @db.Uuid
+  id            String   @default(uuid()) @db.Uuid
```
connor-parke-icf, line 2: ahh i see what you meant when you were getting the no id error. nice catch

server/src/model/publicComment/publicCommentResolvers.ts
```diff
@@ -3,10 +3,23 @@ import {
...
+import { GraphQLContext } from "../../auth";
 import { resolveDeliverable } from "../deliverable";
 import { selectUser } from "../user/queries";
+import { createPublicComment } from ".";
 
 export const publicCommentResolvers = {
+  Mutation: {
+    createPublicComment: async (
+      parent: unknown,
+      args: { deliverableId: string; comment: NonEmptyString },
```
connor-parke-icf, line 16: might want to make a type for this, but not sure. Hard to tell cause these args are expanded in the gql but exist in an object here. 2 scalars is fine though, its apparent in the schema definition what it should be so i doubt it matters.

server/src/model/publicComment/validateUserPermittedToMakePublicComment.ts
```diff
@@ -0,0 +1,24 @@
+import { GraphQLContext } from "../../auth";
+import { PrismaTransactionClient } from "../../prismaClient";
+import { isStatePointOfContactOnDeliverableDemonstration } from "../deliverable";
+
+export async function validateUserPermittedToMakePublicComment(
+  deliverableId: string,
+  context: GraphQLContext,
+  tx: PrismaTransactionClient
+): Promise<void> {
+  if (context.user.personTypeId === "demos-state-user") {
```
connor-parke-icf, line 10: I feel like this should throw a "object not found" error rather than an explicit permission error as the user shouldnt know that this deliverable exists or not, similar to how querying for an object that the user doensnt have access to might.  Revealing that the object exists but the user doent have access to it might be a data leak

server/src/model/publicComment/publicCommentSchema.ts
```diff
@@ -10,6 +10,10 @@ export const publicCommentSchema = gql`
     createdAt: DateTime!
     updatedAt: DateTime!
   }
+
+  type Mutation {
+    createPublicComment(deliverableId: ID!, comment: NonEmptyString): DeliverableComment!
```
connor-parke-icf: we should probably make comment non-nullable in the gql

connor-parke-icf (approved):
LGTM! no blockers, but a couple of items that might can be adjusted.  Mainly the nullable input field in the gql schema

server/src/model/publicComment/queries/insertPublicComment.ts
TARehman: That includes optional arguments we probably wouldn't want to include like ID and the metadata dates, though I see the attraction of doing it.

server/src/model/publicComment/publicComment.prisma
TARehman: Actually that came after this, it was because I forgot to use a `tx` when calling something inside a transaction so things were taking place in different transactions.

server/src/model/publicComment/publicCommentResolvers.ts
TARehman: Even with a type the ID would end up being separate because of how the GQL arguments get passed in. The pattern we've usually followed is `id, inputObjectWithEverythingElse` and since everything else here is one field, I just skipped it.

server/src/model/publicComment/validateUserPermittedToMakePublicComment.ts
TARehman: This doesn't reveal anything except that the user isn't allowed to do the operation, which is what it should do, IMHO. We call this in `createPublicComment` for any input regardless of what it is, so if you're not allowed, it will crash before any information is leaked.

server/src/model/publicComment/publicCommentSchema.ts
TARehman: Good catch

TARehman:
Tried doing `npm upgrade` to resolve some of our security issues. Before and after.

```
root ➜ /workspaces/demos/server (DEMOS-1736-create-public-comment) $ npm outdated
Package                           Current    Wanted    Latest  Location                                      Depended by
@aws-sdk/client-s3               3.1041.0  3.1046.0  3.1046.0  node_modules/@aws-sdk/client-s3               server
@aws-sdk/client-secrets-manager  3.1041.0  3.1046.0  3.1046.0  node_modules/@aws-sdk/client-secrets-manager  server
@aws-sdk/client-sqs              3.1041.0  3.1046.0  3.1046.0  node_modules/@aws-sdk/client-sqs              server
@aws-sdk/s3-request-presigner    3.1041.0  3.1046.0  3.1046.0  node_modules/@aws-sdk/s3-request-presigner    server
@eslint/js                         9.39.4    9.39.4    10.0.1  node_modules/@eslint/js                       server
@faker-js/faker                     9.9.0     9.9.0    10.4.0  node_modules/@faker-js/faker                  server
@types/node                      22.19.17  22.19.19    25.7.0  node_modules/@types/node                      server
@vitest/coverage-v8                 4.1.5     4.1.6     4.1.6  node_modules/@vitest/coverage-v8              server
@vitest/ui                          4.1.5     4.1.6     4.1.6  node_modules/@vitest/ui                       server
graphql                           16.13.2   16.14.0   16.14.0  node_modules/graphql                          server
jwks-rsa                            3.2.2     3.2.2     4.0.1  node_modules/jwks-rsa                         server
tsx                                4.21.0    4.22.0    4.22.0  node_modules/tsx                              server
typescript                          5.9.3     5.9.3     6.0.3  node_modules/typescript                       server
typescript-eslint                  8.59.1    8.59.3    8.59.3  node_modules/typescript-eslint                server
vitest                              4.1.5     4.1.6     4.1.6  node_modules/vitest                           server

root ➜ /workspaces/demos/server (DEMOS-1736-create-public-comment) $ npm outdated
Package           Current    Wanted  Latest  Location                      Depended by
@eslint/js         9.39.4    9.39.4  10.0.1  node_modules/@eslint/js       server
@faker-js/faker     9.9.0     9.9.0  10.4.0  node_modules/@faker-js/faker  server
@types/node      22.19.19  22.19.19  25.7.0  node_modules/@types/node      server
jwks-rsa            3.2.2     3.2.2   4.0.1  node_modules/jwks-rsa         server
typescript          5.9.3     5.9.3   6.0.3  node_modules/typescript       server
```

PR #1236, opened 2026-05-14
DEMOS 2068 and DEMOS 1940 - Signature Level Validations

JeffreyM-GA:
I've included both the FE and BE ticket in this PR to avoid a weird state where errors could occur.

Demonstrations now only accept OA as Sig Level. This value cannot be modified and the select option renders as disabled.
Modifications accept OA and OCD only. It can be modified.

image
image

### DIFF
--- a/client/src/components/application/phases/approval-summary/applicationDetailsSection.tsx
+++ b/client/src/components/application/phases/approval-summary/applicationDetailsSection.tsx
@@ -14,6 +14,9 @@ import { formatDate } from "util/formatDate";
 const LABEL_CLASSES = tw`text-text-font font-bold text-sm tracking-wide h-[14px] flex items-center`;
 const VALUE_CLASSES = tw`text-text-font text-base leading-relaxed h-[40px] flex items-start mt-1`;
 
+export const DEMONSTRATION_SIGNATURE_LEVELS: SignatureLevel[] = ["OA"];
+export const MODIFICATION_SIGNATURE_LEVELS: SignatureLevel[] = ["OA", "OCD"];
+
 export type BaseFormData = {
   name: string;
   description?: string;
@@ -327,8 +330,8 @@ export const ApplicationDetailsSection = ({
               initialValue={sectionFormData.signatureLevel}
               allowedSignatureLevels={
                 sectionFormData.applicationType === "demonstration"
-                  ? ["OA"]
-                  : ["OA", "OCD"]
+                  ? DEMONSTRATION_SIGNATURE_LEVELS
+                  : MODIFICATION_SIGNATURE_LEVELS
               }
               onSelect={(signatureLevel) =>
                 setSectionFormData({ ...sectionFormData, signatureLevel })
--- a/client/src/components/application/phases/approval-summary/applicationDetailsSection.tsx
+++ b/client/src/components/application/phases/approval-summary/applicationDetailsSection.tsx
@@ -325,6 +325,11 @@ export const ApplicationDetailsSection = ({
             <SelectSignatureLevel
               key={`sig-${sectionFormData.signatureLevel || "empty"}`}
               initialValue={sectionFormData.signatureLevel}
+              allowedSignatureLevels={
+                sectionFormData.applicationType === "demonstration"
+                  ? ["OA"]
+                  : ["OA", "OCD"]
+              }
               onSelect={(signatureLevel) =>
                 setSectionFormData({ ...sectionFormData, signatureLevel })
               }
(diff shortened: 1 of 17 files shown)


client/src/components/application/phases/approval-summary/applicationDetailsSection.tsx
```diff
@@ -325,6 +325,11 @@ export const ApplicationDetailsSection = ({
             <SelectSignatureLevel
               key={`sig-${sectionFormData.signatureLevel || "empty"}`}
               initialValue={sectionFormData.signatureLevel}
+              allowedSignatureLevels={
+                sectionFormData.applicationType === "demonstration"
+                  ? ["OA"]
+                  : ["OA", "OCD"]
```
dustbuster: Worth using `export const SIGNATURE_LEVEL = ["OA", "OCD", "OGD"] as const;` from contants instead of hardcoded strings? i dont think these are gonna change. But weirder things...

server/src/model/amendment/amendmentResolvers.test.ts
```diff
@@ -335,6 +335,44 @@ describe("amendmentResolvers", () => {
...
+    );
+
+    it("rejects invalid signature levels during creation", async () => {
+      await expect(
+        __createAmendment(undefined, {
+          input: {
+            demonstrationId: testDemonstrationId,
+            name: testAmendmentName,
+            description: testAmendmentDescription,
+            signatureLevel: "INVALID" as SignatureLevel,
```
TARehman, line 368: Nit: I might have done something like `"OGD" satisfies SignatureLevel` so that we're using a "real" one but honestly this might be better since INVALID isn't likely to ever be a signature level.

client/src/components/dialog/demonstration/CreateDemonstrationDialog.test.tsx
```diff
@@ -125,6 +124,14 @@ describe("CreateDemonstrationDialog", () => {
     // Note: Actual selection of project officer would require more complex interaction
     // This test verifies the fields render correctly
   });
+  it("renders signature level select disabled and with the default value", () => {
```
connor-parke-icf, line 127: nice, this behavior makes sense. showing the signature level for demonstrations (to not change the form layout), but also making it clear that its gona be OA

TARehman (approved):
B/E looks good, approved based on that but I didn't review client-side code.

client/src/components/input/select/SelectSignatureLevel.tsx
```diff
@@ -11,38 +11,43 @@ const labelMap: Record<SignatureLevel, string> = {
...
-  isDisabled?: boolean;
-  isRequired?: boolean;
-}) => {
-  const [signatureLevel, setSignatureLevel] = React.useState<SignatureLevel | undefined>(
-    initialValue
-  );
+  allowedSignatureLevels = SIGNATURE_LEVEL,
+}: SelectSignatureLevelProps) => {
+  const [signatureLevel, setSignatureLevel] = React.useState<
+    SignatureLevel | undefined
```
connor-parke-icf, line 30: aside: not sure why, but i tend to prefer this way, where a nonvalue is undefined rather than the empty string. feels more... explicit maybe? nice

server/src/model/amendment/amendmentResolvers.ts
```diff
@@ -24,11 +25,17 @@ const amendmentApplicationType: ApplicationType = "Amendment";
 const conceptPhaseName: PhaseName = "Concept";
 const newApplicationStatusId: ApplicationStatus = "Pre-Submission";
 
+const VALID_AMENDMENT_SIGNATURE_LEVELS: readonly SignatureLevel[] = ["OA", "OCD"];
```
connor-parke-icf, line 28: we _might_ want to move this to constants.ts, that way both the frontend and the backend can use the same list, but not sure. This is fine either way, since we know we're gona have a follow up ticket for db enforcement, which would be as good a time to move this limiter

server/src/model/amendment/amendmentResolvers.ts
```diff
@@ -54,6 +61,10 @@ export async function __updateAmendment(
   parent: unknown,
   { id, input }: { id: string; input: UpdateAmendmentInput }
 ): Promise<PrismaAmendment> {
+  if (input.signatureLevel && !VALID_AMENDMENT_SIGNATURE_LEVELS.includes(input.signatureLevel)) {
+    throw new Error(`Invalid signature level for amendment.`);
+  }
+
```
connor-parke-icf, line 67: maybe could move this to its own helper, but certainly not a blocker; its a simple enough check, we just do it a handful of times

server/src/model/extension/extensionResolvers.test.ts
```diff
@@ -343,6 +343,44 @@ describe("extensionResolvers", () => {
       expect(transactionMocks.application.create).toHaveBeenCalledExactlyOnceWith(expectedCalls[0]);
       expect(transactionMocks.extension.create).toHaveBeenCalledExactlyOnceWith(expectedCalls[1]);
     });
+
+    it.each(["OA", "OCD"] as const)(
```
connor-parke-icf, line 347: nice use of it.each; i feel like we dont use these enough

connor-parke-icf (approved):
looks great to me, nice work. Glad we were able to dig deeper in the reqs and find that we could trim out the signature level from the demonstration inputs. A couple of comments, but nothing blocking!

PR #1237, opened 2026-05-14
DEMOS-1736-create-private-comment

TARehman:
This PR implements the `createPrivateComment` mutator which allows private comments to be added to a deliverable. A lot of the implementation is very similar to creating public comments, and the validation is actually simpler because we can just use the user type (for now anyway).

### DIFF
--- a/server/src/model/privateComment/createPrivateComment.test.ts
+++ b/server/src/model/privateComment/createPrivateComment.test.ts
@@ -6,7 +6,7 @@ import { DeepPartial } from "../../testUtilities";
 import { GraphQLContext } from "../../auth";
 
 // Functions under test
-import { createPublicComment } from "./createPublicComment";
+import { createPrivateComment } from "./createPrivateComment";
 
 // Mock imports
 vi.mock("../../prismaClient", () => ({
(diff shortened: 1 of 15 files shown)


james-becwar-cms (approved):

server/src/model/privateComment/queries/insertPrivateComment.test.ts
```diff
@@ -0,0 +1,72 @@
...
+// Mock imports
+vi.mock("../../../prismaClient", () => ({
+  prisma: vi.fn(),
+}));
+
+import { prisma } from "../../../prismaClient";
+
+describe("insertPrivateComment", () => {
+  // Test inputs
+  const testInput: InsertPrivateCommentInput = {
```
dustbuster, line 19: NIT: do we need all these comments? Kinda obvious what these things are. No a blocker. Just not needed.

server/src/model/privateComment/createPrivateComment.test.ts
```diff
@@ -0,0 +1,73 @@
+// Vitest and other helpers
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { DeepPartial } from "../../testUtilities";
+
+// Types
+import { GraphQLContext } from "../../auth";
```
dustbuster, line 6: Samsies as above. ^

server/src/model/privateComment/createPrivateComment.ts
```diff
@@ -0,0 +1,25 @@
...
+  return await prisma().$transaction(async (tx) => {
+    validateUserPermittedToMakePrivateComment(context);
+    return await insertPrivateComment(
+      {
+        deliverableId: deliverableId,
+        authorUserId: context.user.id,
+        authorPersonTypeId: context.user.personTypeId,
+        content: comment,
+      },
+      tx
```
dustbuster, line 22: If this function were a map of the US, this would be accurate. 

(TX being at the bottom get it? lol)

server/src/model/privateComment/index.ts
```diff
@@ -0,0 +1,5 @@
+// Functions
+export { createPrivateComment } from "./createPrivateComment";
+export { validateUserPermittedToMakePrivateComment } from "./validateUserPermittedToMakePrivateComment";
```
dustbuster, line 3: Tomment is what i call you.

server/src/model/privateComment/validateUserPermittedToMakePrivateComment.test.ts
```diff
@@ -0,0 +1,60 @@
...
+import { DeepPartial } from "../../testUtilities";
+
+// Types
+import { GraphQLContext } from "../../auth";
+
+// Functions under test
+import { validateUserPermittedToMakePrivateComment } from "./validateUserPermittedToMakePrivateComment";
+
+// Mock imports
+
```
dustbuster, line 12: This guy is lying to me.

dustbuster (approved):
This is fine. But please. If you don't mind the comments in the tests are just not needed. And they are misleading in one area. But this is not blocking.

TARehman:
I found it hard to organize the different kinds of imports without some kind of section headers which is what they're up there to do. The empty one is a copy/paste situation though.

PR #1249, opened 2026-05-15
DEMOS-2117: added insert to application tags from uipath value

dustbuster:
This completed the data pipeline integration that takes values from uipath_value and ports them over the tag_suggestion family that actually displays on the screens. 

Also took the opportunity to insert some separation of concerns. The Exrtaction file was getting really long so i broke it down to a few smaller files. 

But in the end, it works!

Screenshot 2026-05-15 at 12 07 35


lambdas/UIPath/runDocumentUnderstanding.ts
```diff
@@ -1,189 +1,21 @@
...
 import { getProjectIdByName } from "./getProjectId";
-import type { PoolClient } from "pg";
+import {
+  ApplicationTagSuggestionExtractError,
+  persistApplicationTagSuggestionExtracts,
+} from "./db/applicationTagSuggestionExtracts";
+import {
+  persistFinishedUiPathExtraction,
+  persistResultStatus,
+} from "./db/uiPathResults";
```
dustbuster, line 14: This file was getting really long. So i broke it down.

client/src/components/dialog/ConfirmSuggestedSparklyTagDialog.tsx
```diff
@@ -31,7 +31,7 @@ const SourcePassagePdfPreviewPlaceholder = ({ tagName }: { tagName: TagName }) =
   <div>
     <h3 className={STYLES.sourceTitle}>Source Passage</h3>
     <div className={STYLES.sourcePreview}>
-      &quot;...This is a placeholder, This feature does not exist yet, but is specifically references {tagName} in the application
+      &quot;...This specifically references {tagName} in the application
```
dustbuster, line 34: Not really related. But if we do not make it back to this feature before GO LIVE then we have something professional that is true.

lambdas/UIPath/db/applicationTagSuggestionExtracts.ts
```diff
@@ -0,0 +1,95 @@
+import type { PoolClient } from "pg";
+
+const DEMOS_SCHEMA = "demos_app";
+
+const SELECT_TAG_SUGGESTION_FIELD_LIMITS_SQL = `select id from ${DEMOS_SCHEMA}.application_tag_suggestion_extract_field_limit`;
```
connor-parke-icf, line 5: i would think pg would take care of this for us, but i get nervous with unterminated sql strings in code. could maybe add a semicolon to the end so it cant be concatenated to for injection or something. But also assuming pg does this for us in a safe way.

lambdas/UIPath/db/applicationTagSuggestionExtracts.ts
```diff
@@ -0,0 +1,95 @@
...
+const INSERT_TAG_SUGGESTION_EXTRACT_SQL = `insert into ${DEMOS_SCHEMA}.application_tag_suggestion_extract
+  (uipath_value_id, application_id, field_id, value, start_page_no, end_page_no, updated_at)
+ values
+  ($1, $2, $3, $4, $5, $6, now())
+ on conflict (uipath_value_id, application_id, value) do nothing`;
```
connor-parke-icf, line 10: on conflict
Not sure if this is meant to be an upsert or not. The unique key on application_tag_suggestion_extract is uipath_value_id, application_id, value, i _think_ uipath_value_id is a uuid that we're generating ourselves, so it should never conflict because of a unique key violation. So im thinking we shouldnt be treating this as an upsert and should be legitimately erroring if it fails

lambdas/UIPath/db/applicationTagSuggestionExtracts.ts
```diff
@@ -0,0 +1,95 @@
...
+};
+
+export class ApplicationTagSuggestionExtractError extends Error {
+  constructor(message: string, cause?: unknown) {
+    super(message, { cause });
+    this.name = "ApplicationTagSuggestionExtractError";
+  }
+}
+
+function getPageNumber(token: unknown): number | null {
```
connor-parke-icf: if i understand this right, we loop through the tokens and map to null if we cant find a page, then we look through them and throw an error if not found? could we simplify this by throwing an error in this function instead of mapping to null, if we're gona throw an error anyway

lambdas/UIPath/db/applicationTagSuggestionExtracts.ts
```diff
@@ -0,0 +1,95 @@
...
+        value.fieldId,
+        value.value,
+        startPageNo,
+        endPageNo,
+      ]);
+    }
+
+    await client.query("COMMIT");
+  } catch (error) {
+    if (transactionStarted) {
```
connor-parke-icf: this transactionStarted handling is interesting. I think that the only case in which the transaction wouldnt be started in the try catch is the case where the transaction failed to start in the first place.  im also wondering if the client would automatically rollback the transaction if client.query failed or something. I dont know enough about the specific pg behaviors here so it might be fine

lambdas/UIPath/db/applicationTagSuggestionExtracts.ts
james-becwar-cms, line 10: Why are we doing the on conflict?

lambdas/UIPath/db/applicationTagSuggestionExtracts.ts
```diff
@@ -0,0 +1,95 @@
...
+  client: PoolClient,
+  uiPathValues: PersistedUiPathValue[],
+): Promise<void> {
+  if (!uiPathValues.length) return;
+
+  let transactionStarted = false;
+
+  try {
+    await client.query("BEGIN");
+    transactionStarted = true;
```
james-becwar-cms: You don't need to do this as if the BEGIN throws, the ROLLBACK isn't going to matter and the rollback is local to your sql connection.

lambdas/UIPath/db/applicationTagSuggestionExtracts.ts
james-becwar-cms: You can remove this if, if the begin throws the rollback will too.

lambdas/UIPath/db/applicationTagSuggestionExtracts.ts
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    if (error instanceof ApplicationTagSuggestionExtractError) {
+      throw error;
+    }
+
+    throw new ApplicationTagSuggestionExtractError(
+      "Failed to create application tag suggestion extracts from UiPath values.",
+      error,
+    );
```
james-becwar-cms, line 97: This can just be throw error.  Creating the new exception changes the logged stack trace and makes debugging harder.

lambdas/UIPath/runDocumentUnderstanding.ts
```diff
@@ -426,6 +130,10 @@ export async function runDocumentUnderstanding(
 
     throw new Error("UiPath extraction did not succeed within the configured attempts.");
   } catch (error) {
+    if (error instanceof ApplicationTagSuggestionExtractError) {
+      throw error;
+    }
+
```
james-becwar-cms: How come we don't want to persist the result (the failure?)

james-becwar-cms (changes requested):

lambdas/UIPath/db/applicationTagSuggestionExtracts.ts
dustbuster: That on conflict is making the UiPath suggestion-extract insert idempotent.

dustbuster: on conflict prevents two of the same recording being inserted. with the follow conbination.
(uipath_value_id, application_id, value)

james-becwar-cms: Thank you, that makes sense.

lambdas/UIPath/db/applicationTagSuggestionExtracts.ts
dustbuster: I'm gonna just keep it simple and stupid. If there's a better way to do this, we can visit this later.

lambdas/UIPath/db/applicationTagSuggestionExtracts.ts
dustbuster: I'm trying not to bring down the house if uipath_value passes and the tag_suggestions fail. 
Regardless from the user's perspective we wouldn't fail the file upload. I don't think it really matters too much. 

But theres als some questions about the final shape of some of these data structures.

lambdas/UIPath/db/applicationTagSuggestionExtracts.ts
TARehman: So I _think_ this is okay in the short term because the parameterization here isn't a user input, it's just the schema name. That being said, we could also just hard-code the schema name in this case, I don't know that we get a _ton_ of benefit from this string substitution. @connor-parke-icf is right to be leery of injection attacks but in this case I think there's low risk.

lambdas/UIPath/db/applicationTagSuggestionExtracts.ts
dustbuster: Good call.

lambdas/UIPath/db/applicationTagSuggestionExtracts.ts
dustbuster: Outdated. Nuked this slop.

james-becwar-cms (approved):

lambdas/UIPath/db/applicationTagSuggestionExtracts.ts
TARehman: I don't think we want `ON CONFLICT` here, or generally unless there's a specific reason. These types of `ON CONFLICT` clauses can mask issues in the code. The tables in this case have the same grain so there's no reason to think you would insert to them twice.

We might consider doing this via a SELECT INTO for performance reasons. In theory, if this runs after a bunch of things insert into `uipath_value`, what is inserted into `application_tag_suggestion_extract` is just selected from `uipath_value` where the `document_id` is the one just processed and the `field_id` is what is found in the `field_limit` table.

If we do that approach you avoid making numerous small inserts to the table which might work better, and you can just do it once at the end. As an added bonus, it makes it completely impossible to cause issues with duplicates.

You would need to extract the page numbers via JSON in the query, but that isn't super crazy to do.

PR #1251, opened 2026-05-15
Demos 1741 field based auth revamp2

connor-parke-icf:
FInally merging the Field-Based auth.  These permissions define what a user is allowed to query or mutate, and as such describe the capabilities of the user. I like to think of these representing the "Questions" a user is allowed to ask, rather than the filtered "Answers" they get back. 

This works by attaching Graphql Directives to fields/queries/mutations directly in the schema which describe what permission a user needs in order to be allowed to ask about it. If they do, they are responded with an error describing the violations. 

Currently there are only 4 permissions, since the distinction is only really separated by whether you are a CMS user or not.


server/src/model/deliverable/deliverableSchema.ts
```diff
@@ -87,24 +87,31 @@ export const deliverableSchema = gql`
 
   type Mutation {
     createDeliverable(input: CreateDeliverableInput): Deliverable
+      @auth(requires: ["Perform CMS Action"])
     updateDeliverable(id: ID!, input: UpdateDeliverableInput!): Deliverable
+      @auth(requires: ["Perform CMS Action"])
     submitDeliverable(id: ID!): Deliverable
-    startDeliverableReview(id: ID!): Deliverable
+      @auth(requires: ["Perform CMS Action", "Perform State Action"])
```
connor-parke-icf, line 94: submitting a deliverable is accessible to both CMS and State Users

server/src/model/publicComment/publicCommentSchema.ts
```diff
@@ -13,6 +13,7 @@ export const publicCommentSchema = gql`
 
   type Mutation {
     createPublicComment(deliverableId: ID!, comment: NonEmptyString!): DeliverableComment!
+      @auth(requires: ["Perform CMS Action", "Perform State Action"])
```
connor-parke-icf, line 16: creating a public comment is accessible to both CMS and State Users

server/src/model/role/roleSchema.ts
```diff
@@ -2,4 +2,5 @@ import { gql } from "graphql-tag";
 
 export const roleSchema = gql`
   scalar Role
+  scalar SystemRole
```
connor-parke-icf, line 5: Not a big fan of this naming convention; id like to change Role to be DemonstrationRole, but i think that would cascade into more changes than im intending with this PR

server/src/model/deliverable/deliverableSchema.ts
dustbuster: So. I hear you like to Perform CMS Actions

dustbuster: This would include admin too?

server/src/model/publicComment/publicCommentSchema.ts
dustbuster: And what about admin? Or is admin like a blanket permission where. we don't lock them down?

server/src/model/role/roleSchema.ts
dustbuster: System is going to ROLE THE WORLD!

server/src/plugins/fieldAuthPlugin.ts
```diff
@@ -0,0 +1,121 @@
...
+  parentType: GraphQLCompositeType,
+  userPermissions: Permission[]
+): string | false => {
+  const authDirectives: Record<string, Permission[]>[] | undefined = getDirective(
+    schema,
+    fieldDef,
+    "auth"
+  );
+  if (!authDirectives || authDirectives.length === 0) {
+    return false;
```
dustbuster, line 48: SO these seem like the main drivers of the auth system. @auth maps here and runs before allowing the query to run.

server/src/seeder.ts
```diff
@@ -1066,88 +1068,6 @@ async function seedDatabase() {
   await seedApplicationTagSuggestions();
 
   await seedNotes();
-  console.log("🌱 Seeding events (with and without applicationIds)...");
```
dustbuster: I remember events....

server/src/server.ts
```diff
@@ -45,7 +46,7 @@ const server = new ApolloServer<GraphQLContext>({
   typeDefs,
   resolvers,
   ...protection,
-  plugins: [...protection.plugins, authGatePlugin, loggingPlugin],
+  plugins: [...protection.plugins, authGatePlugin, loggingPlugin, fieldAuthPlugin],
```
dustbuster, line 49: okay. so this is what loads the pluggies.

server/src/model/publicComment/publicCommentSchema.ts
dustbuster: Our conversation. 

Image

dustbuster (approved):
This looks pretty good to me. will this be data disruptive? Will every env need a reset?
I pretty sure **not** based on your migrations, but wanted to throw it out there for clarity.

server/src/model/deliverable/deliverableSchema.ts
connor-parke-icf: this does, because admin has access to both the "Perform CMS Action" and "Perform State Action" permissions, while the other two have access to only their respective permission. But on this field (and a couple others) those permissions intersect

server/src/plugins/fieldAuthPlugin.ts
connor-parke-icf: yep! thats correct. I debated for a while whether "no directive listed" meant allow access to everyone or allow no access to anyone. Went with the former to save ink and keep the schema at least somewhat readable. However, for mutations we'd like to make sure each one has an auth, but thats a manual review process. might can bake that in somewhere, but at the scale of the app currently i think its fine to just remember to check.

connor-parke-icf:
> This looks pretty good to me. will this be data disruptive? Will every env need a reset? I pretty sure **not** based on your migrations, but wanted to throw it out there for clarity.

this one shouldnt need a full reset since its purely additive to the DB and the roles themselves are already set up

server/src/model/role/roleSchema.ts
james-becwar-cms: As long as we have a dictionary of roles it should be ok.

james-becwar-cms (approved):
Very clean and easy to follow.

server/src/model/applicationDate/applicationDateSchema.ts
```diff
@@ -27,7 +27,9 @@ export const applicationDateSchema = gql`
 
   type Mutation {
     setApplicationDate(input: SetApplicationDateInput): Application
+      @auth(requires: ["Perform CMS Action"])
     setApplicationDates(input: SetApplicationDatesInput): Application
```
TARehman, line 31: Broader note: we need to go through all the schema files before launch and make sure that non-optional items are properly annotated (e.g. the return on these). Not relevant to this PR directly, just a thing I noticed.

server/src/model/migrations/20260513153422_initialize_field_permissions/migration.sql
```diff
@@ -0,0 +1,41 @@
...
+WHERE 
+  role_id = 'All Users'
+;
+
+DELETE FROM "role" 
+WHERE 
+  id = 'All Users'
+;
+
+insert into permission (id, grant_level_id) values
```
TARehman, line 11: Nit: Formatting, not worth fixing for something we'll crunch, I really need to figure out adding `sqlfluff` at some point.

server/src/constants.ts
```diff
@@ -494,6 +494,13 @@ export const PERMISSIONS = [
   "View All Users",
   "View Users on Assigned Demonstrations",
   "View My User",
+  "View All SystemRoleAssignments",
+  "View My SystemRoleAssignments",
+
```
TARehman, line 501: Nit: It seems weird to me that there's an empty line here?

server/src/model/systemRoleAssignment/index.ts
```diff
@@ -0,0 +1,14 @@
+import {
+  SystemRoleAssignment as PrismaSystemRoleAssignment,
+  Role as PrismaRole,
+  RolePermission as PrismaRolePermission,
+  Permission as PrismaPermission,
+} from "@prisma/client";
+
+export { selectManySystemRoleAssignments } from "./queries/selectManySystemRoleAssignments";
+
+export type SystemRoleAssignmentQueryResult = PrismaSystemRoleAssignment & {
```
TARehman, line 10: Nit: There's nothing wrong with it but some part of me doesn't care for the barrel file to have code in it like this. 🤪

TARehman (approved):
I'm going to get on a call with @connor-parke-icf to make sure I fully grasp the specifics of `fieldAuthPlugin` but approving because the rest looks good and I expect the same is true of that, I just want to learn.

server/src/constants.ts
connor-parke-icf: intended to show separation between the schema permissions and the resource permissions

server/src/model/systemRoleAssignment/index.ts
connor-parke-icf: yeah me too. feels wrong, but wasnt sure where else to put it. I dont like having a separate "types" file

PR #1252, opened 2026-05-15
DEMOS-2104 Server side BN workbook validation

james-becwar-cms:
- Updates the PG devcontainer to better install pg_cron (old way stopped working for me)
- Updates localstack fileprocesor and bn notebook lambads to use source maps for debuging
- Updates localstack fileprocessor and bn lambdas to use node 24
- Sets BN lambda to use 2GB of memory
- On file move to clean bucket, if file is a BN workbook, create an entry for it.
- Run BN validation using the shared library on SQS message.


lambdas/budgetNeutrality/budgetNeutralityValidation.ts
```diff
@@ -15,9 +15,15 @@ export const FINAL_BN_WORKSHEET_DOCUMENT_TYPE = "BN Workbook";
  * @throws An error if the database query fails.
  */
 export async function documentExists(pool: Pool, documentId: string): Promise<boolean> {
-  const query = `SELECT EXISTS(SELECT 1 FROM ${dbSchema}.document WHERE id = $1::UUID) AS exists;`;
+  const query = `SELECT EXISTS(SELECT 1 FROM ${dbSchema}.document WHERE id = $1::UUID);`;
   const result = await pool.query(query, [documentId]);
-  return Boolean(result.rows[0]?.exists);
+  return result.rows[0]?.exists || false;
```
connor-parke-icf, line 20: guessing the EXISTS function already returns a column called "exists" so the "AS "exists"" was redundant?

lambdas/budgetNeutrality/budgetNeutralityValidation.ts
```diff
@@ -15,9 +15,15 @@ export const FINAL_BN_WORKSHEET_DOCUMENT_TYPE = "BN Workbook";
...
+}
+
+export async function getS3Path(pool: Pool, documentId: string): Promise<string | null> {
+  const query = `SELECT s3_path FROM ${dbSchema}.document WHERE id = $1::UUID;`;
```
connor-parke-icf, line 24: i go back and forth on how much this s3_path column is needed. Most of the time, i find myself building out <bucket>/<application_id>/<document_id> anyway. Worth discussing but not i dont think important atm

lambdas/budgetNeutrality/s3.ts
```diff
@@ -0,0 +1,53 @@
+import path from "node:path";
+import os from "node:os";
+import { createWriteStream } from "node:fs";
+import { pipeline } from "node:stream/promises";
+import { Readable } from "node:stream";
+import { GetObjectCommand, S3Client, type GetObjectCommandOutput } from "@aws-sdk/client-s3";
+
+export const s3 = new S3Client({
```
connor-parke-icf, line 8: this file might be a good candidate to move to the shared library, since we have similar operations in the backend and frontend and lambda.

server/src/sql/functions.sql
```diff
@@ -275,6 +275,12 @@ BEGIN
     INTO
         v_document_type_id;
 
+    -- If the document type is "BN Workbook", create a corresponding record in budget_neutrality_workbook
+    IF v_document_type_id = 'BN Workbook' THEN
```
connor-parke-icf, line 279: oh interesting. i wouldnt have thought to do this, but makes sense. So to insert a Document that is a BN workbook, we need to have BN table and the Document Table populated with a row in the same transaction. But we dont know the BN data at the point of the processFile.  So we create essentially a placeholder in the BN table, then update the entry when the BN workbook is processed. Which.... i _assume_ will always happen after the file is processed?

lambdas/fileprocess/index.ts
connor-parke-icf, line 255: yes ok, looks like we wait the processing of the database record _then_ we queue up the lambda, so no race condition. good

server/package.json
```diff
@@ -35,6 +35,7 @@
     "build:ci": "prisma generate && tsc && esbuild dist/server.js --bundle --platform=node --outfile=build/server.cjs --minify --external:@yaacovcr/transform",
     "depcheck": "depcheck",
     "pretest": "prisma generate",
+    "predev": "prisma generate",
```
connor-parke-icf, line 38: as an aside, we should probably prune some of these. i dont think we have depcheck anymore

connor-parke-icf (approved):
looks good to me. Took a sec to figure out the flow, but it makes sense. The trick of use a placeholder in the BN table then update it later threw me off for a bit. We should probably provide some kind of fallback for pruning/handling unpopulated BN data where we expect it. or at least a big alarm bell or something to tell us we're not capturing data correctly.  Otherwise, yeah LGTM

lambdas/budgetNeutrality/budgetNeutralityValidation.ts
james-becwar-cms: This got rid of a TS error when you ran tsc.  Its kind of pointless, but thats typescript right?

lambdas/budgetNeutrality/budgetNeutralityValidation.ts
james-becwar-cms: TBH I wasn't sure how the path was constructed so I kind of just yolo'ed it and made it work in a general case.  I don't love this from a security point of view since data is being used for a path, but its data we write so I think its ok.

lambdas/budgetNeutrality/s3.ts
james-becwar-cms: I agree, but I'm not sure if the client-s3 library should be on client or if it would get tree-shaken away.

server/src/sql/functions.sql
james-becwar-cms: Yep, it tripped me up at first.  Only other thing is that a function is in an implicit transaction.  The constraint is deferred until after the transaction is finished, which is why this works.

james-becwar-cms: And yeah part of the file process recognizes that its a BN workbook and kicks of an sqs message that triggers the BN validation process.  Its split from the fileprocess lambda so we can rerun validations easily.  (just drop an sqs message on the queue with the document id)

server/package.json
james-becwar-cms: Yeah the lambda's build in a "better" way too so we could clean this up a lot and add source maps if we wanted to.  I'm not too worried about source leakage since its already on public github, but we will need to check with security.

PR #1253, opened 2026-05-15
DEMOS-2118: added deliverable resolving support for doc pending upload

connor-parke-icf:
small fix, DocumentPendingUpload wasnt able to resolve a deliverable; this fixes that

### DIFF
--- a/server/src/model/deliverable/deliverableResolvers.test.ts
+++ b/server/src/model/deliverable/deliverableResolvers.test.ts
@@ -7,6 +7,7 @@ import {
   Deliverable as PrismaDeliverable,
   Demonstration as PrismaDemonstration,
   Document as PrismaDocument,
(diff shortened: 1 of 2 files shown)


server/src/model/deliverable/deliverableResolvers.test.ts
```diff
@@ -151,9 +157,15 @@ describe("deliverableResolvers", () => {
   const testDocumentWithDeliverableParent: Partial<PrismaDocument> = {
     deliverableId: testDeliverableId,
   };
+  const testDocumentPendingUploadWithDeliverableParent: Partial<PrismaDocumentPendingUpload> = {
```
connor-parke-icf, line 160: some of this feels redundant cause they are nearly identical in structure, but felt like it made sense to be explicit about it regardless.

dustbuster (approved):

PR #1260, opened 2026-05-18
Demos-2104 Adds CLEAN_BUCKET env var to BN lambda.

james-becwar-cms:
I forgot to add the CLEAN_BUCKET ENV var to the BN lambda, so on dev it can't find the files.


deployment/lib/budgetNeutralityProcessor.ts
```diff
@@ -95,6 +99,7 @@ export class BudgetNeutralityProcessor extends Construct {
           DATABASE_SECRET_ARN: dbSecret.secretName, // pragma: allowlist secret
           LOG_LEVEL: process.env.LOG_LEVEL ?? "info",
           NODE_EXTRA_CA_CERTS: "/var/runtime/ca-cert.pem",
+          CLEAN_BUCKET: cleanReadBucket?.bucketName ?? "clean-bucket",
```
cms-jesse: ` ?? "clean-bucket"` should be removed. That's never valid

cms-jesse (changes requested):

deployment/lib/budgetNeutralityProcessor.ts
james-becwar-cms: I removed it here and from the UIPath lambda also.

cms-jesse (approved):

PR #1261, opened 2026-05-18
DEMOS-2020-remove-legacy-event-model

TARehman:
This PR removes the legacy event model from the database and from the few spots where it was implemented on the front-end. We wrote the event model originally without clearly understanding the future use cases, and as we've gotten those more clear, it's become obvious that the business logic is better met by having specific systems for each kind of event tracking, versus a general purpose system. I _think_ this removal is clean and doesn't leave anything behind, but appreciate eyes on from others as well.


server/src/model/user/userData.ts
```diff
@@ -20,13 +20,6 @@ export const isAStatePointOfContactAssociatedWithUser = (
         some: isAStatePointOfContactAssociatedWithDocument(userId),
       },
     },
-    {
-      events: {
-        some: {
-          application: isAStatePointOfContactAssociatedWithApplication(userId),
```
connor-parke-icf, line 26: thank you, this one was awful to validate since it made a bunch of connections to demonstrations without real explanations why

connor-parke-icf (approved):
LGTM. if im being honest, i dont think i can validate the renumbering of the FKs in the MMD. Maybe we need a better system for that, but idk how much it matters. As far as i can tell looking on the clientside on main, i dont see usages of event or its hooks other than the sandbox, so should be able to be safely removed.  The one connection between event and demonstration was causing me a headache to try and come p with auth rules for, so glad thats gone. There was one change in the seeder in a recent PR of mine that will likely conflict, but the issue should be inapplicable after this merges.

PR #1263, opened 2026-05-18
DEMOS-1986: refactored select query, injected permissions filter, upd…

connor-parke-icf:
This PR introduces a similar style of permissions for Deliverable as the rest; however, diverges from the existing pattern in the case of the "singleton select" actions. This comes after discussion where we solidified the desired behavior when searching for a record in a singleton field. Main change is that, when a singleton field is requested, in every identified case the user should have access to the underlying object. where this is not true is for the top-level singleton queries. In these cases, we have decided to throw an error, logging if there was a permission violation, and delegate the rest of the handling to field-level permissions (i.e. directives).


server/src/model/deliverable/deleteDeliverable.ts
```diff
@@ -17,7 +17,10 @@ export async function deleteDeliverable(
   validateUserPersonTypeAllowed(context, "deleteDeliverable", ["demos-admin", "demos-cms-user"]);
 
   return await prisma().$transaction(async (tx) => {
-    const deliverable = await getDeliverable({ id: deliverableId }, { tx: tx });
+    const deliverable = await selectDeliverable({ id: deliverableId }, tx);
+    if (!deliverable) {
```
TARehman: We've been discussing this, but given that this keeps showing up, I think we want some kind of function that does this. The main approaches we could take are the optional argument to `selectX` to say whether it should return or not, or having something like `selectXOrThrow` that wraps `selectX` and throws, which we use in these cases. I like the argument but don't mind if we do the `selectXOrThrow` option, but I do think we need to change this so that it's taking out this repeated `if(x) then throw` pattern.

server/src/model/deliverable/deliverableData.ts
```diff
@@ -1,8 +1,76 @@
...
+      },
+      tx
+    );
+
+    if (authorizedDeliverable) {
+      return authorizedDeliverable;
+    }
+  }
+
+  const deliverable = await selectDeliverable(where, tx);
```
TARehman, line 52: I *think* this is okay but to make sure I understand it, if the authFilter is null this code will always be entered; is that what we want? In other words, is it correct to say that if authFilter is null and the record exists, the user always is being blocked due to permissions? I'm trying to think through whether this logic belongs inside the conditional above or not.

server/src/model/deliverable/deliverableResolvers.ts
```diff
@@ -75,13 +76,9 @@ export async function resolveDeliverable(
   if (filter === null) {
     return null;
   }
-  const result = await getDeliverable(filter, { includeDeleted: true });
+  return await getDeliverable(filter, context.user);
   // We add the filter here to handle soft-deleted records
```
TARehman: Probably should remove this comment since the code below it got yoinked.

server/src/model/deliverable/index.ts
```diff
@@ -47,11 +47,12 @@ export {
   validateUpdateDeliverableInput,
   validateUserPersonTypeAllowed,
 } from "./validateDeliverableInputs";
+export { getDeliverable, getManyDeliverables } from "./deliverableData";
 
 // Queries
 export { editDeliverable } from "./queries/editDeliverable";
-export { getDeliverable } from "./queries/getDeliverable";
-export { getManyDeliverables } from "./queries/getManyDeliverables";
+export { selectDeliverable } from "./queries/selectDeliverable";
```
TARehman: Might want to move this stuff to a separate `queries/index.ts` if you want to match the rest of the codebase; however, that would be a lot of other small import changes throughout, so maybe makes sense as a separate PR next.

server/src/model/migrations/20260518175511_initialize_deliverable_permissions/migration.sql
```diff
@@ -0,0 +1,16 @@
...
+  ('View All Deliverables', 'System'),
+  ('View Deliverables on Assigned Demonstrations', 'System')
+;
+
+INSERT INTO "role_permission" 
+VALUES 
+  ('Admin User', 'System', 'View All Deliverables'),
+  ('Admin User', 'System', 'View Deliverables on Assigned Demonstrations'),
+  ('CMS User', 'System', 'View All Deliverables'),
+  ('CMS User', 'System', 'View Deliverables on Assigned Demonstrations'),
```
TARehman, line 14: I might be confused here but isn't it redundant to give them this permission?

server/src/model/document/documentResolvers.test.ts
```diff
@@ -66,7 +66,7 @@ vi.mock("../user", () => ({
 }));
 
 vi.mock("../deliverable", () => ({
-  getDeliverable: vi.fn(),
+  selectDeliverable: vi.fn(),
```
TARehman: If this mock was changed, why wasn't any usage of it changed? Just not relevant to the tests, but we needed a mock to be present?

server/src/model/deliverableDemonstrationType/queries/insertDeliverableDemonstrationTypes.ts
```diff
@@ -18,5 +18,9 @@ export async function insertDeliverableDemonstrationTypes(
       data: createManyPayload,
     });
   }
-  return await getDeliverable({ id: input.deliverableId }, { tx: prismaClient });
+  const deliverable = await selectDeliverable({ id: input.deliverableId }, prismaClient);
+  if (!deliverable) {
```
TARehman: Another good reason to modify the function as we've discussed is that otherwise I think you would need to go through all the tests and add a separate path for when the result is null to test the throwing behavior.

TARehman (commented):
Left some comments, and have been discussing in Slack with @connor-parke-icf who is going to make some changes based on our discussions. In general I think this all make sense, but that we should centralize the throwing when something is expected from `selectDeliverable`, and I want to make sure the logic is right for the way that the error throw for permissions errors works.

server/src/model/migrations/20260518175511_initialize_deliverable_permissions/migration.sql
connor-parke-icf: without any permissions, theyd be unable to see anything. That being said, yes granting the "See all" and the "see some" permission _is_ redundant. But because theres no concept of mutual exclusivity in the permission system, ive been opting to grant both. more just convention at this point, no idea if theres a best practice surrounding this.

server/src/model/deliverableDemonstrationType/queries/insertDeliverableDemonstrationTypes.ts
connor-parke-icf: ah... yeah. had to add a return value to the mocks in a bunch of places so eliminating that branch would surely simplify things

server/src/model/deliverable/deliverableData.ts
connor-parke-icf: hmm... i suppose theres a couple of different ways we can look at this. If they have no auth filter, we can just blanket return unauthorized. Or we can conditionally return unauthorized based on whether or not there would be data. This way it behaves the same whether or not you have _any_ auth, or just _the wrong_ auth. Not sure which is better honestly, but we should be careful about info leaking.

server/src/model/deliverable/index.ts
connor-parke-icf: yeah, that pattern of keeping discreet layers between the queries and the Data access layer kinda starts to break down when there are legitimate spots we are looking to call the queries directly from. Im not sure which i wana implement at this point, so i think im going to defer until we have a more solid set of cases

server/src/model/document/documentResolvers.test.ts
connor-parke-icf: turns out we dont need either of these mocks; i removed them. That could be a sign that we're missing a test though. I dont think we test Document.deliverable, but rather just test resolveDeliverable.

server/src/model/deliverable/deleteDeliverable.ts
connor-parke-icf: updated to use a new selectDeliverableOrThrow method, removed callsite null checking

server/src/model/deliverable/deliverableData.ts
connor-parke-icf: discussed, modified single-getter to throw if not found, providing the same error message in either case of not found or permission violation. When a permission violation occurs, the event is logged.

james-becwar-cms (approved):

server/src/model/deliverable/queries/selectDeliverableOrThrow.ts
```diff
@@ -0,0 +1,14 @@
...
+import { PrismaTransactionClient } from "../../../prismaClient";
+import { selectDeliverable } from "./selectDeliverable";
+
+export async function selectDeliverableOrThrow(
+  filter: Prisma.DeliverableWhereInput,
+  tx?: PrismaTransactionClient
+): Promise<PrismaDeliverable> {
+  const deliverable = await selectDeliverable(filter, tx);
+  if (!deliverable) {
+    throw new Error("No deliverable found matching the provided filter");
```
TARehman, line 11: Nit: If you want, you can JSON.stringify the filter here, but I understand not doing that as well.

server/src/model/deliverable/deliverableData.ts
```diff
@@ -1,8 +1,81 @@
...
+  if (deliverable) {
+    log.warn(
+      `User ${user.id} attempted to access Deliverable ${deliverable.id} without sufficient permissions.`
+    );
+  }
+
+  throw new Error("Requested Deliverable not found or User does not have Permission to view it.");
```
TARehman, line 59: Nit: Might be reasonable to include the ID requested here, since the API is invoked with it (meaning they know it anyway) - however, not a big deal either way.

server/src/model/deliverable/deliverableResolvers.ts
```diff
@@ -78,13 +80,7 @@ export async function resolveDeliverable(
...
     return null;
   }
-  const result = await getDeliverable(filter, { includeDeleted: true });
-  // We add the filter here to handle soft-deleted records
-  // Do it here instead of in Prisma filter to avoid throwing when returning no records
-  if (result.statusId === DELETED_DELIVERABLE_STATUS) {
-    return null;
-  }
-  return result;
+  return await selectDeliverableOrThrow(filter);
```
TARehman: So to avoid having this throw in the specific case of documents, we might need to refactor. There is business logic that should stop a deliverable from being soft-deleted once a document is attached, but that's at the API level, not the DB level. So we could plausibly end up with a case where document returns a deliverableId, and then this throws, because we added the deleted filter _to_ the query itself.

server/src/model/document/documentResolvers.test.ts
```diff
@@ -65,11 +65,6 @@ vi.mock("../user", () => ({
   getUser: vi.fn(),
 }));
 
-vi.mock("../deliverable", () => ({
```
TARehman, line 68: The fact that this is removed makes me wonder if we're missing some test coverage here, is the only thing tested for documentResolvers right now a document that doesn't have a deliverable?

server/src/model/deliverable/index.ts
```diff
@@ -47,11 +47,12 @@ export {
...
   validateUserPersonTypeAllowed,
 } from "./validateDeliverableInputs";
+export { getDeliverable, getManyDeliverables } from "./deliverableData";
 
 // Queries
 export { editDeliverable } from "./queries/editDeliverable";
-export { getDeliverable } from "./queries/getDeliverable";
-export { getManyDeliverables } from "./queries/getManyDeliverables";
+export { selectDeliverable, selectDeliverableOrThrow } from "./queries";
+export { selectManyDeliverables } from "./queries/selectManyDeliverables";
```
TARehman, line 55: I assume this is intermediate and will be changed further?

TARehman (commented):
Need to fix that one item resolving deliverable of a doc.

server/src/model/deliverable/deliverableResolvers.ts
```diff
@@ -58,29 +58,23 @@ export async function resolveDeliverable(
...
   switch (parentType) {
     case Prisma.ModelName.Document:
     case Prisma.ModelName.DocumentPendingUpload: {
       const doc = parent as PrismaDocument | PrismaDocumentPendingUpload;
-      filter = doc.deliverableId ? { id: doc.deliverableId } : null;
-      break;
+      if (!doc.deliverableId) {
+        return null;
+      }
+      return await selectDeliverable({ id: doc.deliverableId });
```
TARehman, line 68: NB: I am curious what would happen if we asked for records where id was null. It would probably give back null, I'd imagine.

TARehman (approved):
LGTM 🚀

PR #1265, opened 2026-05-18
DEMOS-2068 update signature level validation

JeffreyM-GA:
Fixes a few errors in the signature level validation


server/src/constants.ts
```diff
@@ -134,6 +134,10 @@ export const APPLICATION_TYPES = ["Demonstration", "Amendment", "Extension"] as
 
 export const SIGNATURE_LEVEL = ["OA", "OCD", "OGD"] as const;
 
+export const AMENDMENT_SIGNATURE_LEVELS = ["OA", "OCD"] as const;
+
+export const EXTENSION_SIGNATURE_LEVELS = ["OA", "OCD"] as const;
```
TARehman, line 139: Nit: we should probably make it so that we enforce that these are a subset of SIGNATURE_LEVEL but its not urgent.

TARehman (approved):
LGTM

server/src/model/signatureLevel/signatureLevelResolvers.ts
```diff
@@ -1,10 +1,24 @@
...
+  AmendmentSignatureLevel: generateCustomSetScalar(
+    AMENDMENT_SIGNATURE_LEVELS,
+    "AmendmentSignatureLevel",
+    "A string representing a signature level for an amendment."
+  ),
+  ExtensionSignatureLevel: generateCustomSetScalar(
+    EXTENSION_SIGNATURE_LEVELS,
+    "ExtensionSignatureLevel",
+    "A string representing a signature level for an extension."
+  ),
```
connor-parke-icf, line 23: i like! now its fully controlled by the constants.ts. not sure if "ModificationSignatureLevel" might be better for both of them, or if keeping them explicitly separate is better. dont think it matters

server/src/model/extension/extensionResolvers.ts
```diff
@@ -24,16 +23,11 @@ import { getManyApplicationTagSuggestions } from "../applicationTagSuggestion";
...
-const VALID_EXTENSION_SIGNATURE_LEVELS: readonly SignatureLevel[] = ["OA", "OCD"];
 
 export async function __createExtension(
   parent: unknown,
   { input }: { input: CreateExtensionInput }
 ): Promise<PrismaExtension> {
-  if (input.signatureLevel && !VALID_EXTENSION_SIGNATURE_LEVELS.includes(input.signatureLevel)) {
-    throw new Error(`Invalid signature level for extension.`);
-  }
-
```
connor-parke-icf, line 36: one nuance here is that the Schema has `signatureLevel: ExtensionSignatureLevel` which allows for that field to have the value of "null". This may be intended though, since technically theres nothing stopping an amendment/extension from not having a signature level at the moment (unless its approved). So with this, one _could_ unset the value i think.  Of course, that may or may not be the desired behavior so this is totally fine.

connor-parke-icf (approved):
looks good to me; i like offloading our own custom validations onto the GraphQL schema itself. Like last time, i dont think we know yet whether or not we want to allow unsetting the signature level (or letting it be null in general) yet, so this all checks out

server/src/model/extension/extensionResolvers.ts
TARehman: I think this is correct - the signature level is allowed to be unset in those places. I am wondering if we properly handle the null inputs and use them to unset in the DB; I think we do?

PR #1266, opened 2026-05-18
DEMOS-2108 — Request Extension button & Current Files toggle UI fixes

nunesmlucas:
**Request Extension button placement / role gating**
`DeliverableDetailsManagementPage.tsx`
- Grouped `Request Extension` + edit/delete into one right-aligned cluster (was scattered by `justify-between`), matching the mockup.
- Added `EDIT_DELETE_PERSON_TYPES` (admin + cms) and gated the edit/delete group on it. State users now see **only** Request Extension; admin sees both in the same section.
- Removed stale "uncomment at your leisure" comment.

**Current Files toggle rendering vertically**
`fileColumns.tsx`
- `CurrentToggle` used stock-Tailwind sizes (`h-5 w-9`), but this project's custom Tailwind v4 spacing theme remaps `--spacing-*`, so it rendered as a tall vertical sliver.
- Replaced the hand-rolled toggle with the shared `react-switch` `<Switch>` (same styling as `ContactColumns.tsx`); fixed px sizing is immune to the spacing theme.

_Note: the Figma italic text was a Figma-only issue — `SecondaryButton` is not italic, no change needed._

image

### DIFF
--- a/client/src/pages/deliverables/sections/fileColumns.tsx
+++ b/client/src/pages/deliverables/sections/fileColumns.tsx
@@ -89,6 +89,9 @@ export function makeCmsFileColumns() {
   ];
 }
 
+// Sizing uses explicit px instead of the spacing scale: this project's custom
+// Tailwind theme remaps `--spacing-*`, so `h-5`/`w-9`/`h-4` resolve to large,
+// mismatched values that render the switch as a tall vertical sliver.
 const CurrentToggle: React.FC<{ fileId: string }> = ({ fileId }) => (
   <button
     type="button"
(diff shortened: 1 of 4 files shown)


client/src/pages/deliverables/sections/fileColumns.tsx
```diff
@@ -89,15 +90,24 @@ export function makeCmsFileColumns() {
...
-  >
-    <span className="inline-block h-4 w-4 translate-x-1 transform rounded-full bg-white" />
-  </button>
+  <div className="inline-flex items-center justify-center">
+    <Switch
+      checked={false}
+      onChange={() => {}}
+      aria-label={`Toggle current file ${fileId}`}
+      onColor="#6B7280"
+      offColor="#E5E7EB"
```
connor-parke-icf, line 103: do these colors exist in the tailwind config files we have? if not, might be good to put there. Reading your comment, makes sense either way as might not be convenient to deal with tw here

connor-parke-icf (approved):
LGTM! my one comment may not be applicable since its outside of tailwind.

PR #1267, opened 2026-05-19
DEMOS-2020-data-model-update-login-actions

TARehman:
This adds in the basics of tracking user logins. I spent some time thinking about the best way to tackle this and I think the intent of the ticket is probably to log when someone does a login action via IDM / Cognito.

Naively, if you just add logging to `userContext.ts` for `findOrCreateContextUserFromClaims`, you end up with lots of records. That's because we invoke those functions whenever anything happens on the back-end; for instance, running locally, I got a record every five seconds because of the introspection queries that run in the Apollo client. That seemed like it was too granular for what was being asked.

The way around this I found was the `auth_time` claim on the JWT, which is immutable and tracks when the authentication initially occurred. Even if the token gets refreshed, the `auth_time` doesn't change, so we can use that and the user ID as a marker of something that is more or less a "session". So within `userContext.ts`, we upsert to the `user_session` table in the database, which is uniquely identified by `user_id, auth_time`. This table also tracks what I called the most recent "auth_event" (basically, when `findOrCreateContextUserFromClaims` was invoked) for that session as well as a total count for the session.

To make this work, I had to make sure that `auth_time` was passed through to the server, which requires changes in two places (since the local server is different from the regular one). Everything _appears_ to be working fine on my local machine, but I'm not sure how to check the Lambda implementation aspect of it without getting it into dev.

This also fixes one or two other small things in tests that were causing `vitest` warnings in my local.


james-becwar-cms (approved):

server/src/model/demonstration/demonstrationResolvers.test.ts
```diff
@@ -461,7 +461,7 @@ describe("demonstrationResolvers", () => {
     it("throws an error if primary project officer is not found", async () => {
       vi.mocked(getDemonstrationRoleAssignment).mockResolvedValueOnce(null);
 
-      expect(
+      await expect(
```
connor-parke-icf, line 464: interesting, was this throwing an error before? maybe a code smell?

server/src/model/userSession/queries/upsertUserSession.ts
```diff
@@ -0,0 +1,33 @@
...
+  personTypeId: UserType,
+  authTime: Date,
+  tx?: PrismaTransactionClient
+): Promise<PrismaUserSession> {
+  const prismaClient = tx ?? prisma();
+  return await prismaClient.userSession.upsert({
+    where: {
+      userId_authTime: {
+        userId: userId,
+        authTime: authTime,
```
connor-parke-icf, line 14: ahhh ok i see why this is a unique key. the idea being that when a user logs in it has a new authTime, then subsequent logins in the same session retain the same authTime, but we upsert to increment the count and set the lastAuthEventTime, while leaving the original authTime in place. we are capturing both the time of authentication and the time of the last action yeah?

server/src/local-server.ts
```diff
@@ -42,6 +42,7 @@ export function extractClaimsFromDecodedToken(decodedToken: JwtPayload): Authori
     givenName: decodedToken.given_name,
     familyName: decodedToken.family_name,
     externalUserId: externalUserId,
+    authTime: new Date(decodedToken.auth_time * 1000),
```
connor-parke-icf, line 45: this conversion here is curious. does the database store milliseconds, and they come in as seconds (since epoch)?

server/src/model/userSession/userSession.prisma
```diff
@@ -0,0 +1,14 @@
+// Flags: NO_CREATED_TS, NO_UPDATED_TS, NO_HISTORY
+// Time is captured by the action timestamp; table already for logging
+model UserSession {
+  userId            String   @map("user_id") @db.Uuid
+  personTypeId      String   @map("person_type_id")
+  authTime          DateTime @map("auth_time") @db.Timestamptz
+  lastAuthEventTime DateTime @map("last_auth_event_time") @db.Timestamptz
+  authEventCount    Int      @map("auth_event_count")
+
+  user User @relation(fields: [userId, personTypeId], references: [id, personTypeId])
```
connor-parke-icf: curious why the personTypeId is included in this table? wouldnt a relation of just userId suffice since its a primary key?

server/src/auth/userContext.ts
```diff
@@ -16,90 +17,108 @@ export type ContextUser = {
...
 export async function findOrCreateContextUserFromClaims(
   claims: AuthorizationClaims
 ): Promise<ContextUser> {
-  const existingUser = await prisma().user.findUnique({
-    where: { cognitoSubject: claims.sub },
-    include: {
-      person: {
-        include: {
-          systemRoleAssignments: true,
+  const existingContextUser = await prisma().$transaction(async (tx) => {
```
connor-parke-icf, line 75: nice touch making this a transaction; these should all succeed or fail as an atom. and good reordering of the logic, it was kinda wonky before with the two paths intermingling

lambdas/authorizer/index.test.ts
```diff
@@ -169,7 +176,7 @@ describe("authorizer", () => {
     };
 
     vi.spyOn(jwt, "verify").mockImplementationOnce((token, gk, _, f) => {
-      f(new JsonWebTokenError("there was an error"), mockDecoded);
+      f!(new JsonWebTokenError("there was an error"), mockDecoded);
```
connor-parke-icf, line 179: hmm... this feels odd, but i _imagine_ this was a fastest-path-to-resolution kinda fix. i suppose f here could be undefined? trying to tell if theres a critical change here or if its just quelling code smells

connor-parke-icf (approved):
couple of questions, but nothing blocking. looks good to me

server/src/model/demonstration/demonstrationResolvers.test.ts
TARehman: `vitest` was handling it but warned that it would stop doing that next major version, so I fixed it.

server/src/model/userSession/queries/upsertUserSession.ts
TARehman: Basically, yes. The `auth_time` on the JWT is an unchanging timestamp of when the authentication occurred, so it basically identifies a "session" for lack of a better term. So we can just use it as the key with `user_id` in the DB.

Now that I think about it, I don't actually think we need to have the person_type_id in the `user_session` table, actually. I might change that quickly.

server/src/local-server.ts
TARehman: This is a JS-ism. The auth_time is a Unix time which is seconds since 1970. JS makes dates out of milliseconds since 1970.

server/src/model/userSession/userSession.prisma
TARehman: I think I was thinking it was necessary to make sure only users were in this session table but of course that's accomplished by just using the key from users.id, so...its unnecessary and I'll do a quick change to replace it.

lambdas/authorizer/index.test.ts
TARehman: It wasn't even really an issue, tests passed fine. It's just that the callback is technically allowed to be null, as I understand it, but in our case it won't be.

PR #1269, opened 2026-05-19
DEMOS-2117: added throttling feature - OUT OF UIPATH CREDITS! - do not merge

dustbuster:
To account for cold lambda start ups. 

Solves this problem: 
Screenshot 2026-05-19 at 10 02 54

### DIFF
--- a/lambdas/UIPath/uipathClient.ts
+++ b/lambdas/UIPath/uipathClient.ts
@@ -1,4 +1,5 @@
-import axios, { AxiosRequestConfig, AxiosResponse } from "axios";
+import axios, { AxiosError, AxiosRequestConfig, AxiosResponse } from "axios";
+import { log } from "./log";
 
 export const UIPATH_BASE_URL = "https://govcloud.uipath.us";
 // This could be made dynamic if we need to support multiple tenants/environments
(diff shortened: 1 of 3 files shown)


james-becwar-cms (changes requested):
Can we use https://www.npmjs.com/package/axios-retry instead?  Also weird that we are getting a 502 from UIPath?  I think this retry logic is hiding an auth error or something.

dustbuster:
> Can we use https://www.npmjs.com/package/axios-retry instead? Also weird that we are getting a 502 from UIPath? I think this retry logic is hiding an auth error or something.

Axios has "retry-after" baked inside. We are just adding logic for the retry, Using a library makes this more opaque.  

```
function getRetryAfterDelayMs(error: AxiosError): number | undefined {
  const retryAfter = getHeaderValue(error.response?.headers, "retry-after");
```

I am considering a refactor of this lambda. I feel like its gotten way too much "ad hoc"
We are out of credits RN anyway. I need to put some thought into this. 

Was not really ready for review. I'll take your suggestions under advisement. 

Adding that library takes control away. We'd need to make the retries on uipathClient less global as this library could cause retries in the like say the extractor. Which could take even longer to complete. I don't feel this library brings a lot fo the table that we don't already have. 

We would ONLY want to retry the GET requests. Specifically only the first initial requests where we gather data from uipath.

lambdas/UIPath/uipathClient.ts
```diff
@@ -51,28 +67,140 @@ export function getProjectId(): string {
   return cachedProjectId;
 }
 
-type UipathRequestOptions = AxiosRequestConfig & {
-  params?: Record<string, string | number | undefined>;
+
```
dustbuster, line 70: Proposed replacement. 

https://www.npmjs.com/package/axios-retry

For retry logic.

dustbuster:
UIPath seems to be okay. Closing this PR

PR #1271, opened 2026-05-19
Demos 2139 deliverable document api updates

connor-parke-icf:
a couple of items bundled together:
* document now returns a isPartOfADeliverableSubmission boolean field
* delete deliverable checks if file is part of deliverable submission before allowing delete 
* UploadDocumentToDeliverable mutation now split into uploadDocumentToDeliverableCmsFiles and uploadDocumentToDeliverableStateFiles


james-becwar-cms (approved):

server/src/model/document/handleDeleteDocument.test.ts
```diff
@@ -2,11 +2,16 @@ import { describe, it, expect, vi, beforeEach } from "vitest";
 import { Document as PrismaDocument } from "@prisma/client";
 import { S3Adapter } from "../../adapters";
 import { handleDeleteDocument, deleteDocumentById } from ".";
+import { validateDocumentCanBeDeleted } from "./validateDocumentCanBeDeleted";
 
 vi.mock("./queries/deleteDocumentById", () => ({
   deleteDocumentById: vi.fn(),
 }));
 
+vi.mock("./validateDocumentCanBeDeleted", () => ({
```
TARehman, line 11: NB: It doesn't really affect anything because of the hoisting behavior of Vitest, but I personally find it more intuitive to put the `vi.mock` call above the import statement; in practice this is what Vitest does as I understand it. It hoists the mock call first and then runs the imports later. It doesn't affect things practically but I have been trying to defer to that pattern to make the file setup resemble what's happening a bit (just for mental model purposes).

server/src/model/documentPendingUpload/documentPendingUploadResolvers.ts
```diff
@@ -68,29 +69,18 @@ export const documentPendingUploadResolvers = {
...
-          deliverableId: input.deliverableId,
-          deliverableIsCmsAttachedFile: input.isCmsAttachedFile,
-          deliverableTypeId: deliverable.deliverableTypeId,
-        });
-      } catch (error) {
-        handlePrismaError(error);
-      }
-    },
+    ): Promise<PrismaDocumentPendingUpload> =>
+      handleUploadDocumentToDeliverable(input, context.user.id, true),
```
TARehman, line 77: I think you should probably await this and the below call as well. I think you can "get away" with not doing it since there's no other calls in there, but it feels safer if we await here.

server/src/model/documentPendingUpload/handleUploadDocumentToDeliverable.ts
```diff
@@ -0,0 +1,29 @@
+import { getS3Adapter } from "../../adapters";
+import { checkOptionalNotNullFields } from "../../errors/checkOptionalNotNullFields";
+import { handlePrismaError } from "../../errors/handlePrismaError";
+import { selectDeliverableOrThrow } from "../deliverable";
+import { UploadDocumentToDeliverableInput } from "./documentPendingUploadSchema";
+
+export async function handleUploadDocumentToDeliverable(
+  input: UploadDocumentToDeliverableInput,
+  ownerUserId: string,
+  isCmsAttachedFile: boolean
```
TARehman, line 10: NB: Do we care about this not working inside a transaction? I'm not sure how the getS3Adapter() part would even play with a transaction, so that's probably out of scope for this ticket.

PR #1272, opened 2026-05-19
task/collapse-db-migrations

TARehman:
Collapses all of the migrations into a set of two files.


server/src/model/migrations/20260312131729_prisma_init_baseline/migration.sql
```diff
@@ -1493,7 +1470,10 @@ ALTER TABLE "application_tag_assignment" ADD CONSTRAINT "application_tag_assignm
 ALTER TABLE "application_tag_suggestion" ADD CONSTRAINT "application_tag_suggestion_status_id_fkey" FOREIGN KEY ("status_id") REFERENCES "application_tag_suggestion_status"("id") ON DELETE RESTRICT ON UPDATE CASCADE;
 
 -- AddForeignKey
-ALTER TABLE "application_tag_suggestion_extract" ADD CONSTRAINT "application_tag_suggestion_extract_suggestion_id_applicati_fkey" FOREIGN KEY ("suggestion_id", "application_id", "value") REFERENCES "application_tag_suggestion"("id", "application_id", "value") ON DELETE RESTRICT ON UPDATE CASCADE;
+ALTER TABLE "application_tag_suggestion" ADD CONSTRAINT "application_tag_suggestion_application_id_fkey" FOREIGN KEY ("application_id") REFERENCES "application"("id") ON DELETE RESTRICT ON UPDATE CASCADE;
```
connor-parke-icf, line 1473: might be worth dropping a comment here, that this is a "virtual constraint" but not needed

connor-parke-icf (approved):
looks like all the non-permission migrations have moved into either the main prisma or main init migration. Comparing the final list of permissions in the migration with the list of consts, everything appears to match. I think we're good

PR #1279, opened 2026-05-20
DEMOS-2020-add-login-info-to-user-object

TARehman:
This just adds the last login field to the `User` object, which right now is only queried directly via `currentUser`. It technically is nullable even though in practice it's not possible to have a user get created without there being at least one session (because of how the creation happens coming from IDM).


connor-parke-icf (approved):
LGTM, nothing to comment on looks pretty much exactly what id expect it to. I think this is the first use of the aggregate prisma functions on the app which is neat i guess. Discussed why the field might be nullable and its a legitimate usecase

PR #1281, opened 2026-05-20
DEMOS-2117: a practical refactor for uipath

dustbuster:
This has been appended to for a while now, this simplifies the lambda and adjusts it to be holistically rewritten.

More info incoming.


dustbuster:
This is probably outdated now. and i got a lot of the needed updates into the last PR.

PR #1282, opened 2026-05-20
Demos 1991 remove unneeded permissions

connor-parke-icf:
after a whole lot of discussion, we reached the conclusion that for the scope of the app, a system that fully separates the concerns of Field-based authorization (what can you ask for) and the concerns of resource-based authorization (what things can you get back) doesnt bring enough value for the current scope. Essentially, 90% of permissions are fully defined already by the route from a critical parent object. This means that given our current API shape, the majority of it is tree-like and permissions are fully described by the relationship between resolvers. The only real violations of this rule are Demonstrations, Amendments, Extensions, Deliverables, and Documents. As such, these are the only types that need permission sets. For instance, if i can see a demonstration, i can see its amendments, their phases, and their phase dates transitively. 

So this PR is meant to peel back some of the changes implemented previously, where the goal was of larger, more generic scope. I suspect that _at some point_ in the future, many of the ideas here will need to implemented back in some way, but until we have a usecase that demands a more flexible system, we should eliminate these edges of the API.


server/src/model/user/queries/selectUser.test.ts
```diff
@@ -1,6 +1,6 @@
 import { User as PrismaUser } from "@prisma/client";
 import { beforeEach, describe, expect, it, vi } from "vitest";
-import { prisma, PrismaTransactionClient } from "../../../prismaClient";
+import { prisma } from "../../../prismaClient";
```
TARehman, line 3: Not sure why I had `PrismaTransactionClient` here? Maybe some kind of copy/paste error?

server/src/model/demonstrationRoleAssignment/queries/selectDemonstrationRoleAssignmentOrThrow.test.ts
```diff
@@ -0,0 +1,29 @@
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { selectDemonstrationRoleAssignment } from "./selectDemonstrationRoleAssignment";
+
+// Mock imports
+import { DemonstrationRoleAssignment as PrismaDemonstrationRoleAssignment } from "@prisma/client";
+import { selectDemonstrationRoleAssignmentOrThrow } from "./selectDemonstrationRoleAssignmentOrThrow";
+import { DemonstrationRoleAssignmentQueryResult } from ".";
+
+vi.mock("./selectDemonstrationRoleAssignment", () => ({
```
TARehman, line 9: NB: Again, I think it makes sense to do these above the import statements, but its not the end of the world or anything.

TARehman (approved):
Despite the large amount of code, since this is mostly a removal of some things and replacement with alternative code, it was fairly painless to review (about half the files changed are just deletions anyway).

server/src/model/user/queries/selectUser.test.ts
connor-parke-icf: im guessing that; as we tighten up the restrictions from the linter, these kinds of issues will drop

server/src/model/demonstrationRoleAssignment/queries/selectDemonstrationRoleAssignmentOrThrow.test.ts
connor-parke-icf: personally i like all the imports being at the top, but also it might make more sense for them to be imported after the mock. Ideally, id love to not have to import them at all and rely on the mock only (defining the function generically as just vi.fn()), but i havnt found a good way to do that. i feel like im missing something in how vitest is set up; feels like thats how it should just work

PR #1283, opened 2026-05-20
DEMOS-830-state-updates-and-fixes

TARehman:
This does a few different small things.

- Changes the sort order of the `STATES_AND_TERRITORIES` constant, which propagates to change the sort order of a few places on the front end.
- Adds in three new territories from the Compact of Free Association which were removed previously.
- Adds the region number to the `state` table in the database, which will be useful for DEMOS-2124 later.
- Updates the data diagrams accordingly.


server/src/model/migrations/20260520180111_state_updates_and_fixes/migration.sql
```diff
@@ -0,0 +1,114 @@
...
+
+-- Add check constraint for region, which checks if I missed any as well
+ALTER TABLE
+    demos_app.state
+ADD CONSTRAINT
+    check_region_is_valid
+CHECK (
+    region BETWEEN 1 AND 10
+);
+
```
connor-parke-icf, line 87: i was thinking about the prospect of making this into a static constraint table, but now that im seeing it, it would literally be a table of rows 1 to 10. so... in this case maybe it isnt worth it? ive seen tables that track only an incrementing ID before, but those were dynamic placeholding values. guessing this is fine enough validation. integers 1-10 _does_ completely describe the enumeration of possible values already; im surprised these dont have more descriptive names or something associated.

server/src/model/migrations/20260520180111_state_updates_and_fixes/migration.sql
```diff
@@ -0,0 +1,114 @@
...
+    demos_app.state AS s
+SET
+    region = srm.region
+FROM
+    state_region_map AS srm
+WHERE
+    s.id = srm.id;
+
+-- Drop the default
+ALTER TABLE demos_app.state ALTER COLUMN region DROP DEFAULT;
```
connor-parke-icf, line 77: ahh thats the annoying part. Need to have is constrained to something so you set a default when you make the table, update the values, drop the default, then add the final constraint. ok

server/src/constants.ts
```diff
@@ -336,28 +340,27 @@ export const STATES_AND_TERRITORIES = [
...
+  { id: "PW", name: "Republic of Palau" },
+  { id: "MH", name: "Republic of the Marshall Islands" },
   { id: "RI", name: "Rhode Island" },
   { id: "SC", name: "South Carolina" },
   { id: "SD", name: "South Dakota" },
   { id: "TN", name: "Tennessee" },
   { id: "TX", name: "Texas" },
+  { id: "VI", name: "US Virgin Islands" },
   { id: "UT", name: "Utah" },
   { id: "VT", name: "Vermont" },
```
connor-parke-icf, line 358: is this intended to be alphabetical order? if so, i think 356 and 357 are inverted (and a couple others like the Marshall Islands)

connor-parke-icf (approved):
LGTM! small PR but good examples of maneuvers im not super familiar with

server/src/model/migrations/20260520180111_state_updates_and_fixes/migration.sql
TARehman: Technically they're like `Region IX` I think, but I actually need the numbers for the purposes of the ID, so I would prefer them as integers. I agree that we could make it a static constraint, but it just didn't seem worth it in this _specific_ case.

server/src/constants.ts
TARehman: They're in alphabetical order by the display name per discussion with Dipika.

PR #1287, opened 2026-05-21
DEMOS-1991: remove unused imports/vars

connor-parke-icf:
noticed i had left some unused args in some of the resolvers which should be cleaned up. Doing so cascaded the changes to test files where i found some other unused vars. So decided to just remove the exception form the linter rule and clean up all the unused vars from the backend. Thus, 88 files changed but just about all are 1-token removals. Main boon is that the linter rules now enforce it, so we can catch these in the future.


TARehman (approved):
Lots of files but the actual changes are easy to check. LGTM.

PR #1288, opened 2026-05-21
Demos 1991 copy deliverable getx pattern to other types

connor-parke-icf:
During the development of the DeliverableData's getDeliverable function, we settled on a pattern that for the "singleton" resolvers, we should protect against returning null when the gql schema expects a non-null value.  Instead, it returns an explicit error stating that the row was either not found or the user doesnt have the permission to view it, logging in the case of a permission violation.  Coincidentally, state users dont have access to the amendments, extensions, or documents queries (because of directive permissions) but nontheless would rather keep the pattern consistent between each of these types at the fetch level. 

This PR copies that pattern for the other main objects such that our 5 core object each utilize this: Deliverables (already done), demonstrations, amendments, extensions, and documents.

### DIFF
--- a/server/src/model/amendment/amendmentData.test.ts
+++ b/server/src/model/amendment/amendmentData.test.ts
@@ -3,9 +3,16 @@ import { beforeEach, describe, expect, it, vi } from "vitest";
 import { buildAuthorizationFilter, ContextUser } from "../../auth";
 import { getAmendment, getManyAmendments } from "./amendmentData";
 import { selectAmendment, selectManyAmendments } from "./queries";
+import { log } from "../../log";
 
 vi.mock("../../auth", () => ({
   buildAuthorizationFilter: vi.fn(),
+}));
+
+vi.mock("../../log", () => ({
+  log: {
+    warn: vi.fn(),
+  },
 }));
 
 vi.mock("./queries", () => ({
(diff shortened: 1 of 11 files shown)


server/src/model/amendment/amendmentData.test.ts
```diff
@@ -45,17 +52,63 @@ describe("amendmentData", () => {
...
+      expect(buildAuthorizationFilter).toHaveBeenCalledWith(user, expect.any(Function));
+      expect(selectAmendment).toHaveBeenNthCalledWith(
+        1,
+        {
+          AND: [where, authFilter],
+        },
+        undefined
+      );
+      expect(selectAmendment).toHaveBeenNthCalledWith(2, where, undefined);
+      expect(log.warn).toHaveBeenCalledExactlyOnceWith(
```
TARehman, line 106: Since we're using Pino IIRC there may be a slightly more idiomatic way to do this where we inject a test sink / transport for logs, but I don't think its essential right now.

server/src/model/extension/extensionData.ts
```diff
@@ -27,24 +28,34 @@ export async function getExtension(
...
-  );
-}
+  const extension = await selectExtension(where, tx);
+  if (extension) {
+    log.warn(
+      `User ${user.id} attempted to access Extension ${extension.id} without sufficient permissions.`
+    );
+  }
 
+  throw new Error("Requested Extension not found or User does not have Permission to view it.");
```
TARehman, line 57: Nit: I don't know if Permission should be capitalized in these sentences.

server/src/model/extension/extensionData.test.ts
```diff
@@ -73,9 +126,11 @@ describe("extensionData", () => {
       expect(result).toBe(extension);
     });
 
-    it("passes transaction client to selectManyAmendments if provided", async () => {
```
TARehman, line 76: Was this just a typo leftover from a previous copy/paste?

server/src/model/document/documentResolvers.ts
TARehman, line 107: Should pull this out before launch if we can; F/E can just query the actual document and parse what comes back.

server/src/model/deliverable/deliverableData.ts
```diff
@@ -28,14 +28,12 @@ export async function getDeliverable(
   where: Prisma.DeliverableWhereInput,
   user: ContextUser,
   tx?: PrismaTransactionClient
-): Promise<PrismaDeliverable | null> {
+): Promise<PrismaDeliverable> {
```
TARehman, line 31: Interesting that this change didn't require a test change; presumably because we don't look at the actual return sig?

TARehman (approved):
Small comments. LGTM overall. 🚢

server/src/model/extension/extensionData.test.ts
connor-parke-icf: yeah, basically any change ive made to extensions and amendments has been do it once, then copy/paste find/replace since theyre basically the same. but mustve missed one.

server/src/model/deliverable/deliverableData.ts
connor-parke-icf: the tests would have no way of knowing what our ts is saying, so no real way to catch in runtime (that i know of). Currently, we dont specify the return types in a lot of our resolvers. I have another PR coming up that addresses the nullability in the schema, and will update the resolver functions to match that expected return type.  For like... total alignment maybe we could define a type specifically for the resolver shape that lives in the schema functions and use a custom linter to match them, but... not worth it as of now.

server/src/model/document/documentResolvers.ts
connor-parke-icf: agreed; but will involve frontend changes. Currently, kinda awkward to support with how the getDocument works (finds or throws). caught this in a followup PR im making where i try to type the return types of each query to match.

server/src/model/extension/extensionData.ts
connor-parke-icf: gona defer this for a future audit of return messaging.  I think youre right though, Permission is much less of a proper noun than Extension is in this context.

server/src/model/amendment/amendmentData.test.ts
connor-parke-icf: yeah, we should probably make a audit of the logging. It would be good if we could route all the permissions violations (like from directives) in the same way.

PR #1291, opened 2026-05-21
Demos 2123 - Add extract of Acutals and Net Variance 

james-becwar-cms:
image

Adds Actuals and net_variance_total to budget_neutrality_workbook table, modifies validator to also extract values.


data/docs/DEMOS_Data_Model.mmd
```diff
@@ -425,8 +423,11 @@ erDiagram
     text document_type_id FK "FK119: (id, document_type_id) ∈ document.(id, document_type_id) DEFERRED<br>FK120: document_type_id ∈ budget_neutrality_workbook_document_type_limit.id"
     text validation_status_id FK "FK121: validation_status_id ∈ budget_neutrality_validation_status.id"
     jsonb validation_data
+    text actuals
+    float netVarianceTotal
```
connor-parke-icf: not sure how much it matters, but the only floating point number we have in the datamodel is defined as a double, not a float. i dont know the nuances of that in postgres tho.

we should name it net_variance_total in camel_case though.

lambdas/budgetNeutrality/index.ts
```diff
@@ -27,21 +28,25 @@ interface Results {
...
 
   await pool.query(query, [
     message.documentId,
     message.documentTypeId,
-    validationResults.length > 0 ? FAILED_VALIDATION_STATUS_ID : SUCCEEDED_VALIDATION_STATUS_ID,
-    JSON.stringify(validationResults),
+    validationResults.isValid ? SUCCEEDED_VALIDATION_STATUS_ID : FAILED_VALIDATION_STATUS_ID,
+    JSON.stringify(validationResults.errors),
+    validationResults.extractedValues?.get("actuals"),
+    validationResults.extractedValues?.get("netVariance"),
```
connor-parke-icf, line 49: hmm something about the extractedValues being nullable here gives me pause. Looking at the .mmd, theres a notion of status being succeeded, so it makes sense that there could be nullable values in here, but we're nopt specifying that in the .mmd with "NULLABLE".  Could probably do with some explicit edge case checking or more explicit null checks

server/src/model/budgetNeutralityWorkbook/budgetNeutralityWorkbookHistory.prisma
```diff
@@ -6,6 +6,8 @@ model BudgetNeutralityWorkbookHistory {
   documentTypeId     String       @map("document_type_id")
   validationStatusId String       @map("validation_status_id")
   validationData     Json         @map("validation_data")
+  actuals            String?
+  netVarianceTotal   Float?       @map("net_variance_total")
```
connor-parke-icf, line 10: the casing is correct here, but in the direct sql in the lambda looks like we reference it as netVarianceTotal and not net_variance_total, which would break.

shared_library/src/BN/extractors/index.test.ts
```diff
@@ -0,0 +1,33 @@
+import { describe, expect, it } from "vitest";
+
+import type { ExcelData } from "../index.js";
+import { extractions } from "./index.js";
+
+function buildWorkbook(b8: string , as436:  number ): ExcelData {
+  const summaryData = Array.from({ length: 436 }, () => [] as (string | number | null)[]);
+
+  summaryData[7]![1] = b8 ?? null;
+  summaryData[435]![18] = as436 ?? null;
```
connor-parke-icf: guessing b8 and as346 are just arbitrary cells to check against?

shared_library/src/BN/extractors/index.test.ts
```diff
@@ -0,0 +1,33 @@
...
+    expect(result).toEqual(new Map([["actuals", "FY25"]]));
+  });
+
+
+  it("extracts net variance from cell AS436", () => {
+    const data = buildWorkbook("FY25", 3200.5);
+
+    const result = extractions[1]!(data);
+
+    expect(result).toEqual(new Map([["netVariance", 3200.5]]));
```
connor-parke-icf: i see, so we're building up an array that contains just the critical data we want to extract as an "extraction" object? Im a little hesistant on what looks like dereferencing using an array index rather than a name

shared_library/src/BN/extractors/index.ts
```diff
@@ -0,0 +1,22 @@
+import type {ExtractionFunction} from "../validation.js";
+import {excelColumnRow} from "../index.js";
+
+
+export const extractions: ExtractionFunction[] = [
+  function extractActuals(data) {
+    const b8 = excelColumnRow("B8", "Summary", data);
```
connor-parke-icf: hmm... i would rather use a const called like "SUMMARY_CELL" or something, B8 doesnt tell us too much. Also, we should compare against a few different workbooks. I suspect b8 doesnt move around much, since its close to the top, but maybe theres an indicator or field label we can search against to find an adjacent cell to pull from?  Im more concerned about this for AS436 since it feels much more arbitrary of a location

shared_library/src/BN/validation.test.ts
```diff
@@ -4,15 +4,19 @@ import type { ExcelData } from "./index.js";
...
     const data: ExcelData = [];
 
-    const result = await validateBNWorkbook(data, []);
+    const result = await validateBNWorkbook(data, [], []);
 
-    expect(result).toEqual([]);
+    expect(result).toEqual({
+      isValid: true,
+      errors: [],
+      extractedValues: new Map(),
```
connor-parke-icf, line 15: i think instead of a dynamic map, we could define a more rigid type that better describes what we expect to get out of the workbook.  Sure, if we need to add more data slots (unlikely) to it later itd be a little more cumbersome, but it would help define a clearer contract

connor-parke-icf (commented):
there are some minor issues with variable casing and column naming/referencing, but those should be pretty obvious when its time to run. 

Overall, i think we may be overgenericizing the extraction a little bit; we only have a small handful of fields we want to pull from the workbook, so i think it would be worth it to make a consistent structure for the extractions with expected types/immutability for each field rather than packing it all into a more generic Map object with values of number | string. 

its a little unclear whether the new columns in the DB should be nullable or not; i think this is intending for them to be nullable, but makes me think that if the extraction failed do we really need to hold onto that data?

server/src/model/budgetNeutralityWorkbook/budgetNeutralityWorkbookHistory.prisma
james-becwar-cms: Thank you, good catch.

shared_library/src/BN/extractors/index.test.ts
james-becwar-cms: Its cell B8 and cell AS436.  This is basically making an in memory representation of the xlsx file.

lambdas/budgetNeutrality/index.ts
james-becwar-cms: Yeah Tom and I talked about the same, but we figured this is good enough for now.  The extractors will throw if the value isn't of the expected type for the value.

shared_library/src/BN/validation.test.ts
james-becwar-cms: I'm 99% sure we will be adding more extractors the sprint after MVP launch which is why I went with the map.

data/docs/DEMOS_Data_Model.mmd
james-becwar-cms: Thanks I updated it.

shared_library/src/BN/extractors/index.test.ts
james-becwar-cms: This ended up hiding a bug, so I rewrote it to load a simple xlsx and pull the values from it.

shared_library/src/BN/extractors/index.test.ts
james-becwar-cms: Yeah this wasn't great, I ended up rewriting it to be more explicit.

shared_library/src/BN/extractors/index.ts
james-becwar-cms: Normally I would agree with you, but when talking to Vivian I asked her about the actuals /net variance and she didn't know what  was talking about.  When I showed her the sheet she was like ok B8 and AS436, so its the business talks about it.  Other mitigating factor is that the var is only used for 3 lines, so I think its a light cognitive load.

shared_library/src/BN/extractors/index.test.ts
```diff
@@ -1,33 +1,32 @@
...
-  it("extracts net variance from cell AS436", () => {
-    const data = buildWorkbook("FY25", 3200.5);
-
-    const result = extractions[1]!(data);
-
-    expect(result).toEqual(new Map([["netVariance", 3200.5]]));
+  it("run all extractors and verify output.", async () => {
+    
+    const currentDir = path.dirname(fileURLToPath(import.meta.url));
+    const fixturePath = path.join(currentDir, "..", ".." , "..", "test", "fixtures", "sample.xlsx");
```
connor-parke-icf, line 12: NB, but we should prob use an absolute path here, in case we want to shift this file's location around. That being said, i like that we are testing against an actual file. Maybe... we should keep that file next to this test, but i dont think it matters either way

connor-parke-icf (approved):
LGTM, no blockers. my concerns from last review were addressed either with a comment or a change.  nice!

shared_library/src/BN/extractors/index.test.ts
james-becwar-cms: I moved it out of the source directory so it wouldn't be included on deploy/build.

PR #1293, opened 2026-05-22
DEMOS-1785: configured start and end dates for phase 8

connor-parke-icf:
Thought it was curious why this wasnt working already. No _real_ code change needed, just a configuration of a const. Not sure how that was missed, but now correctly captures the start date and not just the status. 

Tried to find some context to see if this was intentional, since we had a test case for it, but from what i could find i think that was just to cover a technicallypossible codepath, but wasnt actually a requirement

### DIFF
--- a/server/src/constants.ts
+++ b/server/src/constants.ts
@@ -436,7 +436,10 @@ export const PHASE_START_END_DATES: PhaseStartEndDateRecord = {
     startDate: "Approval Package Start Date",
     endDate: "Approval Package Completion Date",
   },
-  "Approval Summary": {},
+  "Approval Summary": {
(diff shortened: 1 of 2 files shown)


server/src/constants.ts
```diff
@@ -436,7 +436,10 @@ export const PHASE_START_END_DATES: PhaseStartEndDateRecord = {
     startDate: "Approval Package Start Date",
     endDate: "Approval Package Completion Date",
   },
-  "Approval Summary": {},
```
connor-parke-icf, line 439: well theres ya problem!

james-becwar-cms (approved):

PR #1294, opened 2026-05-22
Demos 1991 remove search people query

connor-parke-icf:
One of the goals of this permissions epic was to reduce the overall surface area of the API, this meant removing queries that were not in use or that could be easily refactored out. As i was examining the searchPeople query, i realized that it was incorrectly filtering, causing the search string filter to get overwritten. I examined frontend to see why this wasnt causing problems, and it was because we are implementing the filtrering behavior there as well. So i went ahead and removed this searchPeople query, implementing the only part that was missing from the filtering behavior: state association. Now the frontend just calls the people query, and handles all the filtering it needs itself. 

because of the dependence on this search behavior, there are a lot of aspects to this frontend component that are no longer needed; this whole component could probably lose over 50% of its lines.  However, with release on the horizon and the prospect of redesigning this component, its not worth refactoring now. Instead, i did the minimum amount of changes to get it working with Query.people.

Finally, i found one bug when fetching demonstration role assignments because of a permission that should not belong; so that auth directive was removed.

### DIFF
--- a/client/src/components/dialog/ManageContactsDialog.tsx
+++ b/client/src/components/dialog/ManageContactsDialog.tsx
@@ -24,19 +24,22 @@ import type { ContactRow, ContactType } from "../table/columns/ContactColumns";
 import { ContactColumns } from "../table/columns/ContactColumns";
 import { PaginationControls } from "../table/PaginationControls";
 
-const SEARCH_PEOPLE_QUERY = gql`
-  query SearchPeople($search: String!, $demonstrationId: ID) {
-    searchPeople(search: $search, demonstrationId: $demonstrationId) {
+export const SEARCH_PEOPLE_QUERY = gql`
+  query ManageContactsQuery {
+    people {
       id
       firstName
       lastName
       email
       personType
+      states {
+        id
+      }
     }
   }
 `;
 
-const SET_DEMONSTRATION_ROLE_MUTATION = gql`
+export const SET_DEMONSTRATION_ROLE_MUTATION = gql`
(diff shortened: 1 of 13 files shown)


server/src/model/person/personResolvers.ts
```diff
@@ -15,52 +15,6 @@ export const personResolvers = {
...
-        const demonstration = await prisma().demonstration.findUnique({
-          where: { id: demonstrationId },
-          include: { state: true },
-        });
-
-        if (demonstration) {
-          return await prisma().person.findMany({
-            where: {
-              ...baseWhere,
-              OR: [
```
connor-parke-icf, line 42: here, there was a bug. when demonstration was non null and it attempted to merge the baseWhere with the new OR clause, it would actually replace the existing OR clause. This meant that whenever a demonstration was used in the search, it would replace the filters for the actual query string.

client/src/components/dialog/ManageContactsDialog.tsx
```diff
@@ -594,21 +588,26 @@ export const ManageContactsDialog: React.FC<ManageContactsDialogProps> = ({
...
-
-                return remainingRoles.length > 0;
-              }).filter(
-                (p, index, arr) => arr.findIndex((item) => item.id === p.id) === index
-              );
+              let filteredResults = searchResults
+                .filter((person) => {
+                  return person.states.some((s) => s.id === stateId);
+                })
+                .filter((p) => {
```
connor-parke-icf, line 595: new filter, just compares the person's states with the state on the demonstration, passed in.

james-becwar-cms (approved):

PR #1295, opened 2026-05-22
Demos 1991 remove unused queries

connor-parke-icf:
further reduction of surface area; these queries are not being called from anywhere, so removing them. Might go a bit further and remove State from the api altogether, something we had discussed. Either way, fewer queries means fewer paths without inherent filtering from a parent, and the more closely the gql graphql pulls directly from the permissions structure


dustbuster (approved):
ok

PR #1299, opened 2026-05-26
DEMOS-1937— Date field fixes

nunesmlucas:
# DEMOS-1937 — Date field fixes

Follow-up to the DEMOS-2004 spike and DEMOS-1800 (which already fixed the main
"can't type the year" bug). This PR cleans up what was left.

## What it fixes

- **The box now matches what gets saved.** Out-of-range dates no longer get
  silently dropped; whatever you leave in the field is what the form holds.
- **Validation limits are consistent.** The browser and our code now agree on
  what's in range (and the range is inclusive on both ends).
- **Editing an existing date no longer breaks.** Typing into a subfield (mm,
  dd, yyyy) keeps focus, the cursor advances normally, and the calendar
  pop-up stays open when you use the arrow keys.
- **Past dates are never trapped.** A prior Due Date or other historical date
  loads, displays, and can still be edited or submitted unchanged.

https://github.com/user-attachments/assets/dc5fc61c-3e14-44c8-a0f9-1f3d40d0a4e6

### DIFF
--- a/client/src/components/input/date/DatePicker.tsx
+++ b/client/src/components/input/date/DatePicker.tsx
@@ -35,25 +35,35 @@ export const DatePicker: React.FC<DatePickerProps> = ({
   getValidationMessage,
 }) => {
   const inputRef = useRef<HTMLInputElement>(null);
+  const isFocusedRef = useRef(false);
   const inputMin = minDate ?? DEFAULT_MIN_DATE;
   const inputMax = maxDate ?? DEFAULT_MAX_DATE;
 
-  // The input is uncontrolled (defaultValue + ref-sync) instead of fully controlled.
-  // Native <input type="date"> fires input events with value="" while the user is mid-typing the
-  // year subfield (the date is briefly incomplete). A controlled input would re-render on every
-  // such event and wipe the in-progress digits. With this ref-sync pattern, React only touches
-  // the DOM when the value prop changes externally.
+  // The input is uncontrolled (defaultValue + ref-sync). Native <input type="date"> is fragile:
+  // any DOM mutation, parent re-render, or write to .value while the field is focused can drop
+  // focus, reset the mm→dd→yyyy subfield cursor, or close the native calendar. So while the user
+  // is interacting with the field, React stays out — neither writing to the DOM here nor pushing
+  // updates to the parent (see handleBlur). The DOM is only synced from the value prop on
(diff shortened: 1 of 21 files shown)


client/src/components/input/date/DatePicker.tsx
```diff
@@ -29,33 +31,52 @@ export const DatePicker: React.FC<DatePickerProps> = ({
...
+  // any DOM mutation, parent re-render, or write to .value while the field is focused can drop
+  // focus, reset the mm→dd→yyyy subfield cursor, or close the native calendar. So while the user
+  // is interacting with the field, React stays out — neither writing to the DOM here nor pushing
+  // updates to the parent (see handleBlur). The DOM is only synced from the value prop on
+  // external changes (load/reset) and only when the field is not focused.
   useEffect(() => {
-    if (inputRef.current && inputRef.current.value !== (value ?? "")) {
-      inputRef.current.value = value ?? "";
+    const input = inputRef.current;
+    if (!input || isFocusedRef.current) return;
```
connor-parke-icf, line 50: if i understand right, if the field is currently being focused, we just return to ignore any external updates to the field? I think this works; i dont think we have any operations that might attempt to change the field while the user is currently looking at it.

connor-parke-icf (approved):
i think this is fine. I know there at least used to be some scenarios where we would save something on change of the date field, but even that i think would require a blur operation. Beyond that i dont think there are any scenarios where we want to update the contents of the field while the user is actively looking at the field.  Im a little nervous at how much customization is needed for the date picker to work consistently across browsers, but from what ive seen thats more or less the nature of them; im still surprised the various browsers havnt come together on a consistent spec at this point. lgtm

client/src/components/input/date/DatePicker.tsx
nunesmlucas: yep, you've got it right — while the field is focused, we let the browser own it, because writing to `.value` mid-edit drops focus, resets the `mm/dd/yyyy` cursor, and closes the calendar (the two issues caught). I checked the callers and none push external updates while a date field is being edited, so this is safe.

Totally agree on the cross-browser frustration — `<input type="date">` is just genuinely inconsistent. Going blur-only (commit when focus leaves) is the same pattern "React Hook Form" and "Formik" use by default based on what i researched, so at least all the quirky handling stays centralized in this one component and callers don't have to think about it.

PR #1300, opened 2026-05-26
DEMOS-2148: Remove Some Export and Import Icons / Actions from tables

tresdonicf:
Per DEMOS-2148, removes the export icon from the types table (which had no action) and the import button from the documents table (which was redundant with the more explicit "Add Document" Button)

image

image


james-becwar-cms (approved):

connor-parke-icf (approved):
nice! always good when we can remove a prop as well

PR #1301, opened 2026-05-26
Demos 1991 readd permissions for demonstration role assignment

connor-parke-icf:
because we have a query in use on the frontend that fetches deliverables via the person object, we cannot simply block off the entire person.roles field, since state users need access to it. So, we can utilize the row based permission such that they only have access to the demonstrationRoleAssignments that belong to demonstrations they are assigned to.  This is the same structure as the other permissions we have introduced; with this i believe the row based permissions model is complete.


server/src/model/person/personResolvers.ts
```diff
@@ -19,8 +20,12 @@ export const personResolvers = {
   Person: {
     fullName: (parent: PrismaPerson): string => `${parent.firstName} ${parent.lastName}`,
     personType: (parent: PrismaPerson): PersonType => parent.personTypeId as PersonType,
-    roles: (parent: PrismaPerson): Promise<PrismaDemonstrationRoleAssignment[]> =>
-      selectManyDemonstrationRoleAssignments({ personId: parent.id }),
+    roles: (
```
connor-parke-icf, line 21: we do not need a simialr check from the demonstration object, since that should be transitive. The case here protects against fetching demonstrations from unrelated users, like in the case of Demonstration.deliverables[].cmsOwner.person.roles[].demonstration

server/src/model/migrations/20260526140859_initialize_demonstration_role_assignment_permissions/migration.sql
```diff
@@ -0,0 +1,16 @@
...
+
+INSERT INTO
+    demos_app.role_permission
+VALUES
+    ('Admin User', 'System', 'View All DemonstrationRoleAssignments'),
+    ('Admin User', 'System', 'View DemonstrationRoleAssignments on Assigned Demonstrations'),
+    ('CMS User', 'System', 'View All DemonstrationRoleAssignments'),
+    ('CMS User', 'System', 'View DemonstrationRoleAssignments on Assigned Demonstrations'),
+    ('State User', 'System', 'View DemonstrationRoleAssignments on Assigned Demonstrations')
+;
```
dustbuster: Missing a newline at EOF.

dustbuster (approved):
Add that newline on the migration plz. Not gonna block it. But like these little red circles do something to me!

Image

PR #1304, opened 2026-05-26
DEMOS-184: added new logo, wrapped in component

connor-parke-icf:
changes out the main logo to the new one 
Screenshot 2026-05-26 121807


client/src/components/brand/Logo.tsx
```diff
@@ -0,0 +1,11 @@
+import React from "react";
+
+const LOGO_SRC = "/img/logo.svg";
+
+export const Logo: React.FC = () => {
```
connor-parke-icf, line 7: moved this to its own file since i thought it would be reused in the footer, but looking again theyre actually slightly different. Nonetheless, would be good to place these logo things into a home since the footer will have a couple more to add.

client/src/components/brand/Logo.tsx
```diff
@@ -0,0 +1,11 @@
...
+  return (
+    <a href="/" className="h-[40px]">
+      <img height="40px" src={LOGO_SRC} alt="Logo" />
```
james-becwar-cms: Should we update the alt text to something like DEMOS logo, or something more descriptive?

james-becwar-cms (approved):

client/src/components/brand/Logo.tsx
connor-parke-icf: definitely, fixed! added a testId for the tests to reference instead and the alt text: "DEMOS Logo with text: DEMOS, Demonstration Evaluation Management & Oversight System"

PR #1305, opened 2026-05-26
Demos 2018 update footer for demos

connor-parke-icf:
Implements the new design of the footer. Wasnt too bad to implement, but we are still waiting on a cleaner simplified logo image as the one we have no is too low res. I have also gone ahead and implemented a boilerplate of how the FAQ download would work, but we dont currently have support for that in the server so it just returns an error. 
Demo: 

https://github.com/user-attachments/assets/f9005eb3-b94d-4530-bccb-483c7507f725


client/public/img/logo-simplified.png
connor-parke-icf, line 1: waiting on a cleaner logo image currently, this one is far too low resolution. hopefully, will be SVG

client/src/components/footer/Footer.tsx
```diff
@@ -1,11 +1,81 @@
+import { HhsLogo } from "components/brand/HhsLogo";
+import { LogoSimplified } from "components/brand/LogoSimplified";
+import { useToast } from "components/toast";
 import React from "react";
+import { tw } from "tags/tw";
+import { useDownloadFaq } from "./useDownloadFaq";
+
+const DEMOS_ADDRESS = "7500 Security Boulevard Baltimore, MD 21244";
+const DEMOS_VERSION = "1.0.001";
```
connor-parke-icf, line 9: wasnt sure the best place to put this meta info. Just keeping in the footer for now, but we might could use a separate file for these.

client/src/components/footer/triggerFaqDownload.ts
```diff
@@ -0,0 +1,8 @@
+export const triggerFaqDownload = (url: string) => {
+  const link = document.createElement("a");
+  link.href = url;
+  link.rel = "noopener noreferrer";
+  document.body.appendChild(link);
+  link.click();
+  document.body.removeChild(link);
```
connor-parke-icf, line 7: admittedly, this doesnt feel quite right but seems to work

client/src/components/footer/useDownloadFaq.ts
```diff
@@ -0,0 +1,41 @@
+import { gql, TypedDocumentNode, useLazyQuery } from "@apollo/client";
+import { triggerFaqDownload } from "./triggerFaqDownload";
+
+// TODO: server support for this query does not exist yet, implemented to
+// demonstrate the intended flow between the server and s3. This is likely
+// not the intended query
+export const DOWNLOAD_FAQ_QUERY: TypedDocumentNode<{
```
connor-parke-icf, line 7: this more serves as a mock until we have actual server support for reference materials. I imagine we might have a query for fetching a single reference material by name or ID or something. Or we might have a dedicated query, like we have here. Not sure yet. either way, shows that on click of reference materials, we query the server, get back a presigned, then use it to download the file.

client/src/layout/PrimaryLayout.tsx
```diff
@@ -31,9 +31,9 @@ export const PrimaryLayout = ({ children }: { children: React.ReactNode }) => {
             {!hideSideNav && <SideNav />}
             <div className="flex-1 overflow-auto min-h-0 flex flex-col">
               <div className="p-[16px] flex-1">{children}</div>
-              <Footer />
             </div>
           </div>
+          <Footer />
```
connor-parke-icf, line 36: according the the figma, the Footer is shown at full page width, covering up the bottom part of the sidebar.  Currently, only conflicts locally because of all the debug links so im guessing there may be a cleaner way to handle this. But for now, i think its fine.

client/public/img/logo-simplified.png
dustbuster: I tried to paste an SVG i found here, HOPING it would render as HTML, but just comes up blank. 👎

client/src/components/footer/triggerFaqDownload.ts
dustbuster: HERE'S A CHILD. sike nvm...

dustbuster (approved):
GO YOLO!

client/src/layout/PrimaryLayout.tsx
tresdonicf: This seems solid!

client/src/components/brand/Logo.test.tsx
```diff
@@ -9,7 +9,7 @@ describe("Logo", () => {
     expect(logoImage).toBeInTheDocument();
     expect(logoImage.src).toContain("/img/logo.svg");
     expect(logoImage.alt).toBe(
-      "DEMOS Logo with text: DEMOS, Demonstration Evaluation Management & Oversight System"
+      "DEMOS Logo with text: 1115, DEMOS, Demonstration Evaluation Management & Oversight System"
```
tresdonicf: While it seems a bit odd (because it should be trivial) we could maybe just check that this is equal to `LOGO_ALT` rather than repeat the text. Suggesting that what we care about for this test is that the `alt` attribute text is set correctly

tresdonicf (approved):
Looks good! Really appreciate splitting components up into separate files 😌

client/src/components/brand/Logo.test.tsx
connor-parke-icf: yeah it definitely falls into the realm of testing for a configuration, rather than a behavior. Im down to export the const and test against it instead of the string directly

PR #1307, opened 2026-05-26
DEMOS-2085 — Deliverable State/CMS File Tab Functionality

nunesmlucas:
# DEMOS-2085 — Deliverable State/CMS File Tab Functionality

Wires up add/edit/delete/submit for files on the Deliverables Page, gated by user role and constrained by the deliverable-type → document-type DB mapping.

- **State Files tab** — upload, edit, delete files; submit deliverable (enabled once ≥1 file exists)
- **CMS Files tab** — upload (no File Type field, defaults to General File), edit, delete; **hidden entirely from State users**
- File Type dropdowns are constrained per deliverable type (e.g. Monitoring Report → "Monitoring Report" + "General File")

Mostly reusing our dialogs and document mutations for this..

- `Deliverable.allowedDocumentTypes` GraphQL field, resolved from the existing `deliverable_type_document_type` table

- `AddDocumentToDeliverableDialog` — thin upload wrapper, picks the right mutation by `isCmsFile`
(Let me know if there's any issue with this, since we normally have them separate across our codebase.. i thought it would be appropriate for just the 'Deliverables')

## Out of scope

- "Current" file toggle column wiring — not in the AC

https://github.com/user-attachments/assets/5b902420-8931-4d34-94d2-654f04ad0206


client/src/components/dialog/document/DocumentDialog.tsx
```diff
@@ -440,14 +442,16 @@ export const DocumentDialog: React.FC<DocumentDialogProps> = ({
...
+      {!hideDocumentType && (
+        <DocumentTypeInput
+          value={activeDocument.documentType}
+          onSelect={(val) =>
+            setActiveDocument((prev) => ({ ...prev, documentType: val as DocumentType }))
+          }
+          documentTypeSubset={documentTypeSubset}
+          canEditDocumentType={canEditDocumentType}
+        />
+      )}
```
tresdonicf, line 454: I wonder if this behavior could be implemented using the already existing `documentTypeSubset` instead of adding a prop. so if `[]` was passed it would hide it since there's no document types select-able. Not a huge deal just another way to think about it if it's to support a relatively niche use-case

tresdonicf (approved):
I think the wrapper in `AddDocumentToDeliverable` approach is reasonable for now! Might make sense at some point to revisit calling a conditional mutation in the main dialog but certainly not needed immediately

PR #1309, opened 2026-05-27
WIP DEMOS-1912 update deliverable creation validation for specific demo t…

JeffreyM-GA:
…ypes

Have the code written for the migration to validate at DB level, will add shortly.


server/src/model/deliverable/validateDeliverableInputs.ts
```diff
@@ -80,6 +80,14 @@ export async function validateCreateDeliverableInput(
...
     checkDueDateInFuture(input.dueDate)
   );
+  if (
+    REQUIRED_DEMO_TYPE_DELIVERABLES.includes(input.deliverableType) &&
+    (!input.demonstrationTypes || input.demonstrationTypes.size === 0)
+  ) {
+    errors.push(
+      `Deliverable type ${input.deliverableType} requires at least one demonstration type`
+    );
+  }
```
connor-parke-icf: might be cleaner to put this in a helper so it can be validated alongside the other checkX() functions above it, even if it is a tiny check.

server/src/constants.ts
```diff
@@ -80,6 +80,12 @@ export const ACTIVE_DELIVERABLE_STATUSES = [
   "Under CMS Review",
 ] as const satisfies DeliverableStatus[];
 
+export type DeliverableType = (typeof DELIVERABLE_TYPES)[number];
```
connor-parke-icf: have a DeliverableType under the types.ts file; if we're moving it to here, we might should remove it from there, similar to how we handled DeliverableStatus

connor-parke-icf (approved):
no real blockers, just organizational suggestions. I do think we should remove the previous definition of DeliverableType in types.ts, since it looks like we're moving the def to the constants file. overall, im not sure how much i like the separation of types and constants we have; it kinda breaks down when we want to define a subset of another constant as a type of that constant as you are doing here.  but thats a bigger change than whats needed here

server/src/constants.ts
JeffreyM-GA: Yeah... I thought it was kinda weird. This file currently doesn't have any imports, it felt wrong to add one since it's kind of a self contained constants definition right now. I'll try removing it from types.

server/src/model/migrations/20260527133836_add_demo_type_deliverable_validation/migration.sql
```diff
@@ -0,0 +1,52 @@
...
+ON demos_app.deliverable_demonstration_type
+DEFERRABLE INITIALLY DEFERRED
+FOR EACH ROW
+EXECUTE FUNCTION demos_app.validate_required_deliverable_demo_types();
+
+DROP TRIGGER IF EXISTS validate_required_demo_types_on_deliverable
+ON demos_app.deliverable;
+
+CREATE CONSTRAINT TRIGGER validate_required_demo_types_on_deliverable
+AFTER INSERT OR UPDATE OF deliverable_type_id OR DELETE
```
connor-parke-icf: on delete of a deliverable, i dont think we need to validate anything, since its just fail a FK constraint if deliverable_demonstration_type still held something pointing to the deleted deliverable.

server/src/model/migrations/20260527133836_add_demo_type_deliverable_validation/migration.sql
```diff
@@ -0,0 +1,52 @@
+CREATE OR REPLACE FUNCTION demos_app.validate_required_deliverable_demo_types()
+RETURNS TRIGGER
+LANGUAGE plpgsql
+AS $$
+DECLARE
+    target_deliverable_id UUID;
+    deliverable_type_id TEXT;
+    remaining_count INTEGER;
+BEGIN
+    target_deliverable_id :=
```
connor-parke-icf: on change of the deliverable, wouldnt the deliverable id field just be "id" instead of deliverable_id? It may be simpler here to break this into two triggers, since one is coming from the perspective of a deliverable and the other is from the perspective of a deliverable_demonstration_type.  Im not sure the best way to handle it though, cause the inner part is the same: find all the deliverable demonstration types on a deliverable, if there arent any, throw. Actually, i think we have a similar constraint to this one for checking if demonstrations have at least one demo type when theyre approved. 

```
-- check_demonstration_type_exists_for_approved_demos
CREATE OR REPLACE FUNCTION demos_app.check_demonstration_type_exists_for_approved_demonstrations()
RETURNS TRIGGER
LANGUAGE plpgsql
AS $$
DECLARE
    demonstration_status TEXT;
    demonstration_type_assignment_count INT;
BEGIN
    IF TG_TABLE_NAME = 'demonstration_type_tag_assignment' AND TG_OP = 'DELETE' THEN
        SELECT
            status_id
        FROM
            demos_app.demonstration
        WHERE
            id = OLD.demonstration_id
        INTO
            demonstration_status;

        IF demonstration_status = 'Approved' THEN
            -- Remember, this count is before the execution of the delete
            -- That is why we look for a count of 1
            SELECT
                COUNT(*)
            FROM
                demos_app.demonstration_type_tag_assignment
            WHERE
                demonstration_id = OLD.demonstration_id
            INTO
                demonstration_type_assignment_count;

            IF demonstration_type_assignment_count = 1 THEN
                RAISE EXCEPTION 'Cannot delete the final demonstration type from approved demonstration %', OLD.demonstration_id;
            END IF;
        END IF;
        RETURN OLD;

    ELSIF TG_TABLE_NAME = 'demonstration' AND TG_OP = 'UPDATE' THEN
        IF (OLD.status_id IS DISTINCT FROM NEW.status_id) AND (NEW.status_id = 'Approved') THEN
            SELECT
                COUNT(*)
            FROM
                demos_app.demonstration_type_tag_assignment
            WHERE
                demonstration_id = NEW.id
            INTO
                demonstration_type_assignment_count;

            -- In this case we check for zero because the operation is on the demonstration table
            IF demonstration_type_assignment_count = 0 THEN
                RAISE EXCEPTION 'Cannot set demonstration % to Approved without at least one demonstration type', NEW.id;
            END IF;
        END IF;
        RETURN NEW;

    ELSE
        RAISE EXCEPTION 'This trigger has been assigned to an incorrect table/operation combination';
    END IF;
END;
$$;
```
ahh, ok looks like we check the table name and operation. if its a change on demonstration_type_tag_assignment and a delete, we count and check the status.  if its a change on demonstration and an update, we count and check the status. 

I suppose we can do similar here:
if its a delete on deliverable_demonstration_type, we check the deliverable_type and count. Or if its a insert or update on deliverable, we check the new deliverable_type and count.

server/src/constants.ts
TARehman: So the places where we do this are entirely based on avoiding circular imports. Sometimes, in `constants.ts`, we want to use a type as part of limiting things. If you define the constant, export it, then define the type in `types.ts`, then import from that back into `constants.ts`, you get circular dependency complaints. That's why there's a _small_ amount of types defined originally in `constants.ts` and then imported and re-exported from `types.ts`. 🙂

server/src/constants.ts
```diff
@@ -80,6 +80,12 @@ export const ACTIVE_DELIVERABLE_STATUSES = [
...
+export const REQUIRED_DEMO_TYPE_DELIVERABLES: readonly DeliverableType[] = [
```
TARehman: This needs to be exported `as const` I think.

server/src/types.ts
```diff
@@ -164,7 +164,6 @@ export type TagSource = (typeof TAG_SOURCES)[number];
 export type TagType = (typeof TAG_TYPES)[number];
 export type BudgetNeutralityValidationStatus =
   (typeof BUDGET_NEUTRALITY_VALIDATION_STATUSES)[number];
-export type DeliverableType = (typeof DELIVERABLE_TYPES)[number];
```
TARehman, line 167: So there's a section of the file that exports types that are defined in `constants.ts`. You should import the type from `constants.ts` and export it here, so that all the types are still accessible in `types.ts`.

server/src/model/deliverable/validateDeliverableInputs.ts
TARehman: I agree with Connor here mostly for the ease of testing and mocking.

server/src/model/migrations/20260527133836_add_demo_type_deliverable_validation/migration.sql
TARehman, line 1: My fault because I gave you the wrong advice, but this should actually go in `sql/functions.sql`, and it doesn't need to be a migration at all.

server/src/model/migrations/20260527133836_add_demo_type_deliverable_validation/migration.sql
```diff
@@ -0,0 +1,67 @@
...
+    END IF;
+    IF TG_OP = 'DELETE' THEN
+        RETURN OLD;
+    ELSE
+        RETURN NEW;
+    END IF;
+END;
+$$;
+
+DROP TRIGGER IF EXISTS validate_required_demo_types_on_ddt
```
TARehman: When you move to `sql/functions.sql` this won't be necessary because we'll have dropped everything as part of what runs at the start of that file.

server/src/model/migrations/20260527133836_add_demo_type_deliverable_validation/migration.sql
```diff
@@ -0,0 +1,67 @@
...
+ON demos_app.deliverable_demonstration_type;
+
+CREATE CONSTRAINT TRIGGER validate_required_demo_types_on_ddt
```
TARehman: We should make sure the trigger names match what is in the diagram...but I don't know if we even have them named on the diagram? I can fold changes to the diagram into my current PR.

server/src/model/migrations/20260527133836_add_demo_type_deliverable_validation/migration.sql
```diff
@@ -0,0 +1,67 @@
...
+AFTER INSERT OR UPDATE OR DELETE
```
TARehman: I don't think we want this to run after insert here because we're only concerned with the deletion of stuff in this table.

server/src/model/migrations/20260527133836_add_demo_type_deliverable_validation/migration.sql
```diff
@@ -0,0 +1,67 @@
...
+RETURNS TRIGGER
+LANGUAGE plpgsql
+AS $$
+DECLARE
+    target_deliverable_id UUID;
+    deliverable_type_id TEXT;
+    remaining_count INT;
+BEGIN
+    IF TG_TABLE_NAME = 'deliverable_demonstration_type' THEN
+        IF TG_OP = 'DELETE' THEN
```
TARehman: I think you can use the same logic pattern here as you do below; it's a little confusing that the if statements aren't aligned.

server/src/model/migrations/20260527133836_add_demo_type_deliverable_validation/migration.sql
```diff
@@ -0,0 +1,67 @@
...
+            target_deliverable_id := OLD.deliverable_id;
+        ELSIF TG_OP IN ('INSERT', 'UPDATE') THEN
+            target_deliverable_id := NEW.deliverable_id;
+        ELSE
```
TARehman: I don't think you want or need this and it kinda doesn't make sense; if you aren't careful your trigger could just block a specific kind of operation entirely.

server/src/model/migrations/20260527133836_add_demo_type_deliverable_validation/migration.sql
```diff
@@ -0,0 +1,67 @@
...
+            RAISE EXCEPTION 'Unsupported operation on deliverable_demonstration_type';
+        END IF;
+    ELSIF TG_TABLE_NAME = 'deliverable' THEN
+        target_deliverable_id := NEW.id;
+        IF TG_OP = 'DELETE' THEN
+            target_deliverable_id := OLD.id;
+        END IF;
+    ELSE
+        RAISE EXCEPTION 'Invalid trigger table: %', TG_TABLE_NAME;
```
TARehman: Recommend that you use the same language here that we have in check_demonstration_type_exists_for_approved_demonstrations, just for consistency. I think you could look that over and use it to scope down some of the logic in here. We're really only interested in a few cases:
 - If you're adding a new deliverable, it has to have a demo type conditionally.
 - If you change a deliverable to a different type, it might need to have a demo type conditionally.
 - If you're deleting demo types, you can't delete the last one for deliverables of a specific kind.

So you need to run on INSERT OR UPDATE for deliverable, and on DELETE for deliverable_demonstration_type. There's no need to worry about deleting from deliverable and leaving orphaned deliverable_demonstration_types as the FK will enforce that (and indeed, the FK will make it so you have to insert to `deliverable` first and then `deliverable_demonstration_type`; putting that in a transaction allows the trigger to run deferred.

TARehman (changes requested):
There's some small stuff but I do think we should fix up the export of the types and move the SQL code over to the right place (again, my fault, I gave you the wrong advice in Slack without thinking 😅).

TARehman (approved):
LGTM, I made a few small tweaks for types in one small part but nothing major, and updated a test case.

server/src/sql/functions.sql
```diff
@@ -1381,3 +1381,63 @@ BEGIN
...
+        WHERE id = OLD.deliverable_id
+        INTO deliverable_type_id_value;
+        IF deliverable_type_id_value IN ('Implementation Plan', 'Monitoring Protocol') THEN
+            SELECT COUNT(*)
+            INTO demonstration_types_count
+            FROM demos_app.deliverable_demonstration_type
+            WHERE deliverable_id = OLD.deliverable_id;
+            IF demonstration_types_count = 0 THEN
+                RAISE EXCEPTION
+                    'Cannot delete the final demonstration type from deliverable %',
```
connor-parke-icf, line 1405: NB: maybe specify that its because of the deliverable type in this error

server/src/model/deliverable/validateDeliverableInputs.ts
```diff
@@ -78,16 +79,12 @@ export async function validateCreateDeliverableInput(
   errors.push(
     checkDemonstrationStatus(demonstration),
     checkOwnerPersonType(cmsOwnerUser),
-    checkDueDateInFuture(input.dueDate)
+    checkDueDateInFuture(input.dueDate),
+    checkRequiredDeliverableDemonstrationTypes(
+      input.deliverableType,
+      input.demonstrationTypes
+    )
```
connor-parke-icf, line 86: nice, good to have all the validation under the same structure here

connor-parke-icf (approved):

server/src/types.ts
connor-parke-icf: oh hmm... might be thinking of a different thing, but this was moved because we needed to define a subset of the deliverable types. i coulda sworn we did that for a different const at some point in the past

server/src/sql/functions.sql
TARehman: fine to fix in a future PR

PR #1310, opened 2026-05-27
DEMOS-2139: updated state document upload to validate demonstration a…

connor-parke-icf:
found during testing that the auth directives were blocking off too much of the documentPendingUpload object and needed to be relaxed. This is because, while most of the time CMS users are the only ones that can access it, the state users are able to upload and thus need to be able to see at least the presigned upload url. Because this is the one place where a state user has access to the object, only one check is needed in the mutator. It just checks the role and if its a state user, validates they belong to the demonstration theyre uploading on.

### DIFF
--- a/server/src/model/documentPendingUpload/documentPendingUploadResolvers.ts
+++ b/server/src/model/documentPendingUpload/documentPendingUploadResolvers.ts
@@ -19,6 +19,7 @@ import { selectUserOrThrow } from "../user/queries";
 import { resolveDeliverable } from "../deliverable";
 import { updateAssociatedPhase } from "./updateAssociatedPhase";
 import { handleUploadDocumentToDeliverable } from "./handleUploadDocumentToDeliverable";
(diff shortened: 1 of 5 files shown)


server/src/model/documentPendingUpload/documentPendingUploadResolvers.ts
```diff
@@ -76,8 +77,12 @@ export const documentPendingUploadResolvers = {
       parent: unknown,
       { input }: { input: UploadDocumentToDeliverableInput },
       context: GraphQLContext
-    ): Promise<PrismaDocumentPendingUpload> =>
-      handleUploadDocumentToDeliverable(input, context.user.id, false),
+    ): Promise<PrismaDocumentPendingUpload> => {
```
connor-parke-icf, line 80: we dont need to worry about this for the other document uploads, since this is the only one that state users have access to. This allows the documentPendingUpload type to relax its directive restrictions and allow returning of a presigned upload url and deliverable, since the only way for a state user to access that object is when they have access to the associated demonstration.  The auth directives block off every other path

dustbuster (approved):
ok

PR #1314, opened 2026-05-27
DEMOS 1705 + 1773

JeffreyM-GA:
### DEMOS-1705
- Adds demonstration ID to the details page and the approval summary details section for amendments and extensions
image

### DEMOS-1773
- Instead of hiding unset values for modification details, values are shown as "-"
- Fixes an issue with modifications where dates and signature level could not be unset
image

### DIFF
--- a/client/src/components/dialog/modification/EditExtensionDialog.tsx
+++ b/client/src/components/dialog/modification/EditExtensionDialog.tsx
@@ -1,7 +1,7 @@
 import React from "react";
 import { Modification, ModificationFormData } from "./ModificationForm";
 import { gql, TypedDocumentNode, useMutation, useQuery } from "@apollo/client";
-import { UpdateExtensionInput } from "demos-server";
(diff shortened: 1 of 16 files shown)


client/src/components/dialog/modification/EditExtensionDialog.tsx
```diff
@@ -49,7 +49,10 @@ export const useUpdateExtension = (extensionId: string, refetchQueries: string[]
     await updateExtension({
       variables: {
         id: extensionId,
-        input,
+        input: {
+          ...input,
+          effectiveDate: input.effectiveDate as DateTimeOrLocalDate | null | undefined,
```
tresdonicf, line 54: I think we might want to handle this `effectiveDate` transformation in the passed in `input` (so that it's already correct at this point) in order to group all the logic for the payload together

tresdonicf (approved):
One comment on maybe moving the payload adjustment for the update left in order to keep the mutation a bit more straightforward but overall LGTM!

dustbuster (approved):

PR #1315, opened 2026-05-27
(2/3) DEMOS-2053: Some QoL improvements + bugfixes

tresdonicf:
The second PR for DEMOS-2053, this one fixes a couple bugs related to message ordering (now newest is at the top) and prevents double comment submissions. Also adds some QoL improvements like using fragments and messaging in the button signalling that the comment is being added.

https://github.com/user-attachments/assets/8694fe85-dd48-47d6-9d3d-906e7f04ebc6


client/src/pages/deliverables/sections/comment_box/CommentBox.tsx
```diff
@@ -49,10 +50,13 @@ export const CommentBox = ({ deliverableId }: { deliverableId: string }) => {
 
   const handleAddComment = async (commentText: string) => {
     try {
+      setIsSubmitting(true);
       await addComment(commentText);
       setCurrentComment("");
     } catch {
       // error already displayed via toast in useComments
```
dustbuster: Can't we just do `try` - `finally`? 
If we don't need to catch then can we skip it?

I realize kinda out of scope. But figured i'd pose the question.

tresdonicf: We could! Also was thinking that I might just move the handling to here, don't think the `useComment` hook should be doing any rendering, even of a toast!

This is one of the things I'll tackle in the next PR, for now just want to get it into a state where it can be QA'd to make sure there are no unexpected things to address

tresdonicf: Moved the toast to here so now we the `.tsx` file is responsible for all the rendering and the hook just `throws` the error to hit this catch

james-becwar-cms (approved):

PR #1320, opened 2026-05-28
Bugfix: fix document.cookie call for local development

tresdonicf:
I introduced a bug in #1319 by calling `document.cookie` only once it doesn't set both the `id_token` and `access_token` which messed up local dev, this PR fixes it while maintaining only the single `document.cookie` call to show a single issue under snyk.

Also added a unit test to avoid regression here

image

### DIFF
--- a/client/src/router/DemosApolloProvider.tsx
+++ b/client/src/router/DemosApolloProvider.tsx
@@ -14,7 +14,7 @@ import { useAuth } from "react-oidc-context";
 
 const GRAPHQL_ENDPOINT = import.meta.env.VITE_API_URL_PREFIX ?? "/graphql";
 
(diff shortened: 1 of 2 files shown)


client/src/router/DemosApolloProvider.tsx
tresdonicf, line 1: codified this into a function and removed the comment that suggested this is needed for the apollo sandbox

tresdonicf:
I noticed something else here that I thought might be related but it seems like the `RequireRole` component is not playing very well with mock dev as well and you currently can't get to the demonstration pages with `npm run dev:mocks` - at first I thought these issues were related but sadly they are not

james-becwar-cms (approved):

PR #1321, opened 2026-05-28
DEMOS-2157-data-model-for-references

TARehman:
This is the implementation of the data model to back the new reference system. This is entirely the database side; no GQL is involved yet.


data/docs/mermaid-config.json
```diff
@@ -0,0 +1,3 @@
+{
+    "maxTextSize": 200000
```
james-becwar-cms, line 2: lol its a real project now.

james-becwar-cms (approved):

data/docs/DEMOS_Data_Model.mmd
```diff
@@ -459,11 +481,11 @@ erDiagram
...
+    text action_type_id FK, UK "FK114: (action_type_id, old_status_id, new_status_id) ∈ deliverable_action_configuration.(action_type_id, old_status_id, new_status_id)<br>FK136: (action_type_id, due_date_change_allowed, should_have_note, should_have_user_id, extension_id_optional)) ∈ deliverable_action_type.(id, due_date_change_allowed, should_have_note, should_have_user_id, extension_id_optional)<br>UK: (id, action_type_id)"
     text old_status_id FK "FK114: (action_type_id, old_status_id, new_status_id) ∈ deliverable_action_configuration.(action_type_id, old_status_id, new_status_id)"
     text new_status_id FK "FK114: (action_type_id, old_status_id, new_status_id) ∈ deliverable_action_configuration.(action_type_id, old_status_id, new_status_id)"
     text note "NULLABLE"
-    uuid active_extension_id FK "FK128: (active_extension_id, deliverable_id) ∈ deliverable_extension.(id, deliverable_id)<br>NULLABLE"
+    uuid active_extension_id FK, UK "FK128: (active_extension_id, deliverable_id) ∈ deliverable_extension.(id, deliverable_id)<br>NULLABLE"
```
connor-parke-icf: im not seeing this UK reflected in the current prisma, and theres not a blurb description of it, is this UK intentional?

connor-parke-icf (approved):
most commentary i had i just asked you about directly.  A lot of parts here, but it appears to slot in place appropriately. Nice work.  As as aside, we should be careful about how we surface things like the userId on the materials and the agreements with respect to permissions, since these are documents that everyone has access to. I suspect we dont need to surface it, or can block it off with a field permission though. LGTM

PR #1322, opened 2026-05-28
DEMOS-2174: Only show non-deliverable document types in the main `Add Document +` dialog

tresdonicf:
Per DEMOS-2174 only shows the applicable document types when the user clicks `Add Document +` for demonstrations and modifications (amendments + extensions)

https://github.com/user-attachments/assets/fdecb584-005a-4dda-85b6-15d94599dfbb

### DIFF
--- a/client/AGENTS.md
+++ b/client/AGENTS.md
@@ -57,14 +57,21 @@ This file provides instructions for AI agents to use when generating or editing 
 
 ## Testing
 
+### Unit Testing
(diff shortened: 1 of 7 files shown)


james-becwar-cms (approved):
Thanks for making this happen so quick!

tresdonicf:
> Thanks for making this happen so quick!

Anytime!

PR #1325, opened 2026-05-29
DEMOS-2175: Snyk code pipeline

cms-jesse:
## Overview

This PR adds a new GitHub Actions pipeline that manages Jira tickets for issues that are found by Snyk's Code Analysis. (This is strictly for the Code Analysis findings, not dependency vulnerabilities) The goal was for the script to be dependency-free and to keep things as simple as possible.

Since this is a non-critical pipeline that requires no access to private AWS resources, and it uses similar secrets to what is already available for the Security Sync pipeline, so I set this up as a GitHub Action. 

## Details
The script performs the following actions:
1. Query Snyk for all open code findings. Findings that have been ignored are not included
2. Query Jira for all existing Snyk Code Finding tickets
3. Compare results from previous queries. 
  - If a ticket exists, but does not have a matching Synk finding, that ticket will be closed
  - If a finding exists, but no matching ticket, a ticket will be created
  - If there is a matching ticket and finding, nothing is done

Priority and Due Date are set based on Snyk's severity rating. The parent Epic is set within an Action variable in the repo's settings.

To track findings, a specific label and an External ID are set on the ticket. Modifying these values would break the pipeline's ability to align findings and tickets and would potentially create duplicate tickets. A note is added to the ticket description that modifying these should be avoided. Tickets should not be marked as Done manually. If the ticket is marked as Done prior to the finding being resolved in Snyk, the pipeline will open a new ticket for the finding. To avoid this, statuses should be left in any non-resolved value (In Progress, In Review, etc.) and the pipeline will take care of marking it as done.

## Notes
- The pipeline will run every 6 hours or can be triggered manually
- To avoid writing an additional pipeline for the pipeline code, unit tests are just run before executing the script within the same pipeline. 

Here's an example of the tickets that get created: DEMOS-2194

### DIFF
--- a/pipelines/sync-snyk/sync-snyk-code-to-jira.mts
+++ b/pipelines/sync-snyk/sync-snyk-code-to-jira.mts
@@ -79,7 +79,7 @@ interface SnykIssue {
 const jiraLabel = "snyk-sync";
 const jiraBaseUrl = "https://jiraent.cms.gov/rest/api/2";
 const jiraProjectKey = "DEMOS";
-const jiraDoneTransitionId = 41; // 41 is "Done". The full list of options can be retrieved with a GET to the same endpoint.
+const jiraDoneTransitionId = 41; // 41 is "Done". The full list of options can be retrieved with a GET to the /issue/${issue.key}/transitions endpoint.
 const requiredEnvs = ["SNYK_TOKEN", "SNYK_ORG_ID", "JIRA_TOKEN", "JIRA_EPIC"];
 const snykApiVersion = "2025-11-05";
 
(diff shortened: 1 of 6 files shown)


pipelines/sync-snyk/sync-snyk-code-to-jira.mts
```diff
@@ -0,0 +1,312 @@
...
+      .map((remedy) => remedy.description)
+      .join("\n\n")
+      .trim() || "No suggestions provided";
+
+  // Outline for Jira wiki-style markup body.
+  // Expand sections as needed (links, remediation, file paths, CWE, etc.).
+  const descriptionLines = [
+    "----",
+    "*This issue was generated automatically based on findings in Snyk Code*",
+    "Do not mark this ticket as 'Done' manually. The automation pipeline will close it once the finding is resolved or ignored in Snyk.",
```
james-becwar-cms, line 226: Fancy

james-becwar-cms (approved):

PR #1347, opened 2026-06-01
DEMOS-1283 — Poll BN validation status after upload + display the result

nunesmlucas:
After a user uploads a BN Workbook, the client should poll for the validation
status and show success/failure with error messages.

DEMOS-1280 landed the lambda + DB row, but the GraphQL surface for the client
to read the status was missing. James confirmed it could be attached to the
existing `Document` type rather than a new top-level query, so that's what
this PR does.

https://github.com/user-attachments/assets/6adfc03e-3e96-42ff-b41a-7584b4d1c015


server/src/model/document/documentResolvers.ts
```diff
@@ -96,6 +105,24 @@ export async function resolveHasPendingUIPathResult(parent: PrismaDocument): Pro
...
+): Promise<BudgetNeutralityValidationResult | null> {
+  try {
+    const row = await prisma().budgetNeutralityWorkbook.findUnique({
+      where: { id: parent.id },
+      select: { validationStatusId: true, validationData: true },
+    });
+    if (!row) return null;
+    return {
+      status: row.validationStatusId as BudgetNeutralityValidationStatus,
+      errors: (row.validationData as unknown as BudgetNeutralityValidationError[]) ?? [],
```
connor-parke-icf, line 119: im a _little_ nervous about returning this entire json object in the api. We may want to specify the fields that can be returned. but its hard to say, i dont think we've worked with json in gql yet so im not sure how it might behave.

connor-parke-icf: also nit: i dont think we need to default to [], since that column is immutable in the database, it should at least be the empty object. again, not entirely sure how an empty object interacts with GQL, but the cast to unknown here makes me think we may need either more explicit field selection or that we should make this just return an "Object" and let the callsite deal with it.

server/src/model/document/documentResolvers.ts
```diff
@@ -96,6 +105,24 @@ export async function resolveHasPendingUIPathResult(parent: PrismaDocument): Pro
   }
 }
 
+export async function resolveBudgetNeutralityValidation(
```
connor-parke-icf, line 108: because we've defined the thing that this resolves into as a Type in GQL, we could potentially create a resolver for BudgetNeutralityValidationResult instead of keeping it all in a document field.  That way we can export the query functions into their own place and keep the resolver layer thin.  not sure if more or less ergonomic either way, the arbitrary json data is new to me

client/src/components/dialog/document/useBNValidationStatus.ts
```diff
@@ -0,0 +1,51 @@
...
+import gql from "graphql-tag";
+
+import { BudgetNeutralityValidationResult } from "demos-server";
+
+export const BN_VALIDATION_MAX_ATTEMPTS = 30;
+export const BN_VALIDATION_POLL_INTERVAL_MS = 2_000;
+
+export const BN_VALIDATION_STATUS_QUERY = gql`
+  query BnValidationStatus($documentId: ID!) {
+    document(id: $documentId) {
```
connor-parke-icf, line 11: hmmm... this may pose an issue with the current permissions, since the document query is blocked off to State Users. That being said, i think we can open it up to state users without issue since we are already the documents in the resolver... We would want to remove the "@auth" tag from the document query in the schema. 

```
  type Query {
    document(id: ID!): Document!
    documentExists(documentId: ID!): Boolean!
  }
```
(similar is happening to the documentExists query in one of Dustin's PRs)

client/src/components/dialog/document/useBNValidationStatus.ts
```diff
@@ -0,0 +1,51 @@
...
+    }
+  }
+`;
+
+const TERMINAL_STATUSES = new Set(["Succeeded", "Failed"]);
+
+export const useBNValidationStatus = () => {
+  const [getValidationStatus] = useLazyQuery<{
+    document: { id: string; budgetNeutralityValidation: BudgetNeutralityValidationResult | null };
+  }>(BN_VALIDATION_STATUS_QUERY, { fetchPolicy: "network-only" });
```
connor-parke-icf, line 29: for what its worth, ApolloClient has a built-in polling option that should take care of the majority of this polling logic automatically.  not sure if worth changing, since we do it this way in other locations though (and i havnt myself looked deep enough into it).

connor-parke-icf (approved):
i dont think there is anything blocking in here, and theres a few parts of this that im not familiar enough with the behavior to really know what other solutions there might be. This is the first time i think we are looking to return arbitrary JSON to the frontend, so im not sure the best way to handle the casting and validation of those fields. i worry that the validation data shape might change and we get some odd behaviors, but that may be a concern for Post-MVP. 

i think handling the resolution of the fields all within the documentResolvers is fine, since theres no other place we're using the BN upload results.  I do think we might need to adjust the permissions, since currently State Users are unable to access the document query (they can only get the documents from the parent Deliverable, Application, etc. 

Theres a couple of organizational nitpicks, but i dont think anything that cant wait until a later ticket/pos-mvp (for instance, id love the main prisma querying functions to be moved out of the resolver itself and into the queries folder).  We may be able to get away with flattening the result so that we dont have both a parent object and a child object within the same "validation result" type, but i think it makes sense to relegate the json object to its own type.

PR #1348, opened 2026-06-01
Fix: Race Condition in `UserContext`

tresdonicf:
I noticed that `npm run dev:mocks` was being blocked by the `<RequireRole>` component, no matter what user type I used I would get redirected to `/` when going to `demonstrations/...`

Came to realize that this is a race condition that technically affects our real auth flow as well (though less likely to hit). another benefit of using a mocked server, seeing how the system behaves when removing latency!

## Broken
https://github.com/user-attachments/assets/f75eea4c-b7af-4acb-afd1-bcd010d5ea02

## Fixed
https://github.com/user-attachments/assets/defabcfc-c888-45d9-9125-dd8f212ccc17

### DIFF
--- a/client/src/components/user/UserContext.tsx
+++ b/client/src/components/user/UserContext.tsx
@@ -38,7 +38,7 @@ export function UserProvider({ children }: { children: React.ReactNode }) {
     skip: !auth.isAuthenticated,
   });
 
-  if (loading) {
+  if (auth.isLoading || loading) {
     return (
       <div className="flex items-center justify-center h-screen w-screen">
         <Loading />
--- a/client/src/router/DemosRouter.tsx
+++ b/client/src/router/DemosRouter.tsx
@@ -16,8 +16,9 @@ import { ReportsPage } from "pages/ReportsPage";
 import { DeliverableDetailsManagementPage } from "pages/deliverables/DeliverableDetailsManagementPage";
 import { AdminPage } from "pages/admin/AdminPage";
 import { RequireRole } from "./RequireRole";
+import { PersonType } from "demos-server";
 
-const DEMONSTRATION_ACCESS_ROLES = ["demos-admin", "demos-cms-user"];
+const DEMONSTRATION_ACCESS_ROLES: PersonType[] = ["demos-admin", "demos-cms-user"];
 
 const HomePage = () => {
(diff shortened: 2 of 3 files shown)


client/src/router/RequireRole.tsx
```diff
@@ -1,16 +1,14 @@
 import React from "react";
 import { getCurrentUser } from "components/user/UserContext";
 import { Navigate } from "react-router-dom";
+import { PersonType } from "demos-server";
 
 type RequireRoleProps = {
-  allowedRoles: string[];
+  allowedRoles: PersonType[];
```
tresdonicf, line 7: Tightened up this type a little bit to make sure only a list of valid `PersonType`s can be passed in here

client/src/components/user/UserContext.tsx
```diff
@@ -38,7 +38,7 @@ export function UserProvider({ children }: { children: React.ReactNode }) {
     skip: !auth.isAuthenticated,
   });
 
-  if (loading) {
+  if (auth.isLoading || loading) {
```
tresdonicf, line 41: This was essentially just a race condition, since the mock currentUser query will return immediately this would render the children immediately and thus render `<RequireRole>` while `currentUser` was undefined.

This race-condition technically exists in real cognito as well though less likely as the OIDC transaction takes some actual time

client/src/router/DemosRouter.tsx
```diff
@@ -40,32 +41,37 @@ export const DemosRouter: React.FC = () => {
               <Route element={<DemosLayoutProvider />}>
                 <Route path="*" element={<div>404: Page Not Found</div>} />
                 <Route path="/" element={<HomePage />} />
-                <Route path="demonstrations"
+                <Route
+                  path="demonstrations"
```
dustbuster, line 45: NB; I did not set the format like this cuz it makes the page less "read like a route list" But no big deal.

dustbuster (approved):
Looks good!

client/src/router/DemosRouter.tsx
tresdonicf: Valid! I think this was just `prettier` doing its thing when I saved the `PersonType` changes hahah

PR #1350, opened 2026-06-01
Demos cms files fix - missing attributes on trigger for deliverable files

dustbuster:
Got CMS files upload working

Screenshot 2026-06-01 at 14 04 57

https://github.com/user-attachments/assets/7856489d-0168-4a6c-a50e-bfa3c0b13aac


nunesmlucas (commented):
LGTM - THANK YOU!

nunesmlucas (approved):
LGTM - THANK YOU!

server/src/sql/functions.sql
```diff
@@ -252,6 +252,8 @@ BEGIN
         application_id,
         phase_id,
         deliverable_id,
+        deliverable_type_id,
+        deliverable_is_cms_attached_file,
```
james-becwar-cms, line 256: Aren't these already on deliverable?  Could we just join to deliverable?

james-becwar-cms (changes requested):

server/src/sql/functions.sql
james-becwar-cms: nevermind I'm an idiot.

james-becwar-cms (approved):

server/src/sql/functions.test.ts
connor-parke-icf, line 1: i wouldnt include this test file, since theres a lot more going on here that needs to be validated via integration testing that this wouldnt cover. This is only checking via regex/stringmatch that the functions have what we added in the main function (its essentially testing only a configuration, not a behavior), so i worry itd give us a false sense of security.

connor-parke-icf (commented):
main fix looks exactly what is needed. I had noticed this but hadnt gotten around to resolving it yet; the lambda is likely throwing a sql error. I had completely forgotten about the infected file as well, so good catch on resolving that one as well, adding the deliverableId in addition to the deliverableTypeId and deliverableIsCmsAttachedFile.

I would remove the test file though, since its not really testing anything other than the string contents of functions.sql, not the behavior of the functions. That would need to be tested with a deeper suite, so we have largely relegated to manual testing for these.

server/src/sql/functions.sql
dustbuster: ME TOO!

PR #1354, opened 2026-06-01
DEMOS-1709: fixed state user permissions for documentExists

dustbuster:
Nuked the auth attribute for documentExists per convo with @connor-parke-icf  about permissions

Screenshot 2026-06-01 at 15 59 28

Also edit and delete works. 
Screenshot 2026-06-01 at 16 04 38


connor-parke-icf (approved):
discussed this directly. Both state users and cms users have valid reasons to query documents directly, and because the document fetching layer applies the appropriate filters to the document fetch results, i think its safe to open these queries up.

PR #1355, opened 2026-06-02
DEMOS-2109 add signature level constraint

JeffreyM-GA:
Adds a constraint that enforces valid signature levels for applications.
image

### DIFF
--- a/server/src/model/migrations/20260602115947_check_signature_level/migration.sql
+++ b/server/src/model/migrations/20260602115947_check_signature_level/migration.sql
@@ -1,11 +1,20 @@
 ALTER TABLE demos_app.demonstration
 ADD CONSTRAINT demonstration_signature_level_check
-CHECK (signature_level_id = 'OA');
+CHECK (
+  signature_level_id IS NOT NULL
+  AND signature_level_id = 'OA'
+);
 
 ALTER TABLE demos_app.amendment
 ADD CONSTRAINT amendment_signature_level_check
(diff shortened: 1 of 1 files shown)


server/src/model/migrations/20260602115947_check_signature_level/migration.sql
```diff
@@ -0,0 +1,11 @@
+ALTER TABLE demos_app.demonstration
+ADD CONSTRAINT demonstration_signature_level_check
+CHECK (signature_level_id = 'OA');
+
+ALTER TABLE demos_app.amendment
+ADD CONSTRAINT amendment_signature_level_check
+CHECK (signature_level_id IN ('OA', 'OCD'));
```
connor-parke-icf: hmm... this implies that Null is not a valid option.  If we are truly trying to enforce that an option must exist for these fields, we should probably adjust the Demonstration, Amendment, and Extension .prisma files to remove the question mark (and update the mermaid diagram (and maybe... move this check to be handled by a limit table)). But IIRC that determination was still up in the air. I think what we are really intending is "if there is a value, it must be one of these values", so we should allow null as a valid value OR one of the listed values

connor-parke-icf (dismissed):
if we want null to be a valid value, which should allow it in these check constraints. If we dont want null to be a valid value, we should probably adjust the table definition itself and associated documentation to show that.  i _think_ we are looking to just allow null for now though

connor-parke-icf (approved):

PR #1360, opened 2026-06-02
DEMOS-2205: Cloudwatch alarms and alerting

cms-jesse:
## Overview

We have alerts set up in splunk based on application warnings and errors, but we don't use splunk for infrastructure-related errors. This PR adds several cloudwatch alarms for monitoring the infrastructure, for example database memory consumption, cloudfront error anomalies, etc. Potentially, some of these alerts could be created in Datadog, but they would be delayed based on the ingestion timing. Datadog will be used for higher level alerts and dashboarding, these fine-grained alarms will be left in cloudwatch

### DIFF
--- a/deployment/lib/lambda.ts
+++ b/deployment/lib/lambda.ts
@@ -126,7 +126,7 @@ export class Lambda extends Construct {
         sourceMap: true,
         externalModules: props.externalModules,
         nodeModules: props.nodeModules,
-        logLevel: LogLevel.VERBOSE,
+        logLevel: LogLevel.ERROR,
         commandHooks: props.commandHooks,
         format: props.format,
         esbuildArgs: props.esbuildArgs,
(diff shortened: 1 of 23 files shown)


deployment/lib/lambda.ts
```diff
@@ -126,7 +126,7 @@ export class Lambda extends Construct {
         sourceMap: true,
         externalModules: props.externalModules,
         nodeModules: props.nodeModules,
-        logLevel: LogLevel.VERBOSE,
+        logLevel: LogLevel.ERROR,
```
james-becwar-cms, line 129: Will we still have see the warn messages in the logs?

james-becwar-cms (approved):

deployment/lib/lambda.ts
cms-jesse: This is the log level of the esbuild bundling process that happens during a `cdk deploy ...` command, not the log level of the function itself at runtime

PR #1361, opened 2026-06-02
Demos 1709: removed duplicate submit button, set up state view deliverable buttons

dustbuster:
Screenshot 2026-06-02 at 10 25 14

### DIFF
--- a/client/src/pages/DemonstrationDetail/DemonstrationDetailHeader.tsx
+++ b/client/src/pages/DemonstrationDetail/DemonstrationDetailHeader.tsx
@@ -3,11 +3,12 @@ import React, { useCallback, useState } from "react";
 import { CircleButton } from "components/button";
 import { BaseButton } from "components/button/BaseButton";
 import { AddNewIcon, ChevronLeftIcon, EditIcon, EllipsisIcon } from "components/icons";
-import { Demonstration, Person, State } from "demos-server";
+import { Demonstration, Person, PersonType, State } from "demos-server";
 import { formatDate } from "util/formatDate";
 import { gql, useQuery } from "@apollo/client";
 import { useDialog } from "components/dialog/DialogContext";
 import { useNavigate } from "react-router-dom";
+import { getCurrentUser } from "components/user/UserContext";
 
 export const DEMONSTRATION_HEADER_DETAILS_QUERY = gql`
   query DemonstrationHeaderDetails($id: ID!) {
@@ -42,12 +43,18 @@ interface DemonstrationDetailHeaderProps {
   demonstrationId: string;
 }
 
+const HEADER_ACTION_PERSON_TYPES: ReadonlySet<PersonType> = new Set([
+  "demos-admin",
+  "demos-cms-user",
+]);
+
 export const DemonstrationDetailHeader: React.FC<DemonstrationDetailHeaderProps> = ({
   demonstrationId,
 }) => {
   const [showButtons, setShowButtons] = useState(false);
   const [showDropdown, setShowDropdown] = useState(false);
   const navigate = useNavigate();
+  const { currentUser } = getCurrentUser();
 
   // Ensure this component is rendered inside a DialogProvider in your app;
   const { showEditDemonstrationDialog, showCreateAmendmentDialog, showCreateExtensionDialog } =
@@ -96,6 +103,8 @@ export const DemonstrationDetailHeader: React.FC<DemonstrationDetailHeaderProps>
       value: demonstration.expirationDate ? formatDate(demonstration.expirationDate) : "--/--/----",
     },
   ];
+  const canManageDemonstration =
+    !!currentUser && HEADER_ACTION_PERSON_TYPES.has(currentUser.person.personType);
 
   return (
     <div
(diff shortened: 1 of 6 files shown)


client/src/pages/deliverables/sections/FileAndHistoryTabs.tsx
```diff
@@ -103,6 +108,8 @@ export const FileAndHistoryTabs: React.FC<{
   const userPersonType = currentUser?.person.personType;
   const isCompleteReviewAllowedForUser =
     !!userPersonType && COMPLETE_REVIEW_PERSON_TYPES.has(userPersonType);
+  const canRequestResubmission =
+    !!userPersonType && REQUEST_RESUBMISSION_PERSON_TYPES.has(userPersonType);
```
dustbuster: Handles SU view of buttons.

client/src/pages/deliverables/sections/FileAndHistoryTabs.test.tsx
```diff
@@ -296,7 +286,7 @@ describe("FileAndHistoryTabs", () => {
         setup({ status });
 
         expect(screen.getByTestId(STATE_FILES_ADD_BUTTON_NAME)).toBeDisabled();
-        expect(screen.getByTestId(STATE_FILES_SUBMIT_BUTTON_NAME)).toBeDisabled();
+        expect(screen.getByTestId("button-actions-submit-deliverable")).toBeDisabled();
```
tresdonicf: NB: Looks like this button has changed but if possible it's ideal to use the exported `..._BUTTON_NAME` value here so that if that name happens to change the test doesn't need to be updated

client/src/pages/deliverables/sections/FileAndHistoryTabs.tsx
tresdonicf: NB: Just curious, is there ever a case where we want to continue here if there's not a person type?

client/src/pages/DemonstrationDetail/DemonstrationDetailHeader.tsx
```diff
@@ -165,73 +174,75 @@ export const DemonstrationDetailHeader: React.FC<DemonstrationDetailHeaderProps>
...
-                  className="w-full text-left px-1 py-[10px] hover:bg-gray-100"
+                  <EditIcon />
+                </CircleButton>
+                <CircleButton
+                  name="Create New"
+                  data-testid="create-new-button"
+                  size="small"
+                  onClick={() => setShowDropdown((prev) => !prev)}
+                  tooltip="Create New"
                 >
```
tresdonicf: Not necessary here but at this sort of depth the pattern of conditional rendering becomes a bit hard to read / reason about. A slightly better approach might be a component that encapsulates the logic of the component, and handles details internally so they're not all laid out in the top-level component. For instance here maybe this component would render:
```
<>
<DemonstrationManagementButtons personType={user.personType} />  // Decides to render or not based on person type
{showDropdown && <CreateNewDropdown demonstrationId={}/>
<OverflowButton ... />  <-- This one's a bit overkill
</>
```

But even if we do the calculation ahead of time we get a nice improvement of:
`{canManageDemonstration  && <DemonstrationManagementButtons />}`

I know it's a bit more complex than this because of the overflow button needing to be clicked but just food for thought going forward!

client/src/pages/DemonstrationDetail/DemonstrationDetailHeader.tsx
```diff
@@ -165,73 +174,75 @@ export const DemonstrationDetailHeader: React.FC<DemonstrationDetailHeaderProps>
...
-                tooltip="Create New"
-              >
-                <AddNewIcon />
-              </CircleButton>
-            </span>
-            {showDropdown && (
-              <div className="absolute w-[160px] bg-white text-black rounded-[6px] shadow-lg border z-20">
-                <button
-                  data-testid="button-create-new-amendment"
+      {canManageDemonstration ? (
```
tresdonicf: If we're going to render ` : null` on the other side of the ternary then we could leave that out altogether and do `{canManageDemonstration && <Components />}`

tresdonicf (approved):
A couple food for thought comments but overall looks good!

client/src/pages/deliverables/sections/FileAndHistoryTabs.test.tsx
dustbuster: Good catch. There was alos a version of this where i centralize all the user types that have access. Cms/Admin. but it got real big. So i nixed that. But maybe something to consider once we get to done point.

client/src/pages/deliverables/sections/FileAndHistoryTabs.tsx
dustbuster: Um. no. I don't think so. This would be locked down on API level. We can nuke all these.

client/src/pages/DemonstrationDetail/DemonstrationDetailHeader.tsx
dustbuster: For sure! I think this is not a bad pattern at all.

client/src/pages/DemonstrationDetail/DemonstrationDetailHeader.tsx
dustbuster: Donzo!

dustbuster:
ALL SUGGESTIONS IMPLEMENTED! Minus the pattern change to encapsulate the logic inside the component. Not that I'm opposed just need to get another ticket done!

PR #1363, opened 2026-06-02
DEMOS-2162-add-reference-api-basic-queries

TARehman:
This PR primarily implements the basic resolvers necessary for the reference system. However, as a secondary aspect, it also enhances and modifies how logging works on the server, allowing for more customization and splitting of what is presented to the user versus to our internal logs.

# Data Model

- Added the `owner_person_type_id` to the `reference` and `reference_agreement` tables so ensure we can limit who is allowed to own those objects. Updated history triggers accordingly.
- Added `acceptance_timestamp` to the PK for `reference_agreement_acceptance` so the rows are properly unique; otherwise, you can't accept more than one time, which is not the desired behavior.
- Fixed a typo with a double `))` in the ER diagram.

# Logging System

Added a new system for error codes found in `error/errorCodes.ts`. The goal here is to make it easier to control the level where things are logged and to also mask some error codes to users of the API while still making them available internally.

- The file facilitates the concept of an internal and external custom error code, and allows mapping them so that we can control what the API reports out. It also maps to the `pino` logging levels of `debug`, `info`, etc, allowing you to define a custom internal error, what level it should log to, and what should be shown to the user when throwing.
- The file also contains a typeguard used elsewhere, as well as a few convenience functions like mapping a public code from a private one.
- The most important function in this is the `formatGraphQLErrorCode` function, which is used in both `local-server.ts` and `server.ts`. This function receives the error passed in (which is already formatted, the nomenclature is a little confusing), and specifically in the case of a customization, applies the public facing code; if for any reason there's no error code or it's not a custom one, it just does nothing.
- Added tests for this file.
- Where the formatter is used in `local-server.ts` and `server.ts`, I also cleaned up the plugin and argument order to match between the two, just for ease of comparing them; there's no changes in logic.
- Worth noting here that this specifically requires and catches a GraphQLError specifically. If you throw a bare Error, this gets wrapped into an `INTERNAL_SERVER_ERROR` in GQL, and currently there's no special handling for that.
- Updated `logging.plugin.ts` to make use of the new customization. This adds a function get a log level from a GraphQLError (`getErrorLogLevel`). This works similarly to the formatter above - if there's no error code, or if its not a custom one, then it just leaves it at warn (where it was before). If there _is_ a custom error, it uses the configured level.
  - I also lowered the log level of the other operations in here (that are not customized) to be `debug` rather than `info`; with `info` enabled, we just get queries every five seconds that are the introspection queries, which makes it really hard to know what's happening in the system.

# Reference System Implementation

Added a `Reference` model exposed by GQL along with accompanying queries and helper functions.

- This is a _tiny_ bit confusing because this is actually linked mostly to the `reference_configuration` table. Specifically, when using the `ID` argument in the GQL `Reference` model, we're going to be looking at `reference_configuration.id`. The issue is that the records in `reference` _are_ the references; the name is correct. But from the API perspective, we are concerned with the combination of a reference and a reference agreement (what is called a reference configuration). We _also_ need a unique synthetic key for `reference_configuration` because not all references have associated agreements, and while PostgreSQL is smart enough to understand this, there's no concept like that in Prisma.
- You can see the new logging pattern exhibited throughout the various reference-related files.
- Note that getting a download URL is technically a query, not a mutator, which @connor-parke-icf and I discussed. Also note that the code to get a reference download URL is wrapped in a transaction (because we have to conditionally log acceptance of the agreement, necessitating two DB operations), but the actual agreement downloads are not.
- After some debate, we decided the best way to handle accepting the agreement is to have the API take an optional `acceptedAgreementId` argument. This means the person calling the API has explicitly said "this is the ID of the agreement that I accepted", which I thought was better than just passing a boolean.
- We also refuse to return an agreement if its not assigned to at least one active configuration, which is sort of a proxy for whether the agreement itself is active.
- Note that some of the queries here are a bit deeper nested and more complex, but the tradeoff is that we only need one. I thought given how the reference system worked, this made sense; I don't think we are going to need to do a lot of abstracting between references and other things.

# Miscellaneous

- Updated the seeder to both add new reference records when seeding, and _also_ to fix the deleteMany / truncate commands to work correctly.
- Added a new command, `npm run dev:quiet`, which sets the log level to `info` rather than `debug`, so as to make the logs more useful locally.


server/src/model/reference/getReferenceDownloadUrl.ts
```diff
@@ -0,0 +1,32 @@
...
+      args.id,
+      tx,
+      args.acceptedAgreementId
+    );
+    if (args.acceptedAgreementId) {
+      // If we enter this block, we know there is a referenceAgreement on the configuration
+      await insertReferenceAgreementAcceptance(
+        {
+          referenceId: validatedReferenceConfiguration.reference.id,
+          referenceAgreementId: validatedReferenceConfiguration.referenceAgreement!.id,
```
connor-parke-icf, line 23: as much as the exclamation point feels wrong, i think this is actually a pretty appropriate use case of it. beats making a complicated generic type with conditionals on when its fields can or cant be null (for now anyway)

server/src/model/reference/referenceResolvers.ts
```diff
@@ -0,0 +1,66 @@
...
+    name: (parent: SelectManyReferenceConfigurationsResult): NonEmptyString => {
+      return parent.reference.name;
+    },
+    description: (parent: SelectManyReferenceConfigurationsResult): NonEmptyString => {
+      return parent.reference.description;
+    },
+    agreement: (parent: SelectManyReferenceConfigurationsResult): ReferenceAgreement | null => {
+      return parent.referenceAgreement;
+    },
+    tags: (parent: SelectManyReferenceConfigurationsResult): Tag[] => {
```
connor-parke-icf, line 47: aside: i had ran into this decision point before: is the parent best represented by a PrismaReferenceConfiguration or the type we have for the select? This works because the only way we're fetching the tag is via that query, but its a nuance to keep track of.  Ideally, we could have one object type that isnt specific to prisma or a certain query result, but rather is just a "ReferenceConfiguration".

server/src/model/reference/validateReferenceDownloadRequest.ts
```diff
@@ -0,0 +1,80 @@
...
+      extensions: {
+        code: "REFERENCE_NOT_ACTIVE" satisfies CustomInternalErrorCode,
+      },
+    });
+  }
+
+  const referenceAgreementId = referenceConfiguration.referenceAgreement?.id;
+  const errorMessage = "Cannot download requested reference using provided agreement ID.";
+
+  if (!referenceAgreementId && agreementId) {
```
connor-parke-icf, line 43: do we want this to be a failure state? not sure which makes more sense; providing an optional param feels like it should be allowed, even if its ignored. But it ultimately just moves the check for whether or not we need to make a agreement acceptance

james-becwar-cms (approved):

server/src/model/referenceAgreement/queries/selectReferenceAgreement.ts
```diff
@@ -0,0 +1,36 @@
...
+          },
+        },
+      },
+    },
+  });
+  if (result) {
+    const { _count, ...rest } = result;
+    return {
+      ...rest,
+      inActiveUse: _count.referenceConfigurations > 0,
```
connor-parke-icf, line 31: im guessing this is meant to limit the downloading of arbitrary or retired agreements that arent actually connected to any references? wondering if we should instead include the referenceId with the agreementId when fetching the agreement file.

server/src/local-server.ts
```diff
@@ -27,16 +28,18 @@ const protection = armor.protect();
 const server = new ApolloServer<GraphQLContext>({
   typeDefs,
   resolvers,
-  introspection: process.env.ALLOW_INTROSPECTION === "true",
   ...protection,
+  introspection: process.env.ALLOW_INTROSPECTION === "true",
```
connor-parke-icf, line 32: suppose it makes sense to put these after the spread so they truly override whatever protection puts in place

server/src/seeder.ts
```diff
@@ -590,12 +735,31 @@ async function clearDatabase() {
...
+    prisma().referenceConfiguration.deleteMany(),
+    prisma().referenceAgreement.deleteMany(),
+    prisma().reference.deleteMany(),
+
     prisma().primaryDemonstrationRoleAssignment.deleteMany(),
     prisma().demonstrationRoleAssignment.deleteMany(),
     prisma().applicationDate.deleteMany(),
     prisma().applicationPhase.deleteMany(),
     prisma().applicationNote.deleteMany(),
+    prisma().applicationTagAssignment.deleteMany(),
```
connor-parke-icf, line 751: you werent kidding; there were a lot of additions to the db since the last time the seeder without reset was worked with

connor-parke-icf (approved):
LGTM, dont think i saw any blockers.  General comments and asides, but not many real actionables.  Going to meet directly over the actual database/prisma changes

server/src/model/reference/getReferenceDownloadUrl.ts
TARehman: I think with another dozen lines of code I could convince TS that this is fine (it would involve different derived classes and stuff) but this felt just simpler.

server/src/model/reference/referenceResolvers.ts
TARehman: Yeah, this is a real question. In this _particular case_ given how the references aren't really join points to go through to other things, I felt this was okay, but it wouldn't necessarily be a good general pattern.

PR #1365, opened 2026-06-02
DEMOS-1783 Attestation dialog on BN notebook upload

nunesmlucas:
When a user uploads a Budget Neutrality (BN) notebook, an attestation dialog now
appears and must be acknowledged before anything is submitted.

- **Confirm** → upload proceeds as normal.
- **Cancel / dismiss** → upload is cancelled, nothing is submitted.
- Gated on document types `Final Budget Neutrality Formulation Workbook` and `BN Workbook`.

image


dustbuster (approved):
I ATTEST THIS IS OK.

PR #1368, opened 2026-06-03
DEMOS-2208: 508 Compliance: ProfileBlock

tresdonicf:
Per DEMOS-2208 this change updates the `ProfileBlock` component to work better with assistive technologies.

This ticket's a little out of date given there's only one option under the Profile drop down now (`Sign Out`) but updated this with some needed a11y improvements anyhow


client/src/components/header/ProfileBlock.tsx
```diff
@@ -2,13 +2,32 @@ import React, { useState, useRef, useEffect } from "react";
 import { ChevronDownIcon } from "components/icons";
 import { Avatar } from "./Avatar";
 import { getCurrentUser } from "components/user/UserContext";
-import { SignoutLink } from "../auth/AuthLinks";
+import { useAuthActions } from "components/auth/AuthActions";
 
 export const PROFILE_BLOCK_TEST_ID = "profile-block";
 
+export function SignoutLink() {
```
tresdonicf, line 11: Moved this here since it's only used here

client/src/components/header/ProfileBlock.tsx
```diff
@@ -40,21 +66,35 @@ export const ProfileBlock: React.FC = () => {
...
 
       {open && (
-        <ul className="absolute top-12 right-0 min-w-full bg-white border border-gray-300 rounded shadow-lg z-50">
-          <li className="hover:bg-gray-100 cursor-pointer p-1">
+        <ul
+          role="menu"
+          aria-label="User profile options"
+          className="absolute top-12 right-0 min-w-full bg-white border border-gray-300 rounded shadow-lg z-50"
+        >
+          <li role="none" className="hover:bg-gray-100 cursor-pointer p-1">
```
tresdonicf, line 101: Enhanced the `<ul>` here

client/src/components/header/ProfileBlock.tsx
```diff
@@ -40,21 +66,35 @@ export const ProfileBlock: React.FC = () => {
...
-      </span>
+      <button
+        ref={triggerRef}
+        type="button"
+        aria-haspopup="menu"
+        aria-expanded={open}
+        aria-label={`User profile menu for ${resolvedName}`}
+        className="flex items-center gap-x-1 cursor-pointer"
+        onClick={() => setOpen((value) => !value)}
+      >
```
tresdonicf, line 85: The big change, clickable elements should be `<button>` rather than `<div>` - also added aria attributes

client/src/components/header/ProfileBlock.tsx
```diff
@@ -31,6 +53,13 @@ export const ProfileBlock: React.FC = () => {
     };
   }, [open]);
 
+  const handleKeyDown = (event: React.KeyboardEvent) => {
+    if (event.key === "Escape" && open) {
+      setOpen(false);
+      triggerRef.current?.focus();
```
connor-parke-icf, line 59: neat, so when escaping, it goes back to highlighting the profile button

connor-parke-icf (approved):
LGTM! im not familiar enough with accessibility best practices here, but something about having role="none" for the menu items feels a little off, but it may be the best thing to do. I know we used to have a dev-dependency package for 508 checking, but i think we removed it for snyk issues. may be a good time to re-add it; im curious what the various things and recommendations for them might be

client/src/components/header/ProfileBlock.tsx
tresdonicf: Yeah exactly!

tresdonicf:
> LGTM! im not familiar enough with accessibility best practices here, but something about having role="none" for the menu items feels a little off, but it may be the best thing to do. I know we used to have a dev-dependency package for 508 checking, but i think we removed it for snyk issues. may be a good time to re-add it; im curious what the various things and recommendations for them might be

I thought that was strange too! I did some research on it and it seems like it's to avoid a duplicative `role` assignment that comes from the interactive item in a `<ul>` getting a `menuitem` role: https://stackoverflow.com/a/78529847

Yeah we still have the axe scripts around! For this though I'm just testing it using the browser, both Chrome and Firefox have accessibility tabs in the developer tools that'll do a pretty good job of flagging some stuff! Screen reader order is where this stuff might get a little bit more tricky

PR #1369, opened 2026-06-03
DEMOS-2124-generate-medicaid-chip-ids-new-demos

TARehman:
First PR for the Medicaid / CHIP IDs. This adds two columns to the `demonstration` table to contain these values. There is a `BEFORE INSERT` trigger which fires to create each of them at insert time, and a `BEFORE UPDATE` trigger that prevents editing of `state_id`, `medicaid_id`, or `chip_id` after creation.

The migration handles this in a bit of a sequenced fashion so that I could avoid having to do a reset on the DB.


james-becwar-cms (approved):

server/src/model/demonstration/demonstration.prisma
```diff
@@ -11,6 +11,8 @@ model Demonstration {
   currentPhaseId    String    @map("current_phase_id")
   stateId           String    @map("state_id")
   clearanceLevelId  String    @default("CMS (OSORA)") @map("clearance_level_id")
+  medicaidId        String    @map("medicaid_id") @default(dbgenerated())
```
connor-parke-icf, line 14: interesting, havnt seen the dbgenerated tag before. good to know this exists and is supported by prisma

server/src/model/migrations/20260602201004_add_medicaid_chip_id_values/migration.sql
```diff
@@ -0,0 +1,105 @@
...
+    demos_app.medicaid_id_number_seq
+START WITH
+    11000
+INCREMENT BY
+    1
+MINVALUE
+    11000
+MAXVALUE
+    99999
+NO CYCLE;
```
connor-parke-icf, line 25: does the NO CYCLE indicate that it will just error if it hits 100000 rather than looping?

server/src/sql/functions.sql
```diff
@@ -439,7 +439,7 @@ FOR EACH ROW
 EXECUTE FUNCTION demos_app.check_application_type_record_exists();
 
 -- update_application_current_phase_on_phase_update
-CREATE OR REPLACE FUNCTION demos_app.update_application_current_phase_on_phase_update()
```
connor-parke-icf, line 442: hmm, arent we intending for these functions to be idempotent? wouldnt removing the replace here mean that it could fail if the function already exists?

connor-parke-icf (approved):
assuming theres a reason to make the functions no longer create or replace, this seems find to me. The ID generation is easy to follow and everything surrounding it is fairly simple; its just a scalar afterall

server/src/model/demonstration/demonstration.prisma
TARehman: They've hinted at deprecating it, which would be an issue for us...

server/src/model/migrations/20260602201004_add_medicaid_chip_id_values/migration.sql
TARehman: Yes, that's correct. I estimate this to be of low risk for us. 😁

server/src/sql/functions.sql
TARehman: At the top of this file is a loop that already removes all the functions. OR REPLACE is entirely unnecessary here. I should remove it from the history triggers as well given that we do the same looping thing.

server/src/model/demonstration/demonstration.prisma
connor-parke-icf: naturally

server/src/sql/functions.sql
connor-parke-icf: oh nice, i didnt see that we remove it already. then yes, we should remove it here so it doesnt cover up some unknown thing causing the function to survive

TARehman:
I have combined some changes between the other PR I opened which isn't necessary now, and this one. Understanding the sequencing of when the various files in `sql/` run along with the chances where it may or may not run has been interesting. I believe this version will fix the issues but would appreciate a re-review.

james-becwar-cms (approved):

connor-parke-icf (approved):
combines parts of previous PRs which ive already reviewed; discussed about the various changes and why. A lot of sql gymnastics and threading of needles here, but makes sense. Very good to know these things to at least be able to reference ahead of the prod release.

PR #1371, opened 2026-06-03
DEMOS-2124-make-triggers-more-conditional

TARehman:
This does three main cleanup items on the idempotent parts of the DB code.

- History triggers are no longer generated with `CREATE OR REPLACE`. This was a leftover from before we had the loop at the start that deletes them all.
- `check_demonstration_primary_project_officer` no longer runs when we are updating `demonstration`. Nothing in an update should change the primary project officer anyway, so this was unnecessary, and it was causing me some issues with #1369. I think this was here originally because there was a column for that on the table, but now that it's removed, this can just run `AFTER INSERT`.
- Some other triggers that run on update were narrowed using the `OF column` syntax of PostgreSQL to cause them not to trigger unless the relevant columns are edited. Note that no changes were made to any of the actual function code, so there's some redundancy in the guards that were already placed in the functions in some places. We can always go back and streamline those later.
  - A few `CREATE OR REPLACE` calls in `functions.sql` were removed as well. The only remaining one is specifically in `permissions.sql` for a function we create and drop in the same file; putting it that way ensures that the function will still get created if something got aborted halfway through running before cleanup.

### DIFF
--- a/server/src/sql/functions.sql
+++ b/server/src/sql/functions.sql
@@ -121,7 +121,7 @@ END;
 $$;
 
 CREATE CONSTRAINT TRIGGER check_demonstration_primary_project_officer
-AFTER INSERT OR UPDATE ON demos_app.demonstration
+AFTER INSERT ON demos_app.demonstration
 DEFERRABLE INITIALLY DEFERRED
 FOR EACH ROW
 EXECUTE FUNCTION demos_app.check_demonstration_primary_project_officer();
@@ -161,7 +161,7 @@ END;
 $$;
 
 CREATE CONSTRAINT TRIGGER check_demonstration_retains_primary_project_officer
-AFTER UPDATE OR DELETE ON demos_app.primary_demonstration_role_assignment
+AFTER UPDATE OF role_id OR DELETE ON demos_app.primary_demonstration_role_assignment
 DEFERRABLE INITIALLY DEFERRED
 FOR EACH ROW
 EXECUTE FUNCTION demos_app.check_demonstration_retains_primary_project_officer();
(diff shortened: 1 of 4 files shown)


server/src/sql/functions.sql
```diff
@@ -161,7 +161,7 @@ END;
 $$;
 
 CREATE CONSTRAINT TRIGGER check_demonstration_retains_primary_project_officer
-AFTER UPDATE OR DELETE ON demos_app.primary_demonstration_role_assignment
+AFTER UPDATE OF role_id OR DELETE ON demos_app.primary_demonstration_role_assignment
```
connor-parke-icf, line 164: wouldnt we need this to also check demonstrationId? or is that an impossible situation given the other constraints

server/src/sql/functions.sql
```diff
@@ -623,7 +623,7 @@ END;
 $$;
 
 CREATE TRIGGER update_application_status_on_phase_update
-AFTER UPDATE ON demos_app.application_phase
+AFTER UPDATE OF phase_id, phase_status_id ON demos_app.application_phase
```
connor-parke-icf, line 626: can the phase_id of an application_phase ever change? my guess is no, because of other constraints, so this feels redundant. but the redundancy might come from an intersection of separate concerns. if so, should primary_demonstration_role_assignment check for changes on demonstration_id?

connor-parke-icf (approved):
discussed these changes directly, explains the changes in the previous PR where the OR REPLACE was removed from those relevant functions. Im curious what the collision was that prompted this update though. 

couple of questions regarding which columns are selected for the updated at, but assuming those are fine this seems good to me

server/src/sql/functions.sql
TARehman: If so, we need to update the actual function, as it doesn't look at that at all.

server/src/sql/functions.sql
TARehman: I just lifted the variables here from the function being called.

TARehman:
Closing in favor of #1369 which needs to have both fixes in it.

PR #1380, opened 2026-06-04
DEMOS-2207: 508 Compliance - Tabbing through New Demonstration Dialog

tresdonicf:
Per DEMOS-2207 This PR addresses the focus shifting back to the "X" when tabbing through the fields in the Create New Demonstration Dialog.

https://github.com/user-attachments/assets/24cc2227-418e-4939-adf7-96338d9ed2b1

### DIFF
--- a/client/src/components/dialog/BaseDialog.tsx
+++ b/client/src/components/dialog/BaseDialog.tsx
@@ -85,7 +85,12 @@ export const BaseDialog: React.FC<BaseDialogProps> = ({
       >
         {!hideHeader && (
           <>
-            <button onClick={onCloseClicked} className={CLOSE_BUTTON} aria-label="Close dialog">
+            <button
+              type="button"
+              onClick={onCloseClicked}
+              className={CLOSE_BUTTON}
+              aria-label="Close dialog"
+            >
               ×
             </button>
             <h2 className={TITLE}>{title}</h2>
--- a/client/src/components/input/select/AutoCompleteSelect.tsx
+++ b/client/src/components/input/select/AutoCompleteSelect.tsx
@@ -72,6 +72,12 @@ export const AutoCompleteSelect: React.FC<AutoCompleteSelectProps> = ({
     setActiveIndex(-1);
   };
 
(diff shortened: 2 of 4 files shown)


client/src/components/dialog/BaseDialog.tsx
```diff
@@ -85,7 +85,12 @@ export const BaseDialog: React.FC<BaseDialogProps> = ({
       >
         {!hideHeader && (
           <>
-            <button onClick={onCloseClicked} className={CLOSE_BUTTON} aria-label="Close dialog">
+            <button
+              type="button"
```
tresdonicf, line 89: Added `type="button"` here as this may get `type="submit"` implicitly and thus the focus shifting back to that when the form is filled in

client/src/components/input/select/AutoCompleteSelect.tsx
```diff
@@ -88,19 +95,9 @@ export const AutoCompleteSelect: React.FC<AutoCompleteSelectProps> = ({
     setFilterValue("");
     onFilterChangeProp?.("", true);
     closeSelect();
+    inputRef.current?.focus();
```
tresdonicf, line 98: Specifically re-focus the same input that was just closed instead of implicitly moving focus

client/src/components/input/select/AutoCompleteSelect.tsx
```diff
@@ -88,19 +95,9 @@ export const AutoCompleteSelect: React.FC<AutoCompleteSelectProps> = ({
...
-  // Close on outside click
-  useEffect(() => {
-    const onClick = (e: MouseEvent) => {
-      if (containerRef.current && !containerRef.current.contains(e.target as Node)) {
-        closeSelect();
-      }
-    };
-    document.addEventListener("click", onClick);
-    return () => document.removeEventListener("click", onClick);
-  }, []);
```
tresdonicf, line 102: Codify this into `handleBlur` so it handles clicks and tabbing away

connor-parke-icf (approved):
LGTM, dont really have much to say, nice work!

PR #1381, opened 2026-06-04
Demos 2226 edit delete documents permissions

connor-parke-icf:
an adjustment to the permissions on documents:
* before, the permission was based on the document's association to an application. That, however, would leak documents the state user shouldnt be able to see. They should only be able to see deliverable document, not any document on an application. This was fixed by refining the filters on the permission.
* the gql schema permission directives themselves completely blocked off state users from using the delete or edit document mutations. Those fields were modified to accept the "Perform State Action" permission now. 
* but, because state users are now allowed to use those mutations, we needed to integrate the resolvers into the permission system. Thankfully, that wasnt too bad. Now state users can only mutate documents belonging to deliverables on demonstrations they are assigned to _and_ the document must be a state document, not a CMS document. This caveat meant that the mutation permission was slightly different from the view permission. As more instances of this appear, there might be a better way to organize this, but for now, theyre all listed independently. 

Added migration to initialize/modify the new permissions, updated the seeded to give us a couple more documents to work with, and some adjustments to the existing helper functions for documents to support the whole structure. 

https://github.com/user-attachments/assets/b4879cb0-564d-4b1c-ad76-33d7e90bc33c


server/src/model/document/documentData.ts
```diff
@@ -19,10 +15,38 @@ const getPermissionFilters = (userId: string) =>
         },
       },
     },
-    "View Documents on Assigned Demonstrations":
```
connor-parke-icf, line 22: main changes here. Removed the permission for viewing owned documents, and replaced the permission for viewing documents on assigned Demosntrations with a new one that focuses just on deliverables. now, state users can only see deliverable documents on deliverables attached to demonstrations they are assigned to (this excludes other documents, like application workflow docs)

server/src/model/document/documentData.ts
```diff
@@ -19,10 +15,38 @@ const getPermissionFilters = (userId: string) =>
...
+      NOT: {
+        id: {
+          in: [],
+        },
+      },
+    },
+    "Edit State Documents on Assigned Deliverables": {
+      deliverable: isAStatePointOfContactAssociatedWithDeliverable(userId),
+      deliverableIsCmsAttachedFile: false,
+    },
```
connor-parke-icf, line 36: very similar permission to the view one, with the one caveat that while they can see cms docs, they cannot edit or delete them

server/src/model/document/documentData.ts
```diff
@@ -79,3 +103,70 @@ export async function getManyDocuments(
     tx
   );
 }
+
+export async function editDocument(
```
connor-parke-icf, line 108: edit and delete document handlers now work very similarly to the getDocument function, but instead of returning the found document, it calls the edit or delete handler. If a violation is found, it logs it but returns the user the same error as if the document could not be found.

server/src/model/document/documentSchema.ts
```diff
@@ -47,9 +47,11 @@ export const documentSchema = gql`
...
   type Mutation {
     updateDocument(id: ID!, input: UpdateDocumentInput!): Document!
-      @auth(requires: ["Perform CMS Action"])
-    deleteDocument(id: ID!): Document! @auth(requires: ["Perform CMS Action"])
-    deleteDocuments(ids: [ID!]!): Int! @auth(requires: ["Perform CMS Action"])
+      @auth(requires: ["Perform CMS Action", "Perform State Action"])
+    deleteDocument(id: ID!): Document!
+      @auth(requires: ["Perform CMS Action", "Perform State Action"])
+    deleteDocuments(ids: [ID!]!): Int!
+      @auth(requires: ["Perform CMS Action", "Perform State Action"])
```
connor-parke-icf, line 54: state users can now call these mutations, the filtering is handled in the resolvers.

server/src/model/document/handleDeleteDocument.ts
```diff
@@ -1,15 +1,16 @@
...
 
 export async function handleDeleteDocument(
-  tx: PrismaTransactionClient,
-  s3Adapter: S3Adapter,
-  id: string
+  where: Prisma.DocumentWhereUniqueInput,
+  tx: PrismaTransactionClient
 ): Promise<PrismaDocument> {
-  const document = await deleteDocumentById(tx, id);
+  const s3Adapter = getS3Adapter();
```
connor-parke-icf: moved some stuff around in here, main thing is summoning the s3adapter in here so we dont need to worry about it elsewhere. The document operations are special because of that, so felt it was appropriate to keep the same "handleDeleteDocument" helper.

server/src/seeder.ts
```diff
@@ -259,6 +259,48 @@ async function seedDeliverables(actionUserId: string, actionUserPersonTypeId: Pe
   return createdDeliverables;
 }
 
+async function addDocumentsToDeliverable(deliverable: Deliverable) {
```
connor-parke-icf, line 262: went ahead and added some documents to a deliverable to test against. Note, that these cannot be actually deleted since they dont exist in s3, but the error indicated that it attempted to delete them. Other things might hold up the delete, such as having an associated application tag suggestion or BN workbook association, but these are going to be handled via the db error's return for now.

server/src/model/document/queries/updateDocument.test.ts
```diff
@@ -1,40 +1,48 @@
...
-        documentTypeId: testInput.documentType,
-      },
-    };
+  it("should update document via a transaction", async () => {
+    vi.mocked(mockTransaction.document.update).mockResolvedValue({
+      id: "doc-123-456",
+      name: "Updated Document Name",
+    });
+    const updatedDocument = await updateDocument(testWhere, testData, mockTransaction);
+    expect(mockTransaction.document.update).toHaveBeenCalledExactlyOnceWith({
```
TARehman: I think once upon a time in storied past, there was a reason we had the separate `xMocks` object and then `mockTransaction`. But for the life of me I cannot remember why, so I'm cool with this.

server/src/model/document/queries/updateDocument.ts
```diff
@@ -1,18 +1,12 @@
...
-    where: { id: id },
-    data: {
-      name: input.name,
-      description: input.description,
-      documentTypeId: input.documentType,
-    },
-  });
+  return tx
+    ? tx.document.update({ where, data })
+    : prisma().$transaction((transaction) => transaction.document.update({ where, data }));
```
TARehman: This transaction wrap doesn't actually add any value here, and I think it makes the test updates a bit confusing too because you made the tx optional. If this needs to run in a transaction (as a _part_ of a transaction) every time, then tx cannot be null. If the tx is optional, then you need two test paths: one to test if the prisma() call is made and one if it isn't.

server/src/constants.ts
```diff
@@ -499,12 +499,15 @@ export const PERMISSIONS = [
...
   "View All Documents",
-  "View Documents on Assigned Demonstrations",
-  "View Owned Documents",
+  "View Documents on Assigned Deliverables",
   "View All Deliverables",
   "View Deliverables on Assigned Demonstrations",
   "View All DemonstrationRoleAssignments",
   "View DemonstrationRoleAssignments on Assigned Demonstrations",
+  "Edit All Documents",
+  "Edit State Documents on Assigned Deliverables",
```
TARehman, line 508: NB: Are we using Assigned Deliverables as shorthand for "deliverables on demonstrations to which I am assigned?"

server/src/model/document/handleDeleteDocument.ts
```diff
@@ -1,15 +1,16 @@
...
+
+  const document = await deleteDocument(where, tx);
```
TARehman, line 10: Shouldn't we validate that the document can be deleted before we call delete document?

TARehman (commented):
I think we need to fix a small thing with transaction handling. Skimmed otherwise and it _seems_ correct; will follow up with @connor-parke-icf.

server/src/model/document/queries/updateDocument.test.ts
connor-parke-icf: the reason is because we're not actually doing anything with the transaction, just passing it down to the correct method, so it doesnt need to actually be anything special here. If we were using it to actually run a prisma query, then yeah we'd need a deeper mock

server/src/model/document/queries/updateDocument.ts
connor-parke-icf: good call. The update here _can_ actually be optional. So i made it reflect the structure of the selectX queries, including the tests. then i propagated it upwards and removed the transaction init from the resolver, since its not needed.

server/src/constants.ts
connor-parke-icf: yep! hard to strike a balance between concise and misleading, but i figure this is good enough, and reflects the pattern of the other permissions. Certainly looking to try and come up with a better convention, but this is fine for now.

connor-parke-icf (commented):
updated per comments. Adjusted transaction handling in updateDocument, moved document validation to before handleDocumentDelete

server/src/model/document/queries/deleteDocument.test.ts
```diff
@@ -24,4 +24,12 @@ describe("deleteDocument", () => {
     await deleteDocument(where, mockTransaction);
     expect(transactionMocks.document.delete).toHaveBeenCalledExactlyOnceWith({ where });
   });
+
+  it("should throw an error if the document cannot be deleted", async () => {
+    const where = { id: testDocumentId };
+    transactionMocks.document.delete.mockRejectedValueOnce("Prisma error :(");
+
+    await expect(deleteDocument(where, mockTransaction)).rejects.toThrow("Prisma error :(");
```
TARehman, line 32: NB: Isn't this just testing a passthrough error throw? If the behavior of the function doesn't do something like catch the error and call a handler or anything, what are we testing here?

server/src/model/document/documentData.ts
```diff
@@ -157,6 +158,7 @@ export async function removeDocument(
     );
 
     if (authorizedDocument) {
+      validateDocumentCanBeDeleted(authorizedDocument);
```
TARehman, line 161: NB: So I don't think its a blocker but I am thinking about the risks here of having a `deleteDocument`, a `removeDocument`, and a `handleDeleteDocument` in the same module. You have to call `removeDocument` in order to run the validation that it's acceptable to delete the document at all (regardless of what user it is). If we ever have a case where we're deleting a document separate of a user context, that might bite us. I don't think it is worth changing for now, but something about blending the two types of validation (user permissions versus business logic) feels a little off.

TARehman (approved):
Looks fine to me. I think there's some valid questions about where to put `validateDocumentCanBeDeleted` versus permissions checking but I don't think there's anything wrong per se with this.

server/src/model/document/queries/deleteDocument.test.ts
connor-parke-icf: yeah, its just testing the passthrough. if there was more going on in this function itd be a little less redundant. But i think it is a good thing to validate that if it errors we expect that error to bubble up, not get caught

server/src/model/document/documentData.ts
connor-parke-icf: i agree with this; it was kinda a way to avoid getting the document _again_ in the handleDeleteDocument method. Surely theres a better way to organize this cause youre right, the rules for whether a document can be deleted or not and the rules for who can delete a document are addressing separate concerns

PR #1382, opened 2026-06-04
DEMOS-2210: 508 Compliance - `<hr>` elements get `aria-hidden`

tresdonicf:
Per DEMOS-2210 I believe this one addresses the problem of extra text being announced before the `Cancel` and `Submit` buttons. The ticket doesn't mention which text is being read out but it seems likely that "Separator" may be announced for some screen readers  (as this can be semantically important if a separator is used to create a quasi-list or something)

Upon re-test I will ask the tester specifically which text is being read out if this doesn't address it, but this change is accurate for our purposes anyhow (these separators are just decorative and shouldn't be in the aria tree)

### DIFF
--- a/client/src/components/application/amendment/AmendmentWorkflow.tsx
+++ b/client/src/components/application/amendment/AmendmentWorkflow.tsx
@@ -50,22 +50,9 @@ export const GET_AMENDMENT_WORKFLOW_QUERY = gql`
   ${WORKFLOW_DOCUMENT_FIELDS}
 `;
 
-export type ApplicationWorkflowAmendment =
-  WorkflowApplication &
-  Pick<
-    Amendment,
-    "name" |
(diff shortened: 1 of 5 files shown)


client/src/components/application/amendment/AmendmentWorkflow.tsx
```diff
@@ -50,22 +50,9 @@ export const GET_AMENDMENT_WORKFLOW_QUERY = gql`
   ${WORKFLOW_DOCUMENT_FIELDS}
 `;
 
-export type ApplicationWorkflowAmendment =
-  WorkflowApplication &
-  Pick<
-    Amendment,
```
tresdonicf, line 56: Prettier changes I guess

james-becwar-cms (approved):

PR #1384, opened 2026-06-04
DEMOS-1772: added polling system - WIP - Pushing back and trying a couple other things

dustbuster:
At upload this will poll for changes every 5 seconds for two minutes, then stops. 
upon a fresh State Application upload it will start again.

Then voila! It finds a results. 
Screenshot 2026-06-04 at 13 20 05

### DIFF
--- a/client/src/components/application/phases/ApplicationIntakePhase.tsx
+++ b/client/src/components/application/phases/ApplicationIntakePhase.tsx
@@ -1,6 +1,6 @@
-import React, { useState } from "react";
+import React, { useEffect, useState } from "react";
 
-import { gql, TypedDocumentNode, useMutation } from "@apollo/client";
+import { gql, TypedDocumentNode, useApolloClient, useMutation } from "@apollo/client";
 
 import { Button, SecondaryButton } from "components/button";
 import { ExportIcon } from "components/icons";
@@ -16,7 +16,15 @@ import { getPhaseCompletedMessage } from "util/messages";
 import { useToast } from "components/toast";
 import { DatePicker } from "components/input/date/DatePicker";
 import { ApplicationHealthTypeTags } from "components/tags/ApplicationHealthTypeTags";
-import type { Application, LocalDate, PhaseName, PhaseStatus, Tag, TagName } from "demos-server";
+import type {
+  Application,
+  LocalDate,
+  PhaseName,
+  PhaseStatus,
+  Tag,
+  TagName,
+  UploadDocumentToPhaseInput,
+} from "demos-server";
 import { SET_APPLICATION_TAGS_MUTATION } from "components/dialog/ApplyTagsDialog";
 import { ConfirmSuggestedSparklyTagDialog } from "components/dialog/ConfirmSuggestedSparklyTagDialog";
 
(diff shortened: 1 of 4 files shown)


tresdonicf (commented):
I'm a little confused about the need for this feature. Are we communicating to the user that they should stay on the same page and wait for 2 minutes for suggestions? Are we expecting them to do that or are we expecting them to navigate away?

The way it might be implemented now it seems like it'll continue polling even if the user clicks on this tab and continues going through them which might slow down the application for the subsequent phases.

I would probably push back on the need for this feature and instead suggest it's better if the user comes back to the phase selector with some suggestions in there (as these phases take a good amount of clock time anyhow)

client/src/components/application/phases/ApplicationIntakePhase.tsx
```diff
@@ -16,21 +16,23 @@ import { getPhaseCompletedMessage } from "util/messages";
...
+const UIPATH_SUGGESTION_POLL_INTERVAL_MS = 5_000;
+// stops after 2 minutes or when user leaves the phase.
+const UIPATH_SUGGESTION_POLL_TIMEOUT_MS = 120_000;
 
-- **State Application Submitted Date** - When the state formally submits their application.
-
-- **Application Intake Completion Date** - Completed when user clicks "Finish" to progress to the next phase.
-
-Note: If the user skips the concept phase this will be marked completed when the user clicks Finish on the Application Intake Phase.
- */
```
dustbuster, line 33: I feel like we all know how phases work generally. Perhaps time to pul this out.

client/src/components/application/phases/ApplicationIntakePhase.tsx
```diff
@@ -16,21 +16,23 @@ import { getPhaseCompletedMessage } from "util/messages";
...
 import { ConfirmSuggestedSparklyTagDialog } from "components/dialog/ConfirmSuggestedSparklyTagDialog";
 
-/** Business Rules for this Phase:
- * - **Application Intake Start Date** - Can start in one of two ways, whichever comes first:
-  - a. User clicked Skip or Finish on the Concept Phase
-  - b. When a change is submitted on this phase - document or date update.
+// Polling for refetch suggested tags for uipath every 5 seconds after file upload
+const UIPATH_SUGGESTION_POLL_INTERVAL_MS = 5_000;
+// stops after 2 minutes or when user leaves the phase.
+const UIPATH_SUGGESTION_POLL_TIMEOUT_MS = 120_000;
```
dustbuster, line 35: These are not really re-usable. So i got them as far to the top of the page as i could.

dustbuster:
> I'm a little confused about the need for this feature. Are we communicating to the user that they should stay on the same page and wait for 2 minutes for suggestions? Are we expecting them to do that or are we expecting them to navigate away?
> 
> The way it might be implemented now it seems like it'll continue polling even if the user clicks on this tab and continues going through them which might slow down the application for the subsequent phases.
> 
> I would probably push back on the need for this feature and instead suggest it's better if the user comes back to the phase selector with some suggestions in there (as these phases take a good amount of clock time anyhow)

So the need for this. You can do the entire phase walkthru without refreshing anything. This IS where UIPath meets the app. I would prefer it follow the same flow as the rest of the code and just pop up when ready. that would be ideal. But we do not have a simple way to do that. 

it does clutter things up a bit, But the phases are pretty cluttery on their own anyway. 
But this is a web app. With business logic. And like to make the uipath code complete i think we need to auto pop those suggestions without refresh,

However I think RN I'm not super happy with this YAPS, "yet another polling system". UIPath polls, then this polls for UIPath. It's a whole lotta weird polling. I checked GQL subscriptions, which seem to exist for use cases like this, and that's a heavier lift. 

I am pushing on this, however, I do want to try a couple things. See if i can make it simpler.

dustbuster:
Spoke with David Niedergeses about this, not going to move forward on the complexity of polling yet.

PR #1385, opened 2026-06-04
DEMOS-2226: made auth capable of picking from a list

connor-parke-icf:
simple change to have the create user function of the auth capable of finding the relevant person type among the value of the claims.role field, which is actually a comma-separated list of values rather than always a single value. Used basic string and array functions to do this, and tested for various configurations of the role being, or not being, or being _too_ present. Validated locally after reseeding and saw my users are still created 
image

### DIFF
--- a/server/src/auth/userContext.ts
+++ b/server/src/auth/userContext.ts
@@ -2,6 +2,7 @@ import { prisma } from "../prismaClient";
 import { Permission, SystemRole, UserType } from "../types";
 import { AuthorizationClaims } from "./auth.util";
 import { upsertUserSession } from "../model/userSession/queries";
+import { USER_TYPES } from "../constants";
 
 const initialUserTypeRoles: Record<UserType, SystemRole> = {
   "demos-admin": "Admin User",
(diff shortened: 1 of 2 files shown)


server/src/auth/userContext.ts
```diff
@@ -16,11 +17,27 @@ export type ContextUser = {
   permissions: Permission[];
 };
 
+export const getPersonTypeFromClaims = (claims: AuthorizationClaims): UserType => {
```
connor-parke-icf, line 20: I tried using the set.intersection method here, but apparently it does not exist in our current TS config, which is set for lib and target to be "es2021". Apparently, set.intersection is newer than that? thought about bumping this up a couple years but then... decided that it may not be worth the trouble. I have no idea the side effects of that, and figured a change like that might be best saved for its own dedicated ticket...post mvp. 

So just sticking with boring old array mechanics.  I figure these lists will remain pretty small, so its not like theres a runtime complexity consideration that should be made.

dustbuster: Booo!

server/src/auth/userContext.ts
```diff
@@ -2,6 +2,7 @@ import { prisma } from "../prismaClient";
 import { Permission, SystemRole, UserType } from "../types";
 import { AuthorizationClaims } from "./auth.util";
 import { upsertUserSession } from "../model/userSession/queries";
+import { USER_TYPES } from "../constants";
 
 const initialUserTypeRoles: Record<UserType, SystemRole> = {
   "demos-admin": "Admin User",
```
dustbuster, line 8: so these roles are not what's in IDM for this? 

Image

dustbuster (approved):
See my question. probably have idm set up differently than in all roles from PMDA. but thought i'd ask.

server/src/auth/userContext.ts
connor-parke-icf: no, this is the PersonType/UserType of which we only have three options in DEMOS: demos-admin, demos-cms-user, and demos-state-user. We have another set of Roles which are used in the permission system though, which eventually i assume would correspond more closely to these roles. Currently, the Roles we have are a one-to-one mapping of the personTypes, which is why it may feel redundant: Admin User, CMS User, and State User. But i know we will be adding more roles post-mvp, like Readonly CMS User, and another one i cant remember the name of.  Those will be users with a personType of demos-cms-user, but will have a Role of Readonly CMS User, for instance.

PR #1387, opened 2026-06-04
DEMOS-1819-basic-report-running-infrastructure

TARehman:
This is the basics of the report-running infrastructure - there is a lot of work yet to come, but this is the initial stuff and I thought breaking the code into smaller PRs might be easier to review. This adds a concept of an on-demand report to the database and a module for configuring and running arbitrary SQL reports, along with a simple test resolver that calls it to confirm it works. This also adds `zod` to help verify what's coming in - because this is the first time we're going to write SQL directly and then need to work with the objects that come back I felt validation was a good idea to ensure that we don't have drift between the SQL query and the type in TS.

### DIFF
--- a/server/src/errors/errorCodes.ts
+++ b/server/src/errors/errorCodes.ts
@@ -1,4 +1,4 @@
-import { GraphQLFormattedError } from "graphql";
+import { GraphQLError, GraphQLFormattedError } from "graphql";
 
 export const CUSTOM_INTERNAL_ERROR_CODES = [
   "ON_DEMAND_REPORT_ZOD_ERROR",
@@ -21,11 +21,6 @@ export function isCustomInternalErrorCode(value: string): value is CustomInterna
   return (CUSTOM_INTERNAL_ERROR_CODES as readonly string[]).includes(value);
 }
 
-const CUSTOM_PUBLIC_ERROR_MESSAGES: Record<CustomPublicErrorCode, string | undefined> = {
-  ON_DEMAND_REPORT_ERROR: "An error occurred while running an on-demand report.",
-  REFERENCE_ERROR: undefined,
-};
-
 export const CUSTOM_ERROR_CODES: Record<
   CustomInternalErrorCode,
   { publicErrorCode: CustomPublicErrorCode; logLevel: ErrorLogLevel }
@@ -38,6 +33,11 @@ export const CUSTOM_ERROR_CODES: Record<
   REFERENCE_AGREEMENT_NOT_ACTIVE: { publicErrorCode: "REFERENCE_ERROR", logLevel: "debug" },
 } as const;
 
+const CUSTOM_PUBLIC_ERROR_MESSAGES: Record<CustomPublicErrorCode, string | undefined> = {
+  ON_DEMAND_REPORT_ERROR: "An error occurred while running an on-demand report.",
+  REFERENCE_ERROR: undefined,
+};
+
 export function getPublicErrorCodeFromInternal(
   errorCode: CustomInternalErrorCode
 ): CustomPublicErrorCode {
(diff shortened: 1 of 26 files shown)


server/src/model/migrations/20260604191542_prisma_add_on_demand_report_tables/migration.sql
```diff
@@ -0,0 +1,35 @@
...
+-- CreateTable
+CREATE TABLE "on_demand_report" (
+    "id" UUID NOT NULL,
+    "requesting_user_id" UUID NOT NULL,
+    "report_type_id" TEXT NOT NULL,
+    "status_id" TEXT NOT NULL,
+    "report_generated_at" TIMESTAMP(3) NOT NULL,
+
+    CONSTRAINT "on_demand_report_pkey" PRIMARY KEY ("id")
+);
```
james-becwar-cms, line 13: Would it make sense to include the s3 file path to the generated report?

james-becwar-cms (commented):
We prob need to add some role authentication around the reports so a state user can't run them.

james-becwar-cms (approved):

server/src/errors/errorCodes.ts
connor-parke-icf, line 1: I like this; im going to need to mess with it to get a feel for it. Im sure we will open up a ticket at some point soon to audit the backend to flesh out all the return messages here

server/src/model/migrations/20260604191621_add_on_demand_report_tables/migration.sql
```diff
@@ -0,0 +1,17 @@
...
+ADD CONSTRAINT
+    check_non_empty_s3_path
+CHECK (
+    trim(s3_path) != ''
+);
+
+INSERT INTO
+    demos_app.on_demand_report_status
+VALUES
+    ('Available');
```
connor-parke-icf, line 12: im assuming that we'll have a couple more statuses inserted here like "pending" or "failed" later?

server/src/model/onDemandReport/onDemandReportSchema.ts
```diff
@@ -0,0 +1,8 @@
+import { gql } from "graphql-tag";
+
+export const onDemandReportSchema = gql`
+  type Mutation {
+    generateOnDemandReport(reportType: OnDemandReportType!): String!
+      @auth(requires: ["Perform CMS Action"])
```
connor-parke-icf, line 6: nice!

server/src/onDemandReports/schemas/basicTestReportSchema.ts
```diff
@@ -0,0 +1,12 @@
+import { z } from "zod";
+import { APPLICATION_STATUS, STATES_AND_TERRITORIES } from "../../constants";
+
+export const basicTestReportSchema = z
```
connor-parke-icf, line 4: ohh i see! it can bridge typescript and sql by acting like a typeguard. this is pretty useful

connor-parke-icf (approved):
LGTM! that zod stuff is pretty cool, i imagine with this we can safely type the report in case we even needed to do something more complicated with it.  I imagine next is the uploading to s3 and download fields and whatnot

server/src/model/migrations/20260604191621_add_on_demand_report_tables/migration.sql
TARehman: This is mostly futureproofing. Eventually we might want to do something like "delete all reports older than a year" in S3 or something like that, and we'll need a status to indicate that they're no longer present in the system.

server/src/onDemandReports/schemas/basicTestReportSchema.ts
TARehman: I don't think it directly gets us our custom types inferred from Zod - I need to investigate that further (and figure out if it's even worth doing). But what it does get us at least is knowing the number of columns, their primitive types, etc.

PR #1395, opened 2026-06-08
DEMOS-1283 Client-side BN Workbook pre-upload validation

nunesmlucas:
Before a user uploads a BN Workbook, parse it in the browser and run it through
the shared validation library (`shared_library/BN`, the same one the lambda
uses server-side). If any rule fails, show the errors inline and **block the
upload** until a valid file is picked.

The post-upload polling from the original DEMOS-1283 still runs as before — the
lambda remains the source of truth. This adds client-side validation as a fast
first check.

## UX

| Doc type | Behavior |
|---|---|
| `BN Workbook` + invalid file | Red bordered notice listing every rule error, Upload button disabled |
| `BN Workbook` + valid file | Green "File Validated Successfully" notice, Upload enabled |
| `BN Workbook` + still parsing | "Validating workbook..." line, Upload disabled |
| Any other doc type | No validation runs, Upload enabled normally |

image

https://github.com/user-attachments/assets/a4fe89b4-181b-4eb3-b8b2-c36f14aa8b27

### DIFF
--- a/client/src/components/dialog/document/DocumentDialog.test.tsx
+++ b/client/src/components/dialog/document/DocumentDialog.test.tsx
@@ -2,8 +2,8 @@ import "@testing-library/jest-dom";
 
 import React from "react";
 
-import { describe, it, expect, vi } from "vitest";
-import { fireEvent, render, screen } from "@testing-library/react";
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { fireEvent, render, screen, waitFor } from "@testing-library/react";
 
 import { ToastProvider } from "components/toast/ToastContext";
 import { DocumentType } from "demos-server";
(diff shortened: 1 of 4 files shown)


nunesmlucas:
Quick note on the shared validation library (worth its own ticket): three of the 12 rules (rules 8, 9, 10) crash when a workbook has fewer rows than the rule's hardcoded range — they ask for row 248 of a sheet that ends at row 150 and throw. The library runs all rules in a single chain, so as soon as one crashes the rest never run and the user only sees one cryptic error. The lambda has the same bug, just silently. In this PR I run each rule in its own try/catch in the client so we still get the full list of errors, but the real fix is in the library itself (bound-check rows or stop the chain from short-circuiting).

dustbuster (approved):
This seems about right for BN workbooks!

PR #1396, opened 2026-06-08
Disable data trace on the api gateway

cms-jesse:
Since we have not needed it for any troubleshooting in the lower environments, this is being disabled to prevent any sensitive data logging

https://docs.aws.amazon.com/apigateway/latest/api/API_MethodSetting.html

```
This can be useful to troubleshoot APIs, but can result in logging sensitive data. We recommend that you don't enable this option for production APIs.
```


dustbuster (approved):
Thank you!

PR #1397, opened 2026-06-08
DEMOS-2240: added Status lock down for deliverables

dustbuster:
Deliverables Can only be deletes if upcoming or past due!

Also updated the tooltip and came up with the most succinct way i could come up with for the HTML element title/tooltip.

Screenshot 2026-06-08 at 15 07 52

### DIFF
--- a/client/src/components/table/tables/DeliverableActionButtons.tsx
+++ b/client/src/components/table/tables/DeliverableActionButtons.tsx
@@ -11,7 +11,7 @@ import { selectionTooltip } from "./actionTooltips";
 import type { FormattedDeliverableTableRow } from "./DeliverableTable";
 
 export const DELIVERABLE_CANT_DELETE_HAS_FILES =
-  "Deliverables can only be deleted \nif they are Upcoming or Past Due \nand have no comments or files.";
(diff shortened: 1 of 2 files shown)


client/src/components/table/tables/DeliverableActionButtons.tsx
```diff
@@ -42,7 +49,9 @@ export const DeliverableActionButtons: React.FC<{
   const selectedCount = selectedRows.length;
   const selectedDeliverables = selectedRows.map((row) => row.original);
   const selectedHasFilesOrComments = selectedDeliverables.some(hasFilesOrComments);
-  const deleteEnabled = selectedCount >= 1 && !selectedHasFilesOrComments;
+  const selectedAllHaveDeletableStatuses = selectedDeliverables.every(hasDeletableStatus);
+  const deleteEnabled =
+    selectedCount >= 1 && selectedAllHaveDeletableStatuses && !selectedHasFilesOrComments;
```
dustbuster, line 54: It gets the job done! Even if it's a bit wordy.

james-becwar-cms (approved):

PR #1398, opened 2026-06-08
DEMOS-2085 - Fix for post submission deliverable documents

nunesmlucas:
Documents on a deliverable's **State Files** and **CMS Files** tabs can now only be
**deleted before the deliverable is submitted**. Add/Edit and View/Download behavior
is unchanged.

## Rules
- **Delete** allowed only when status is `Upcoming` or `Past Due`.
  Blocked for `Submitted`, `Under CMS Review`, `Accepted`, `Approved`, `Received and Filed`.
- **Add / Edit** allowed until a final status (`Accepted`, `Approved`, `Received and Filed`).
- **View / Download** always available (still requires access to the deliverable).
- Applies to **all users** (CMS/Admin and State).
  - CMS/Admin: manage both tabs.
  - State users: manage State Files only; view/download CMS Files.

## The bug
Delete was gated on the same flag as Add/Edit, which only locks the 3 final statuses, so
delete stayed enabled in `Submitted` / `Under CMS Review`. The backend only blocked
documents stamped at submission time, so files **added after** submission could still be
deleted via the API.
Screenshot 2026-06-08 at 3 17 43 PM

### DIFF
--- a/client/src/pages/deliverables/sections/FileAndHistoryTabs.tsx
+++ b/client/src/pages/deliverables/sections/FileAndHistoryTabs.tsx
@@ -46,6 +46,18 @@ export const RESUBMISSION_DISABLED_STATUSES: ReadonlySet<DeliverableStatus> =
 export const isResubmissionDisabled = (status: DeliverableStatus): boolean =>
   RESUBMISSION_DISABLED_STATUSES.has(status);
 
+// Documents may only be deleted before the deliverable has been submitted. Once it
+// is Submitted, Under CMS Review, or in any final status, files are locked from
+// deletion for every user (Add/Edit remain available until a final status — see
+// isDeliverableEditable). Expressed as an allow-list so any new status defaults to locked.
+export const FILE_DELETION_ALLOWED_STATUSES: ReadonlySet<DeliverableStatus> = new Set([
+  "Upcoming",
+  "Past Due",
+]);
+
+export const isFileDeletionLocked = (status: DeliverableStatus): boolean =>
+  !FILE_DELETION_ALLOWED_STATUSES.has(status);
+
 const CMS_STAFF_PERSON_TYPES: ReadonlySet<PersonType> = new Set(["demos-admin", "demos-cms-user"]);
 
 const STATE_FILE_MANAGER_PERSON_TYPES: ReadonlySet<PersonType> = new Set([
@@ -89,6 +101,7 @@ export const FileAndHistoryTabs: React.FC<{
   const cmsFiles = deliverable.cmsDocuments;
   const historyRows: DeliverableHistoryRow[] = deliverable.deliverableActions.map(toHistoryRow);
   const isFinalized = !isDeliverableEditable(deliverable.status);
+  const isDeleteLocked = isFileDeletionLocked(deliverable.status);
   const refetchAfterFileChange = [DELIVERABLE_DETAILS_QUERY];
 
   if (!currentUser) {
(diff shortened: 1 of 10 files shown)


server/src/model/document/documentData.ts
```diff
@@ -158,7 +160,11 @@ export async function removeDocument(
     );
 
     if (authorizedDocument) {
-      validateDocumentCanBeDeleted(authorizedDocument);
+      const deliverableStatus = authorizedDocument.deliverableId
+        ? ((await selectDeliverableOrThrow({ id: authorizedDocument.deliverableId }, tx))
+            .statusId as DeliverableStatus)
+        : null;
+      validateDocumentCanBeDeleted({ ...authorizedDocument, deliverableStatus });
```
nunesmlucas, line 167: looks up deliverables status and passes to validator

server/src/model/document/validateDocumentCanBeDeleted.ts
```diff
@@ -1,10 +1,26 @@
...
     );
   }
+
+  if (
+    document.deliverableStatus &&
+    DOCUMENT_DELETION_LOCKED_DELIVERABLE_STATUSES.has(document.deliverableStatus)
+  ) {
+    throw new Error(
+      `Document with ID ${document.id} cannot be deleted because its deliverable has been submitted.`
+    );
```
nunesmlucas, line 24: an extra layer here using our new set

james-becwar-cms (approved):

server/src/model/document/validateDocumentCanBeDeleted.ts
connor-parke-icf: hmm, this may be a requirements issue. As I understand it, users can submit a deliverable more than once, which means they can 
1) upload a file,
2) submit the deliverable,
3) upload another file. 

That second file as i understand it should still be eligible for deletion, even though the deliverable is in a submitted state. This is what the check against the `is part of a deliverable submission` is doing, to differentiate between files that have been submitted and files that are uploaded, but not yet submitted. 

I think we may need to open this conversation up; we had a lot of discussion regarding what issues allowing multiple submissions may cause, and this seems to point towards a single-submission idea. I could be misremembering though; its been a while since those conversations.

client/src/pages/deliverables/sections/FileAndHistoryTabs.tsx
```diff
@@ -46,6 +46,14 @@ export const RESUBMISSION_DISABLED_STATUSES: ReadonlySet<DeliverableStatus> =
 export const isResubmissionDisabled = (status: DeliverableStatus): boolean =>
   RESUBMISSION_DISABLED_STATUSES.has(status);
 
+export const FILE_DELETION_ALLOWED_STATUSES: ReadonlySet<DeliverableStatus> = new Set([
+  "Upcoming",
+  "Past Due",
+]);
```
dustbuster, line 52: I literally added something very similar. maybe we can refactor both our code and add this to constants.

server/src/model/document/validateDocumentCanBeDeleted.ts
dustbuster: I've been taking the assumption this was a revolving door situation. But i did not consider the deletion of files. Would we just not return the deliverable status to Upcoming if resubmission was requested? 

I think the PR is ok, because then we could see where we are at to make adjustments once we get all our current tickets completed and tested? idk

dustbuster (approved):
I say we get it working, and tweak as needed to fit resubmitted deliverable files.

PR #1400, opened 2026-06-08
DEMOS-1819-upload-reports

TARehman:
This PR persists the results of on-demand results into S3, writes them into the database, and returns a pre-signed download URL. It does not yet format things for Excel.

- Adds the new report type, `Deliverable Status Report`, that we actually need to generate. I swapped the basic test report SQL with into that as a temporary step so that the basic test report is truly basic.
- Report generation now runs in a transaction, captures metadata in the DB, sticks the results into S3, and returns a presigned URL to download the data. If report generation fails, it cleans up the file if it exists, and also logs if that cleanup fails.
- Added S3 adapter items for the on-demand results.
- Moved the report queries into `ts` files to avoid issues with needing to move SQL files around that were causing problems between CJS and ESM, which means I also changed what was in the build commands in `package.json`.
- Added the `upload` permission so that we can write things to the clean bucket directly.
- In working on updating the tests, I ended up nearly completely rewriting `AwsS3Adapter.test.ts` to better conform to how almost all of our other tests are structured.

Should definitely get a quick look from @cms-jesse to be sure I didn't foul up the infrastructure stuff.


server/src/adapters/s3/LocalS3Adapter.ts
```diff
@@ -22,6 +23,11 @@ export function createLocalS3Adapter(): S3Adapter {
         return `${HOSTNAME}/${BUCKET_NAME}/${key}?download=true&expires=3600`;
       }
 
+      const onDemandReport = uploadedOnDemandReports.get(key);
+      if (onDemandReport) {
+        return `${HOSTNAME}/${BUCKET_NAME}/${key}?download=true&expires=3600&size=${onDemandReport.byteLength}`;
```
james-becwar-cms, line 28: Should we hardcode the expires?  1 hour seems long to start the download.

server/src/model/onDemandReport/onDemandReport.prisma
```diff
@@ -1,7 +1,7 @@
 // Flags: NO_CREATED_TS, NO_UPDATED_TS, NO_HISTORY
 // File logs when on demand reports are generated, intended as append-only
 model OnDemandReport {
-  id                String   @default(uuid()) @db.Uuid
+  id                String   @db.Uuid
```
james-becwar-cms, line 4: Just double check that we don't' want a default id?

james-becwar-cms (changes requested):
During the last pen test they noted the long expires, I think we needed to change it to 5 minutes?

server/src/adapters/s3/LocalS3Adapter.ts
james-becwar-cms: whoops its the local adapter...

james-becwar-cms (approved):

cms-jesse (approved):
The only real change I see in the deployment dir is `cleanBucket.grantPut(graphqlLambda.lambda.role);`, and that looks good. Everything else seems to be just formatting

server/src/model/onDemandReport/onDemandReportResolvers.ts
```diff
@@ -1,25 +1,63 @@
...
+            Buffer.from(JSON.stringify(reportResults))
+          );
+          await insertOnDemandReport(
+            {
+              id: reportId,
+              s3Path: fileS3Path,
+              requestingUserId: context.user.id,
+              reportTypeId: args.reportType,
+              statusId: "Available",
+              reportGeneratedAt: new Date(),
```
connor-parke-icf, line 37: isnt this the same thing as a createdAt column?

connor-parke-icf (approved):
nothing that really stands out to me. i like that the query files are encapsulated in runtime objects so we dont need to mess with files themselves. Moving the autogen UUID makes sense to me. Theres a bunch in the zod schema that im unfamiliar with, but it seems to follow a basic shape of what id expect

server/src/model/onDemandReport/onDemandReportResolvers.ts
TARehman: Yeah, not totally sure why I didn't just use createdAt? Maybe because we didn't use createdAt in other logging-style records.

server/src/model/onDemandReport/onDemandReport.prisma
TARehman: We don't, because we need to get the ID in advance anyway to build the s3Path, so I thought it made sense to force explicitness.

PR #1401, opened 2026-06-08
DEMOS-2237: Fix UserProvider for Unautenticated Users

tresdonicf:
Oops, I was using cached creds and realized I broke the flow for when you're signed out. A user that is not authenticated needs to render children, as this will render the sign in page:

I did scope creep a little here and add a git commit line to the footer as well that way we can see in screenshots what commit a change is on, this is guarded by the `<DebugOnly>` component which means it will only render in local development

https://github.com/user-attachments/assets/d070704a-b7a2-4350-ba0f-6a7855d0def9

### DIFF
--- a/client/src/components/footer/Footer.tsx
+++ b/client/src/components/footer/Footer.tsx
@@ -1,5 +1,6 @@
 import { HhsLogo } from "components/brand/HhsLogo";
 import { LogoSimplified } from "components/brand/LogoSimplified";
+import { DebugOnly } from "components/debug/DebugOnly";
 import { useToast } from "components/toast";
 import React from "react";
 import { tw } from "tags/tw";
@@ -100,7 +101,9 @@ const FooterUpper: React.FC = () => (
 
 const FooterLower: React.FC = () => (
   <div className="flex w-full bg-brand text-white p-1">
-    <div className="w-1/3" />
+    <div className="w-1/3">
+      <DebugOnly>git commit: {__GIT_COMMIT__}</DebugOnly>
+    </div>
     <div className="w-1/3" />
     <div className="w-1/3 text-right">{DEMOS_ADDRESS}</div>
   </div>
--- a/client/src/components/user/UserProvider.tsx
+++ b/client/src/components/user/UserProvider.tsx
@@ -38,8 +38,13 @@ export function UserProvider({ children }: { children: React.ReactNode }) {
     );
   }
 
-  // Render authentication failure component if there was an error
-  if (error || !data?.currentUser) {
+  // Not yet authenticated — render children so withAuthenticationRequired can redirect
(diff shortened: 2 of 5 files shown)


client/src/components/user/UserProvider.tsx
```diff
@@ -38,8 +38,13 @@ export function UserProvider({ children }: { children: React.ReactNode }) {
     );
   }
 
-  // Render authentication failure component if there was an error
-  if (error || !data?.currentUser) {
+  // Not yet authenticated — render children so withAuthenticationRequired can redirect
+  if (!auth.isAuthenticated) {
+    return <>{children}</>;
+  }
```
tresdonicf, line 44: These are users who haven't tried to put in credentials yet (or have recently signed out). We just show children in this case a it will redirect to the login page

client/src/components/footer/Footer.tsx
```diff
@@ -100,7 +101,9 @@ const FooterUpper: React.FC = () => (
 
 const FooterLower: React.FC = () => (
   <div className="flex w-full bg-brand text-white p-1">
-    <div className="w-1/3" />
+    <div className="w-1/3">
+      <DebugOnly>git commit: {__GIT_COMMIT__}</DebugOnly>
```
tresdonicf, line 105: Added git commit line to footer in local dev

client/src/components/user/UserProvider.tsx
dustbuster: Do we want to display this on production systems? or should we limit this to non-production? 
idk if this brings value to our clients. It brings value to use. but not them.

tresdonicf: This is the line to fix the main issue so I think  your comment might be about the git hash. Currently it is limited to non-production, or even further it's only present in local development

client/vite.config.ts
```diff
@@ -1,12 +1,24 @@
...
 import tailwindcss from "@tailwindcss/vite";
 import react from "@vitejs/plugin-react";
 
+const getGitCommit = (): string => {
+  try {
+    return execSync("git rev-parse --short HEAD").toString().trim();
+  } catch {
+    return "unknown";
+  }
+};
```
dustbuster, line 15: nevermind. Missed the <DebugOnly>

client/src/components/footer/Footer.tsx
dustbuster: DebugOnly ah ok. this makes a lot more sense now!

client/src/components/user/UserProvider.tsx
dustbuster: Nevermind! missed the <DebugOnly>

dustbuster (approved):
LGTM!

PR #1407, opened 2026-06-09
(2/2) DEMOS-2237: Cleanup checks for `!currentUser`

tresdonicf:
Now that `currentUser` is not optional on the `UserContext` we can clean up all this handling for `if(!currentUser)` around the codebase 🥳

### DIFF
--- a/client/src/pages/deliverables/sections/FileAndHistoryTabs.tsx
+++ b/client/src/pages/deliverables/sections/FileAndHistoryTabs.tsx
@@ -40,8 +40,13 @@ export const SUBMIT_DELIVERABLE_MUTATION = gql`
   }
 `;
 
-export const RESUBMISSION_DISABLED_STATUSES: ReadonlySet<DeliverableStatus> =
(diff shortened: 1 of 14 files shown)


james-becwar-cms (approved):

client/src/pages/deliverables/sections/FileAndHistoryTabs.tsx
```diff
@@ -100,14 +105,10 @@ export const FileAndHistoryTabs: React.FC<{
   const isDeleteLocked = isFileDeletionLocked(deliverable.status);
   const refetchAfterFileChange = [DELIVERABLE_DETAILS_QUERY];
 
-  if (!currentUser) {
-    throw new Error("FileAndHistoryTabs requires an authenticated user.");
-  }
```
tresdonicf, line 105: We had to have a bunch of these sprinkled about in the codebase since `getCurrentUser` could return `currentUser: undefined` - now it's no longer needed

PR #1408, opened 2026-06-09
DEMOS-1819-create-excel-report-file

TARehman:
This allows for creating of a formatted Excel report file from a query. I went back and forth awhile on the best ways to handle the typing and I feel that this is a good approach.

- Settled on using `excelts` as the library for the time being, and added this to the install files. This appears to be a decent, fairly up-to-date, TS-native implementation.
- Moved the `queries` into the same folders as the formerly-called `schemas` and renamed it to `configs`.
  - The `*Config.ts` files contain all the formatting and type information for a single report.
  - The `*Query.ts` files contain the actual SQL query as a multi-line string.
  - For every report, we'll have one of each; the `*Query.ts` file gets imported into the `*Config.ts` file, where we also set up various things like nicely-named string headers, etc.
  - I added some custom types for `configs` that are in their own file (to avoid circular dependencies with putting them in `index.ts`. These types are more generic so that downstream functions are more aware of what should be in every report type in terms of columns, etc.
  - In `index.ts` there is the exported report configurations, as well as an accessor (which makes some of the mocking of tests easier). This is also where we can infer a type from the Zod configuration for each row, which can be used in the formatting functions.
- Simplified the `runOnDemandReport.ts` implementation very mildly and simplified the tests by ignoring the types in specific cases where they're not relevant, with comments to document why.
- Added `formatOnDemandReportInExcel.ts` with very simple formatting for now; however, my hope is that if we need to add complexity, it can become part of the report configuration, and be passed into the function, keeping it fairly generic.
- Updated the resolver to actually call the Excel formatter. Tested locally and it seems to be working!

Some housekeeping items:
- Added `S3_SKIP_SIGNATURE_VALIDATION` to the docker-compose.yml file for local dev; this makes `localstack` enforce the timeout on download links among other things.
- Updated `concurrently` to close a critical security bug; only used in dev dependencies.


.devcontainer/docker-compose.yml
```diff
@@ -62,6 +62,7 @@ services:
       CORS_ALLOWED_HEADERS: "Content-Type,Authorization,X-Amz-Date,X-Api-Key,X-Amz-Security-Token,X-Amz-User-Agent"
       CORS_EXPOSE_HEADERS: "ETag"
       CORS_ALLOW_CREDENTIALS: "true"
+      S3_SKIP_SIGNATURE_VALIDATION: 0
```
james-becwar-cms, line 65: I didn't see this being used?

james-becwar-cms (approved):

.devcontainer/docker-compose.yml
TARehman: If you don't enable that, `localstack` doesn't enforce the timeout of the pre-signed URLs and other things locally.

connor-parke-icf (approved):
looks good to me; good find on the S3_SKIP_SIGNATURE_VALIDATION fix. Overall the xlsx formatting looks straightforward. no real comments

PR #1409, opened 2026-06-09
DEMOS-2225: removed extra validation, since its covered elsewhere

connor-parke-icf:
in my previous fix, i had forgotten this one place where the role is validated before progressing through user generation. I removed it, since that validation is handled at the user generation itself, when it extracts the relevant role.

### DIFF
--- a/server/src/auth/auth.util.test.ts
+++ b/server/src/auth/auth.util.test.ts
@@ -82,15 +82,6 @@ describe("validateClaims", () => {
     expect(() => validateClaims(claims)).toThrow("Authorizer claims missing required 'role' field");
   });
 
-  it("throws when role is invalid", () => {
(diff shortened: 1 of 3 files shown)


james-becwar-cms (approved):

server/src/auth/auth.util.test.ts
```diff
@@ -82,15 +82,6 @@ describe("validateClaims", () => {
     expect(() => validateClaims(claims)).toThrow("Authorizer claims missing required 'role' field");
   });
 
-  it("throws when role is invalid", () => {
-    const claims: Partial<AuthorizationClaims> = {
-      ...validClaims,
-      role: "not-a-real-role",
```
dustbuster, line 88: Not any moe!

dustbuster (approved):
okey dokie!

PR #1410, opened 2026-06-09
Demos 2235 2: soften errors, Skip Non-PDFs, changed errors to log.info unless the error would cause uipath to fail on a good file 

dustbuster:
okay. the re-adjustment of uipath error handling. 
1. stopped execution of uipath if non-pdf is deternined based on file rtoe. 
2. Changed any `log.errors` to `log.info`
3. Unless there's a config error then it will throw. 

There's still an error I've not been able to find in 2 days. But need to test this on dev/test.
```json
{"name":"error","message":"duplicate key value violates unique constraint \"uipath_result_request_id_key\""}
```

Screenshot 2026-06-10 at 11 29 10

Screenshot 2026-06-10 at 11 25 26


lambdas/UIPath/index.ts
```diff
@@ -53,17 +53,35 @@ async function resolveS3InputFromMessage(body: string): Promise<ResolvedS3Input>
...
+    );
+    return null;
+  }
+
+  if (detectedType.mime !== "application/pdf") {
+    log.info(
+      { s3Bucket, s3Key, localPath: downloadedObject.localPath, detectedType },
+      "Skipping UiPath extraction because document is not a PDF"
+    );
+    return null;
```
dustbuster, line 84: PDF checking!

lambdas/UIPath/log.ts
```diff
@@ -3,6 +3,12 @@ import { AsyncLocalStorage } from "node:async_hooks";
 
 const isNotOnAWS = () => !process.env.AWS_EXECUTION_ENV;
 const REDACTED = "[REDACTED]";
+const serializeError = (value: unknown) =>
+  value instanceof Error ? pino.stdSerializers.err(value) : value;
```
dustbuster, line 7: Should fix the issue where the error is not readable in logs.

lambdas/UIPath/log.ts
```diff
@@ -3,6 +3,12 @@ import { AsyncLocalStorage } from "node:async_hooks";
...
+const serializeLogObject = (value: object) =>
+  Object.fromEntries(
+    Object.entries(value).map(([key, entryValue]) => [key, serializeError(entryValue)])
+  );
```
james-becwar-cms, line 11: Just FYI, node has this to serialize a in memory object to a string: https://nodejs.org/api/util.html#utilinspectobject-options

james-becwar-cms (approved):

lambdas/UIPath/log.ts
dustbuster: Thanks! Good find! I checked it out. 
I asked my agent if this could reduce our foot print and serialize data better. 

> Yes, but I would not use util.inspect() as the main fix here.
> 
> It helps turn weird in-memory values like Error, circular objects, or class instances into a readable debug string. But for this Lambda, we want structured JSON logs with redaction. If we inspect(error) and log it as a string, we lose Pino’s path-based redaction and make CloudWatch querying worse.
> The better production path is what we have in log.ts (line 6):
> 
> That fixes the { error: {} } issue because Error.message and Error.stack are non-enumerable, and our formatter was nesting the object under ctx before Pino could naturally serialize it.

PR #1411, opened 2026-06-09
DEMOS-2048: added banner for chip type

connor-parke-icf:
simply provides a little warning if adding the chip type would result in this being considered a chip demo. However, it is a little misleading since the chip id has already been generated in the backend (it is generated on demo creation). Nonetheless, the warning still makes sense since this action reveals it. 

https://github.com/user-attachments/assets/e6f9a5d7-5ed8-4bb6-9aa2-1859ded10752


client/src/components/dialog/DemonstrationTypes/DemonstrationTypesList.test.tsx
```diff
@@ -13,6 +13,15 @@ vi.mock("components/icons", () => ({
...
 }));
 
+vi.mock("components/notice", () => ({
+  Notice: ({ title, description }: { title: string; description: React.ReactNode }) => (
+    <div data-testid="notice">
+      <h2>{title}</h2>
+      {description}
+    </div>
+  ),
+}));
```
tresdonicf: Since this component is rather straightforward it might be better to test the real component here. It might just need a `data-testid` calculation or something to be able to get a better handle on it

tresdonicf (approved):
One suggestion to not mock the notice component but otherwise LGTM!

client/src/components/dialog/DemonstrationTypes/DemonstrationTypesList.test.tsx
connor-parke-icf: hmm, i can see what i can do. without mocking the notice component, it would complain since we're mocking an icon in a different test here, so the notice component would fail to find the icons it needs. But we likely dont need to mock that icon in the first place.

client/src/components/dialog/DemonstrationTypes/DemonstrationTypesList.test.tsx
```diff
@@ -5,23 +5,6 @@ import { describe, it, expect, vi, beforeEach } from "vitest";
...
-vi.mock("components/notice", () => ({
-  Notice: ({ title, description }: { title: string; description: React.ReactNode }) => (
-    <div data-testid="notice">
-      <h2>{title}</h2>
-      {description}
-    </div>
-  ),
-}));
-
 describe("DemonstrationTypesList", () => {
```
connor-parke-icf, line 8: these mocks were indeed not needed! i went ahead and added a testId onto the notice component while i was at it "notice-${variant}"

PR #1412, opened 2026-06-10
DEMOS-2239 & DEMOS-2243: DatePicker Blur Fixes

tresdonicf:
Addresses some issues we're seeing with needing to click off of the DatePicker before its changes are propagated out. This is for things like enabling buttons based on a date or calculating another date. The issue is  some custom `onFocus` and `onBlur` handlers that we don't really need. (introduced in #1299, can kinda see this is just working back those changes)

The design here puts us in a bit of an odd position where we need to sometimes show a different (invalid) value in the datepicker because we show the "Out-of-range" dates. The only way to simplify from here would be to lean on the native datepickers `min / max` properly which is to say we allow it to just clear that invalid value and reset to the last valid value. This way there's no need for a `displayedDate` that is different from the last "committed" value.

https://github.com/user-attachments/assets/83fee222-0165-41d7-8bc4-e5d0f5583362


client/src/components/dialog/DemonstrationTypes/EditDemonstrationTypeDialog.test.tsx
```diff
@@ -200,10 +196,8 @@ describe("EditDemonstrationTypeDialog", () => {
 
       await user.clear(effectiveDateInput);
       await user.type(effectiveDateInput, "2024-12-31");
-      fireEvent.blur(effectiveDateInput);
       await user.clear(expirationDateInput);
       await user.type(expirationDateInput, "2024-01-01");
-      fireEvent.blur(expirationDateInput);
```
tresdonicf, line 206: We don't need to manually fire blur now in these tests, this essentially papered over the bug that was introduced as far as the tests were concerned

client/src/components/input/date/DatePicker.tsx
tresdonicf, line 1: Overall just simplifications here, cleaning up the manual `onFocus` / `onBlur` handlers - feel free to ask about any of these changes

client/src/components/input/date/DatePicker.tsx
```diff
@@ -21,62 +32,48 @@ interface DatePickerProps {
...
-  minDate,
-  maxDate,
-  getValidationMessage,
 }) => {
-  const inputRef = useRef<HTMLInputElement>(null);
-  const isFocusedRef = useRef(false);
-  const inputMin = minDate ?? DEFAULT_MIN_DATE;
-  const inputMax = maxDate ?? DEFAULT_MAX_DATE;
+  // Displayed date is needed to show out of range values in the input while not propagating them to the parent component via onChange.
+  const [displayedDate, setDisplayedDate] = React.useState(value ?? "");
```
tresdonicf, line 37: It would be awesome to not need this, but our current design requires it

client/src/components/input/date/DatePickerSandbox.tsx
tresdonicf, line 1: Added this sandbox component for the demonstration video in the PR

james-becwar-cms (approved):

PR #1414, opened 2026-06-10
DEMOS-2255: updated the back button for deliverabls

dustbuster:
Now you get taken back to the demo you're working in as CMS/admin users

# CMS/ADMIN USERS
Demo back button works correctly. 

https://github.com/user-attachments/assets/cfbf4f1b-e9c0-4510-8639-1160549a2279

Deliverable back button: 

https://github.com/user-attachments/assets/eb96f7f4-5e6c-4df4-af43-adf67826e022

State Users:

### DIFF
--- a/client/src/pages/deliverables/DeliverableDetailsManagementPage.test.tsx
+++ b/client/src/pages/deliverables/DeliverableDetailsManagementPage.test.tsx
@@ -2,7 +2,7 @@ import React from "react";
 import { Route, Routes } from "react-router-dom";
 import { render, screen, waitFor } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
-import { describe, expect, it } from "vitest";
+import { describe, expect, it, vi } from "vitest";
 import {
   DeliverableDetailsManagementPage,
   DELIVERABLE_DETAILS_QUERY,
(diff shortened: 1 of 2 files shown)


client/src/pages/deliverables/DeliverableDetailsManagementPage.test.tsx
```diff
@@ -65,6 +50,28 @@ const buildCurrentUser = (
   person: { ...developmentMockUser.person, personType },
 });
 
+const renderAtRoute = (
+  deliverableId: string,
+  personType: CurrentUser["person"]["personType"] = "demos-cms-user"
```
dustbuster, line 55: Now we actually look at the user type vs. relying on redirect of unauthorized pages.

client/src/pages/deliverables/DeliverableDetailsManagementPage.tsx
```diff
@@ -217,8 +223,17 @@ export const DeliverableDetailsManagementPage: React.FC<{
...
+      const demonstrationId = data?.deliverable?.demonstration?.id;
+      if (!demonstrationId) {
+        throw new Error("Cannot navigate to demonstration deliverables without a demonstration id.");
+      }
+      navigate(`/demonstrations/${demonstrationId}`);
+      return;
+    }
+
+    navigate("/");
+  }, [data?.deliverable?.demonstration?.id, navigate, onBack, userPersonType]);
```
tresdonicf, line 236: I don't think we need `useCallback` here (and could get rid of this dependency array). Could probably just do `const handleBack = () => { ... all this stuff }` Perhaps I'm wrong about this but give it a shot, would be nice to not destructure `data.deliverable.demonstration.id` twice

tresdonicf (approved):
Think we might be able to unnest this from the `useCallback` but otherwise LGTM!

PR #1415, opened 2026-06-10
DEMOS-1819-implement-deliverable-status-report

TARehman:
This PR finally implements the actual deliverable status report. There may be some tweaks to the logic based on back-and-forth with the team; however, I think we should get the bulk of this reviewed because those changes would be tweaks.

Most of the files changed here are actually me adding `sqlfluff` as an option in the `data/` folder. There's not an obvious good way to put `sqlfluff` into action for the code because we're storing the queries in JS variables; I ended up manually running it to help format my code. I am going to look at ways we could automate it later.

This tweaks a few places where we're handling Zod errors to make it easier to figure out what broke by putting it in the logs. The main changes are in the actual report file itself, mostly in the SQL.


james-becwar-cms (approved):

server/src/onDemandReports/configs/deliverableStatusReportQuery.ts
```diff
@@ -1,13 +1,258 @@
...
+    INNER JOIN
+        demos_app.person AS p
+        ON
+            da.user_id = p.id AND da.action_type_id = 'Started Review'
+),
+
+most_recent_valid_bn_data AS (
+    SELECT DISTINCT
+        doc.deliverable_id,
+        first_value(bn.net_variance_total)
```
connor-parke-icf, line 121: the select distinct over on two independent first_value columns feels kinda weird. Wouldnt it be better to select firstValue() from the columns as a unit? feels like it could report unrelated values for each column here

server/src/onDemandReports/configs/deliverableStatusReportQuery.ts
```diff
@@ -1,13 +1,258 @@
...
+    FROM
+        demos_app.deliverable_action
+    WHERE
+        action_type_id = 'Requested Resubmission'
+),
+
+most_recent_public_comment AS (
+    SELECT
+        deliverable_id,
+        'Yes' AS updated_post_acceptance,
```
connor-parke-icf, line 95: this 'Yes' here feels weird, since could a comment be added regardless of the acceptance state? having a comment does not imply that the deliverable has been accepted

server/src/onDemandReports/configs/deliverableStatusReportConfig.ts
```diff
@@ -1,30 +1,114 @@
...
+    ddme_analyst: z.string(),
+    deliverable_type: z.enum(DELIVERABLE_TYPES),
+    deliverable_name: z.string(),
+    demonstration_types: z.string(),
+    cms_owner: z.string(),
+    due_date: usDateString,
+    submission_date: usDateStringOrDash,
+    extension_requested: z.enum(["Yes", "No"]),
+    extension_date_requested: usDateStringOrDash,
+    reason_for_extension: z.enum([...DELIVERABLE_EXTENSION_REASON_CODES, "-"]),
```
connor-parke-icf, line 65: this is powerful. This gives us a way to ensure that our constants align with what is in the database and potentially vice versa; maybe we could introduce a validation script of some kind to select from our limit tables and compare against constant.ts

data/.sqlfluff
```diff
@@ -0,0 +1,32 @@
+[sqlfluff]
```
connor-parke-icf, line 1: good that we have some linting/prettifying on sql now. i dont understand the bulk of this, but i know its something we've wanted for a while. Is there a vscode extension that pairs with this? might be good to throw that in dev container (if/when we extend this to the entire project, not just the data folder)

connor-parke-icf (approved):
Left a couple of comments but not sure how valid; largely the sql is beyond what i can faithfully review in a PR, but i do understand the overall strategy to it. I see the value in the zod validation layer; there a lot going on in here. It also might be able to give us insight of drift between application code and db config. exciting stuff!

server/src/onDemandReports/configs/deliverableStatusReportQuery.ts
TARehman: Yeah, the reason for this is later we do a `coalesce` on this after it's left-joined to the main query. If you left join and nothing comes back, the value is `NULL`, so doing `coalesce(updated_post_acceptance, 'No')` lets you easily get either Yes or No in the right spot. The only risk is that it depends on the join condition, but that's always going to be true.

server/src/onDemandReports/configs/deliverableStatusReportQuery.ts
TARehman: So window functions don't really work that way, though it would be nice. Each one has to be selected individually. I might have been able to do DISTINCT ON here with some tweaking but this is a pretty common pattern.

data/.sqlfluff
TARehman: This is going to be a bigger thing, unfortunately. This doesn't do anything automatically, and because the SQL is actually embedded in TS files as variables, sqlfluff can't do anything with it by default. So I need to spend some time thinking about how to better automate that.

server/src/onDemandReports/configs/deliverableStatusReportConfig.ts
TARehman: Yeah, it wouldn't be hard to write that validator, at least - that's a good thought.

PR #1416, opened 2026-06-10
DEMOS-2218 — Dialog dropdowns get cut off the screen

nunesmlucas:
Custom dropdowns (`AutoCompleteSelect`, `AutoCompleteMultiselect`) render their option
list as a `position: absolute` element. Inside a dialog (`overflow-y-auto`), a list near
the bottom of the modal overflowed and got clipped by the modal edge.

New hook `useDropdownPosition(triggerRef, isOpen)` that, on open (and on scroll/resize)

Both select components now apply the direction (flip classes) and the clamped `maxHeight`.
The list stays `absolute` — it never needs to escape the modal, so it can't be clipped.

The "industry standard" for this is a positioning library like **Floating UI**
  (`@floating-ui/react`) — handles flip/shift/clip/scroll robustly — but it's a new
  dependency and a larger refactor of the shared select components.

is clamp-to-fit good enough, or do we want to invest in
Floating UI for the shared dropdown components?

image

### DIFF
--- a/client/src/components/input/select/AutoCompleteMultiselect.tsx
+++ b/client/src/components/input/select/AutoCompleteMultiselect.tsx
@@ -1,5 +1,6 @@
 import React, { useEffect, useRef, useState } from "react";
 import { ChevronDownIcon } from "components/icons/Symbol/ChevronDownIcon";
+import { useDropdownPosition } from "hooks/useDropdownPosition";
 import { tw } from "tags/tw";
 import { Option } from "./Select";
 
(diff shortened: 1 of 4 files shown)


tresdonicf:
This is more of a polish item than a hair on fire item so it would be nice to shop around for the best solution to maintain long-term and I think this custom approach is a bit too much code - would you mind putting up a PR for the `floating-ui`  approach?

My strong preference would be to solve this using things like the Popover API and the Anchor Positioning API though these are unfortunately a bit less mature web technologies.

FWIW I see that the ticket mentions that it should pop upward but I don't actually think that's needed. I did report this bug and when I saw it I was just thinking we should at least pop the list over the dialog z-index wise but it's not so serious that we need it to expand upward IMO. I'll reach out to Evan about this - but essentially not thinking we want to maintain more than like 10 lines of code for this behavior

tresdonicf (changes requested):
Requesting changes to block merge here while we look at alternatives

PR #1421, opened 2026-06-11
Rename: `formatDate` --> `formatDateForDisplay`

tresdonicf:
Renames our `util` function `formatDate` to `formatDateForDisplay` which is a bit more explicit and should help communicate when this function should be used and avoid clashing with the `date-fns` function of the same name

### DIFF
--- a/client/AGENTS.md
+++ b/client/AGENTS.md
@@ -58,8 +58,12 @@ This file provides instructions for AI agents to use when generating or editing 
 - Co-locate each `gql` document with the component/hook that owns it.
 - Export query/mutation documents if tests, mocks, or `refetchQueries` need them.
 - After successful mutations, refetch affected queries.
-- All outbound dates must use `formatDateForServer` from `util/formatDate`.
-- Inbound dates may need to use `getDateEst(date)`.
+
+### Dates
+
+- The type `LocalDate` is just an alias for `string`.
+- All outbound dates must use `formatDateForServer` from `util/formatDate`. This may require `as LocalDate` on the type.
+- Inbound dates from the server that should display as a date (no time component) should use `getDateEst(dateFromServer)`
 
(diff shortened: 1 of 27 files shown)


client/AGENTS.md
```diff
@@ -58,8 +58,12 @@ This file provides instructions for AI agents to use when generating or editing
...
 - Export query/mutation documents if tests, mocks, or `refetchQueries` need them.
 - After successful mutations, refetch affected queries.
-- All outbound dates must use `formatDateForServer` from `util/formatDate`.
-- Inbound dates may need to use `getDateEst(date)`.
+
+### Dates
+
+- The type `LocalDate` is just an alias for `string`.
+- All outbound dates must use `formatDateForServer` from `util/formatDate`. This may require `as LocalDate` on the type.
+- Inbound dates from the server that should display as a date (no time component) should use `getDateEst(dateFromServer)`
```
connor-parke-icf, line 66: is getDateEst(dateFromServer) meant to be formatDateForDisplay here?

connor-parke-icf (approved):
lgtm! one bonus of these small refactors in a bunch of places is that files that havnt been touched in a while get the updated code formats. one comment about the instruction in agents.md, i dont think we mention the use of formatDateForDisplay in there

client/AGENTS.md
tresdonicf: Good eye! But no this one is correct, it could be `formatDateForDisplay(getDateEst(dateFromServer))` in some cases but this was just an attempt to help the agent understand a bit more how to handle these dates on the F/E

tresdonicf: Added a line here that mentions when it's appropriate to use this utility function (though am hoping that the name change really helps to do some of that lifting as well)

PR #1425, opened 2026-06-12
DEMOS-2252: on tables,  removing horizontal scroll Demos table only so far, and added submission date

dustbuster:
Demonstration Table and Deliverable table 
- "remove horizontal scroll" 
- Also the submission date was missing for all user
- The CMS/Admin had complete query, just needed a function. 
- Add actionTimestamp node for the SU query for submission date 
- Also adjusted the titles and other columns to remove scroll. 
- Min page to show everything is 999px with menu collapsed. Which works on iPad pro. 
- Also added line break in state/territory

A video is worth a bunch of bullets!

https://github.com/user-attachments/assets/b6e8c736-96cd-4db5-8745-3246079a01af

I justified the view button and expander to the right, Title wraps now, i think at the right width, and let the state/terr header wrap if needed to make the col smaller, and everything else is pretty set width like application header that stuff won't change so I set some widths with those and let others wrap but i wanted to avoid the table collapsing into ugly land. but like i said.


client/src/components/table/columns/DemonstrationColumns.tsx
```diff
@@ -18,10 +18,13 @@ export function DemonstrationColumns(projectOfficerOptions: Pick<Person, "fullNa
   return [
     columnHelper.accessor("state.id", {
       id: "stateId",
-      header: "State/Territory",
+      header: "State/\u200BTerritory", // using zero width space for word break
```
dustbuster, line 21: SO this is a zero space space. This will work break when the columns get too small. Lil bit hacky, but we don't have to compromise on UI with "State / Terr.." We can keep it tight and control the break.

client/src/components/table/tables/DeliverableTable.tsx
```diff
@@ -215,6 +218,16 @@ export const formatDeliverableStatus = (
...
 
+export const getSubmissionDate = (
+  deliverableActions: DeliverableTableRow["deliverableActions"]
+) =>
+  deliverableActions
+    .filter((action) => action.actionType === "Submitted Deliverable" && action.actionTimestamp)
+    .sort(
+      (a, b) =>
+        new Date(b.actionTimestamp!).getTime() - new Date(a.actionTimestamp!).getTime()
+    )[0]?.actionTimestamp;
```
dustbuster, line 230: If we got actionType that matches, display it!

client/src/components/table/Table.tsx
```diff
@@ -68,8 +73,9 @@ function TableBody<T>({
     return table.getRowModel().rows.map((row) => (
       <tr key={row.id} className={row.depth > 0 ? STYLES.subrow : STYLES.tr}>
         {row.getVisibleCells().map((cell) => {
+          const meta = cell.column.columnDef.meta as TableColumnMeta | undefined;
```
dustbuster, line 76: This gives us more power over the layout. 
By adding:
`meta?.cellClassName`
`meta?.headerClassName`
`meta?.headerContentClassName`

connor-parke-icf (approved):
looks a whole lot better! It still appears to shrink based on a percentage width, but it also looks like everything is defined with strict min and max widths. Im not entirely sure how that works, css is wild, but regardless looks a lot better. 
At some point post-mvp i want to see if our base tanstack components can be reworked a bit; im not huge into their current control flow. But using the meta var to inject styles seems like a nice solution

dustbuster:
> looks a whole lot better! It still appears to shrink based on a percentage width, but it also looks like everything is defined with strict min and max widths. Im not entirely sure how that works, css is wild, but regardless looks a lot better. At some point post-mvp i want to see if our base tanstack components can be reworked a bit; im not huge into their current control flow. But using the meta var to inject styles seems like a nice solution

yeah it feels like we are cramming logic into a pinhole. Ya dig? but yeah, I'm quite happy with how this turned out. I'm glad you agree! lot of this couple be objective. But i had clear parameters. "I don't want wasted space" once you get to like BIG BIG 4K screen it gets a bit spaced out. idk if there's much i can do about it. 

So like the w-[300px] is like a strong suggestion i think is the best way to put it. My agent spit out a bunch of nonsense so this was all (mostly) me! CSS i get. the Tailwind translation i'm still working on. 

But thanks for the review @connor-parke-icf !

PR #1426, opened 2026-06-12
DEMOS-1817-application-details-report

TARehman:
This is the implementation of the application details report. There are a number of things to note here.

- This actually takes two separate SQL queries being executed. Because of that, I had to refactor the design of the runner slightly. It now takes an array of queries, and returns whatever comes back from the last one. I updated tests and the other reports accordingly.
- The _reason_ this takes two separate queries is because there are a _lot_ of different date types in the database, and I am concerned that if we write a manual pivot for all of them, a small change in a date title might be missed because it would fail silently. To avoid that, I actually pivot the data from `application_date` using `plpgsql`. This grabs the pretty `date_type` values and turns them into SQL variable names, and then builds and executes a query to do the pivot. The results are placed into a temporary table.
  - The value here is that we _also_ use the SQL variable names in the configuration file, to align column names and Zod validations. So if one of the date strings is changed in the DB, the name of the variable in the query will also change - this will cause a Zod validation failure and a loud error, rather than silently letting things pass.
  - I didn't bother with this for the file dates, because there's significantly fewer of those, but it is worth noting that the same risk of a silent error could occur in that context. However, I judged that as a lower risk because a) we don't change the document types with nearly the same frequency as the date types and b) there's only 7 of those, instead of 40-something for the other dates
  - I put the data into a temporary table because in PostgreSQL you can't return from a function if you don't know the shape in advance, which obviously we couldn't know if we're procedurally building the query. You also can't return data directly from a DO block, so the temp table is the best option.


james-becwar-cms (approved):
You may want to put some comments in server/src/onDemandReports/configs/applicationDetailsReportQueries.ts to remind your self what the dynamic sql is doing and why each CTE exists.

connor-parke-icf (approved):
this seems to reflect the same structure as the previous one, will some variations. Its a good call to generate the sql for the dates so we dont have to remember to update this every time we change a date name.  seems like it will be relatively non-invasive to add new reports of any scope now that multiple queries can be utilized.

GonuDvc (commented):
Approach is solid. Minor nit: this would benefit from a brief comment explaining why we skip validation for internal callers.

PR #1427, opened 2026-06-12
DEMOS-2270: made fields required when approved, allowed unsetting sdg

connor-parke-icf:
part 1 of 2.  This covers the demonstration edit modal; the same changes will need to be done to the edit modifications modal. 

Two main changes: 
1) when a demonstration is not approved, the user should be allowed to update the sdg division, effective date, and expiration date freely. This includes unsetting those fields. Support only needed to be added for unsetting the sdg division. 
2) when a demonstration _is_ approved, the sdg division, expiration date, and effective dates should be considered required fields, disallowing their unsetting. 

small cleanups as well. I tried looking at a larger refactor here, cause the create and edit modals have diverged and the introduction of a isApproved variable just adds another dimension to it. But it ended up looking like a rabbithole, so while i definitely think we need a refactor here (separate out create/edit, purge unneeded state variables, control flow concerns), it can probably wait till post mvp. 

https://github.com/user-attachments/assets/09cb374b-f9a7-4a4d-928a-1c532bb66691

### DIFF
--- a/client/src/components/dialog/demonstration/DemonstrationDialog.tsx
+++ b/client/src/components/dialog/demonstration/DemonstrationDialog.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from "react";
+import React, { useState } from "react";
 
 import { BaseDialog } from "components/dialog/BaseDialog";
 import { Textarea } from "components/input";
@@ -9,7 +9,7 @@ import { SelectUsers } from "components/input/select/SelectUsers";
 import { TextInput } from "components/input/TextInput";
 import { Demonstration, SignatureLevel } from "demos-server";
 import { DatePicker } from "components/input/date/DatePicker";
-import { EXPIRATION_DATE_ERROR_MESSAGE } from "util/messages";
+import { EXPIRATION_DATE_ERROR_MESSAGE, getRequiredFieldWhenApprovedMessage } from "util/messages";
 import { SubmitButton } from "components/button/SubmitButton";
 import { isBefore } from "date-fns";
 import { Input } from "components/input/Input";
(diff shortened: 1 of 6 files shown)


client/src/components/dialog/demonstration/DemonstrationDialog.tsx
```diff
@@ -54,44 +54,43 @@ const DemonstrationDescriptionTextArea: React.FC<{
 };
 
 const DateInputs: React.FC<{
+  isApproved: boolean;
   effectiveDate: string;
   expirationDate: string;
   setEffectiveDate: (date: string) => void;
   setExpirationDate: (date: string) => void;
-}> = ({ effectiveDate, expirationDate, setEffectiveDate, setExpirationDate }) => {
-  const [errorMessage, setErrorMessage] = useState<string>("");
```
connor-parke-icf, line 62: a lot of this didnt need to be wrapped in useEffect nor use state variables. So went ahead and trimmed those out

client/src/components/dialog/demonstration/DemonstrationDialog.tsx
```diff
@@ -119,6 +118,12 @@ export const checkFormIsValid = (demonstration: DemonstrationDialogFields) => {
   if (!demonstration.projectOfficerId) {
     return false;
   }
+  if (
```
connor-parke-icf, line 121: if approved, you cannot remove these fields

client/src/components/dialog/demonstration/DemonstrationDialog.tsx
```diff
@@ -174,7 +174,7 @@ export const DemonstrationDialog: React.FC<{
               label="State/Territory"
               value={activeDemonstration.stateId}
               isRequired
-              onSelect={(stateId) => handleChange({ ...activeDemonstration, stateId })}
```
connor-parke-icf, line 177: handleChange was more or less just updating state variables we didnt need and calling the state setter, so just ripped it out and called it directly

client/src/components/dialog/demonstration/DemonstrationDialog.tsx
```diff
@@ -229,20 +232,34 @@ export const DemonstrationDialog: React.FC<{
...
-          />
+          <div className="flex flex-col gap-xs">
+            <SelectSdgDivision
+              isRequired={isApproved}
+              initialValue={activeDemonstration.sdgDivision}
+              onSelect={(sdgDivision) =>
+                setActiveDemonstration({ ...activeDemonstration, sdgDivision })
+              }
+            />
+            {isApproved && !activeDemonstration.sdgDivision && (
```
connor-parke-icf, line 250: added a validation message. Was thinking about copying the validation message pattern we had for the date picker, but id want to take a more dedicated session of experimentation to dive into that

client/src/components/dialog/demonstration/EditDemonstrationDialog.tsx
```diff
@@ -2,60 +2,103 @@ import React from "react";
...
+  | "sdgDivision"
+  | "signatureLevel"
+  | "effectiveDate"
+  | "expirationDate"
+> & {
+  state: Pick<ServerState, "id">;
+  primaryProjectOfficer: Pick<ServerPerson, "id">;
+};
+
+export const GET_DEMONSTRATION_BY_ID_QUERY: TypedDocumentNode<
```
connor-parke-icf, line 38: these didnt have types before, so added them while i was adding a field to the gql

client/src/components/dialog/demonstration/EditDemonstrationDialog.tsx
```diff
@@ -71,7 +114,7 @@ export const getUpdateDemonstrationInput = (
     description: demonstration.description?.trim(),
     effectiveDate: (demonstration.effectiveDate as LocalDate) || null,
     expirationDate: (demonstration.expirationDate as LocalDate) || null,
-    sdgDivision: demonstration.sdgDivision,
+    sdgDivision: demonstration.sdgDivision || null,
```
connor-parke-icf, line 117: main change for the submit was forwarding null if sdgDivision was falsey so that users could unset sdgDivision if they so choose

tresdonicf (approved):

PR #1428, opened 2026-06-12
DEMOS-2259: Fix some `formatDate` handling so modification effectiveDates match

tresdonicf:
Per DEMOS-2259 this PR addresses the issue of the edit modification dialog having an incorrect date between the details section / edit dialog.

https://github.com/user-attachments/assets/f8df779b-12b7-43f0-ba4d-fe8549e4b3ec

### DIFF
--- a/client/src/components/dialog/modification/ModificationForm.tsx
+++ b/client/src/components/dialog/modification/ModificationForm.tsx
@@ -4,7 +4,6 @@ import { SignatureLevel } from "demos-server";
 import { SelectSignatureLevel } from "components/input/select/SelectSignatureLevel";
 import { SelectDemonstration } from "components/input/select/SelectDemonstration";
 import { DatePicker } from "components/input/date/DatePicker";
-import { getDateEst } from "util/formatDate";
 
 export type Modification = {
   id: string;
@@ -34,9 +33,7 @@ export const hasChanges = (
   createModificationFormData: ModificationFormData,
   initialModification: Partial<Modification>
 ): boolean => {
-  const initialEffectiveDate = initialModification.effectiveDate
-    ? getDateEst(new Date(initialModification.effectiveDate))
-    : undefined;
+  const initialEffectiveDate = initialModification.effectiveDate ?? undefined;
 
   return !!(
     createModificationFormData.name != initialModification.name ||
@@ -51,9 +48,7 @@ export const getFormDataFromModification = (
 ): ModificationFormData => ({
   name: modification.name,
   description: modification.description ?? undefined,
-  effectiveDate: modification.effectiveDate
-    ? getDateEst(new Date(modification.effectiveDate))
-    : undefined,
+  effectiveDate: modification.effectiveDate ?? undefined,
   signatureLevel: modification.signatureLevel ?? undefined,
   demonstrationId: modification.demonstration?.id,
 });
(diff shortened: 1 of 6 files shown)


client/src/util/formatDate.fuzz.test.ts
tresdonicf, line 1: Added a bunch of fuzz tests to figure out what the issue was here, these ended up helping see that it was when the date was passed as a string it would behave differently

client/src/util/formatDate.ts
```diff
@@ -1,17 +1,29 @@
...
+
+const formatLocalDateStringForDisplay = (date: LocalDate): string =>
+  format(parse(date, ISO_DATE_FORMAT, new UTCDate()), US_DATE_FORMAT);
+
 export const formatDateForDisplay = (date: DateArgument): string => {
+  if (typeof date === "string") {
+    // LocalDate strings (yyyy-MM-dd) must be parsed in UTC to avoid timezone-induced off-by-one shifts.
+    if (isLocalDateString(date)) {
+      return formatLocalDateStringForDisplay(date as LocalDate);
+    }
```
tresdonicf, line 22: This was the fix, if something is already in ISO format we don't want to re-`parse` it as this will introduce timezone errors, so this path parses it as UTC, avoiding this altogether

client/src/util/formatDate.fuzz.test.ts
connor-parke-icf: love this. hopefully this has highlighted all the weird edge cases we've been dealing with (assuming we're using this method)

client/src/components/dialog/modification/ModificationForm.tsx
```diff
@@ -51,9 +48,7 @@ export const getFormDataFromModification = (
 ): ModificationFormData => ({
   name: modification.name,
   description: modification.description ?? undefined,
-  effectiveDate: modification.effectiveDate
-    ? formatDateForServer(modification.effectiveDate)
-    : undefined,
+  effectiveDate: modification.effectiveDate ?? undefined,
```
connor-parke-icf, line 51: guessing this can be removed because the server handles direct string parsing on its own, so we can just send it the field contents without reformatting?

connor-parke-icf (approved):
lgtm! bout time we had a comprehensive test suite for a bunch of date scenarios.  Its kinda odd, in most cases im sure it "just works" but because we want things to be reflected as DC time, our use case might be outside the realm of typical usage of most date pickers/libraries. Hopefully this addressed all the weird edge cases

client/src/components/dialog/modification/ModificationForm.tsx
tresdonicf: Yeah or this already is even in `yyyy-mm-dd` so it was also being re-parsed, might need to adjust that function to short circuit in that case though (kinda like I did for `formatDateForDisplay`)

client/src/util/formatDate.fuzz.test.ts
tresdonicf: It was super helpful! And yeah this certainly needs a little bit of an audit and maybe some thoughtful "pass-through" behavior so we can call these functions more regularly with different behavior depending on the type / shape

Though at this point the best fix seems to be re-examining the storage of the dates with the timestamp on the server so we just avoid this altogether

PR #1443, opened 2026-06-15
DEMOS-2251 add submitted date calculation

JeffreyM-GA:
Adds submitted date to the deliverable info page. Had to modify the source function for the table to make the input types match

image


client/src/components/table/tables/DeliverableTable.tsx
```diff
@@ -219,15 +221,22 @@ export const formatDeliverableStatus = (
...
+  const submissions = deliverableActions.filter(
+    (action) => action.actionType === "Submitted Deliverable"
+  );
+
+  if (submissions.length === 0) {
+    return undefined;
+  }
+
+  submissions
+    .sort((a, b) => isAfter(a.actionTimestamp, b.actionTimestamp) ? -1 : 1);
```
connor-parke-icf: might can use compareAsc here, but it would require a mapping first anyway, so may not be needed. Not sure if not having a 0 able to be returned here is an issue or not; guessing it doesnt matter.

client/src/components/table/columns/dateColumn.tsx
```diff
@@ -10,18 +10,19 @@ type DateFilterValue = {
...
   return columnHelper.accessor((row: RowData) => row[fieldName], {
     id: fieldName,
     header,
     cell: ({ getValue }) => {
       const value = getValue() as string | undefined;
-      return value ? formatDateForDisplay(value) : "";
+      return value ? formatDateForDisplay(value) : defaultValue;
     },
     filterFn: (row, columnId, filterValue: DateFilterValue) => {
       const value = row.getValue(columnId) as string | undefined;
```
connor-parke-icf, line 24: does row.getValue here ever return the translated value, or does it pull from the core object? i suspect value can never = defaultValue unless in the base data we send in the empty string or dash or whatever.

But if it calls the cell function, then this may be needed

client/src/components/table/columns/DeliverableColumns.tsx
```diff
@@ -76,7 +76,8 @@ export function DeliverableColumns({
   const submissionDateColumn = createDateColumnDef(
     columnHelper,
     "submissionDate",
-    "Submission Date"
+    "Submission Date",
+    "-"
```
connor-parke-icf, line 80: i do like this convention of supplying a default value in the column def itself instead of in the cell function. moves the configuration up a bit

client/src/pages/deliverables/sections/DeliverableInfoFields.tsx
```diff
@@ -49,7 +52,7 @@ export const DeliverableInfoFields = ({
   const baseFields: DeliverableInfoField[] = [
     { label: "Deliverable Type", value: deliverable.deliverableType },
     { label: "Due Date", value: formatDateForDisplay(deliverable.dueDate) },
-    { label: "Submission Date", value: "—" },
+    { label: "Submission Date", value: submissionDate ?? "—" },
```
connor-parke-icf, line 55: hmm... it may be best to pass undefined here instead of translating it to dash. Then in the filtering/sorting we can ignore falsey values then do the rendering as a dash closer to the html. Then we also dont need to keep track of which value is the defaultValue if we know that we can treat all undefined values as showing dash

connor-parke-icf (approved):
I dont think anything blocking, but i think that we may be able to simplify some of the logic regarding keeping track of which value is the default value if we pass undefined or some other falsey into the table for the submitted date instead of transforming it to "-" early.  This might make it so that we dont have to check whether the value equals the default value when filtering/sorting. Other than that though, lgtm!

client/src/components/table/columns/dateColumn.tsx
JeffreyM-GA: Yep, you're right, it uses the object value.

client/src/pages/deliverables/sections/DeliverableInfoFields.tsx
JeffreyM-GA: This one is actually just for a label display in the info section rather than in a table, so sorting/filtering shouldn't be a big concern I think? For the tables I used the createDateColumnDef with the default value as you saw earlier.

PR #1444, opened 2026-06-15
Demos 1489 created tool for copying demonstrations

connor-parke-icf:
This is a tool developed for the purposes of cloning a demonstration. With this, a test/template demonstration can be created on frontend then copied to either the same environment or a different environment an arbitrary amount of times. This was used to populate 10 test demonstrations for the UAT going on today, so figured it might as well make it into the codebase officially. 

caveats:
* because demonstrations are dependent on the users in the system, rather than initialize users that dont exist, this simply requires that a predefined State User and CMS User exist with explicit usernames. there are likely some edge cases here that are unaddressed, and currently the script is hardcoded to use David's test CMS and State Users. this is controlled by the block at line 206 of import.sql. 
* If the demonstration being copied belongs to a state that the state user does not belong, the state user will be assigned to that state. 
* for this reason, the contacts tab is not copied 1 to 1. 
* Documents are copied both in database and s3 contents. However, the documents are copied locally to the script runner's machine.  This means **_this script should not be used to copy real or production data_** (among other reasons). 

It works by running a psgl script, outputting the results to individual csv files.  then the csv files are processed and read back to the target environment. This means that if we wanted to keep a copy of a demonstration long-term, we just need to save the collection of csvs. after that, it loops through the found s3 files and performs a copy for each. During the processing, user ids and and references to global data are dealt with, either by replacing with predefined targets, upserting, or on conflict do nothing.


data/uat_scenario/.gitignore
```diff
@@ -0,0 +1,2 @@
+output/*
+!output/
```
TARehman:
```suggestion
!output/

```

data/uat_scenario/copy_s3_paths_localstack.sh
```diff
@@ -0,0 +1,293 @@
...
+  fi
+
+  copy_count=$((copy_count + 1))
+done < "$mapping_file"
+
+if [[ "$dry_run" == "true" ]]; then
+  echo "Prepared $copy_count copy commands; skipped $skip_count rows."
+else
+  echo "Completed $copy_count copies; skipped $skip_count rows."
+fi
```
TARehman:
```suggestion
fi

```

data/uat_scenario/run_clone.sh
```diff
@@ -0,0 +1,142 @@
...
+  -f import.sql
+
+if [[ "$copy_s3" == "true" ]]; then
+  # Step 4: replay the document S3 object copy using the generated key mapping. Any
+  # real-AWS flags are simply forwarded to the lower-level S3 copy helper.
+  echo "Copying document objects from source S3 to target S3"
+  "$script_dir/copy_s3_paths_localstack.sh" --mapping-file output/s3_path_mapping.csv "${s3_args[@]}"
+fi
+
+echo "Done. UUID map written to output/uuid_remap.json and s3 mapping written to output/s3_path_mapping.csv"
```
TARehman:
```suggestion
echo "Done. UUID map written to output/uuid_remap.json and s3 mapping written to output/s3_path_mapping.csv"

```

data/uat_scenario/copy_s3_paths_localstack.sh
```diff
@@ -0,0 +1,293 @@
+#!/usr/bin/env bash
```
TARehman, line 1: FWIW, this is complex enough that at a certain point I might have moved from `bash` to something a bit more full-featured like `python`. That being said, there's nothing inherently wrong with bash, its just any time my bash scripts are starting to be this complex, I yearn for a real language. Especially since you were using `python` elsewhere. 🙂

data/uat_scenario/remap_ids.py
```diff
@@ -0,0 +1,173 @@
...
+    with s3_mapping_path.open("w", newline="", encoding="utf-8") as handle:
+        writer = csv.DictWriter(handle, fieldnames=["old_s3_path", "new_s3_path"])
+        writer.writeheader()
+        writer.writerows(s3_mappings)
+
+    return 0
+
+
+if __name__ == "__main__":
+    raise SystemExit(main())
```
TARehman:
```suggestion
    raise SystemExit(main())

```

data/uat_scenario/copy_s3_paths_localstack.sh
```diff
@@ -0,0 +1,293 @@
...
+  --real-aws            Do not pass endpoint URLs for source or target operations
+  --temp-dir PATH       Temp directory for object transfer files. Default: mktemp -d
+  --dry-run             Print aws s3 cp commands without executing them
+  --help                Show this help
+EOF
+}
+
+# Build one AWS CLI argument array per side so source and target can use different
+# regions, credentials, and endpoint behavior.
+build_aws_cmd() {
```
TARehman, line 46: It feels vaguely wrong to do this where we mutate a variable inside a function scope, but I guess its standard in `bash`?

TARehman (approved):
Some very nitpicky suggests to add newlines to things. @connor-parke-icf and I talked about this offline. This is fairly brittle but also has a very specific purpose that won't be relevant after we launch, and can be rewritten to be more elegant at that point. Personally, I am offended on a spiritual level by the inclusion of tabs in the SQL file but in the interests of ecumenicalism I'm going to let it slide. 🤪 This is already in use / has been used, so I'm approving because we are just documenting what was done.

data/uat_scenario/copy_s3_paths_localstack.sh
connor-parke-icf: agreed; given more time post mvp this could serve as a great tool for debugging and what not. But it absolutely should be rewritten in a real framework. Id prefer a little npm script to keep it in the same language and so like TS can validate some things if we reference the server

data/uat_scenario/copy_s3_paths_localstack.sh
connor-parke-icf: no idea the standard conventions when it comes to bash, or if there are any. definitely not my language of choice for anything i need to look at more than three times

PR #1445, opened 2026-06-15
DEMOS-2271: made tab selection default from query param

connor-parke-icf:
This change makes it so the demonstration details page loads with the amendment/extension preselected that has been passed in as a query param. if its not found, or otherwise doesnt exist on the list of amendments or extensions, it defaults to the first item. Just passes along the query param it gets at the base load of the details page. 

https://github.com/user-attachments/assets/8f2528ff-60f3-4fef-874f-e23b1e25797c

### DIFF
--- a/client/src/pages/DemonstrationDetail/DemonstrationDetail.tsx
+++ b/client/src/pages/DemonstrationDetail/DemonstrationDetail.tsx
@@ -217,6 +217,7 @@ export const DemonstrationDetail: React.FC = () => {
                 demonstrationId={demonstration.id}
                 medicaidId={demonstration.medicaidId}
                 amendments={demonstration.amendments}
+                selectedAmendmentId={amendmentParam || undefined}
               />
             </Tab>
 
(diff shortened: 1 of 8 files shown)


client/src/pages/DemonstrationDetail/modifications/AmendmentsTab.test.tsx
```diff
@@ -10,13 +11,24 @@ vi.mock("components/dialog/DialogContext", () => ({
   }),
 }));
 
+vi.mock("./ModificationTabs", () => ({
+  ModificationTabs: vi.fn(() => <div data-testid="modification-tabs">Modification Tabs</div>),
+}));
```
connor-parke-icf, line 16: mocking since this component would just pass through the selected item id, and theres no visual indicator of that id on the page other than in conjunction with the tabs, which is already tested elsewhere. So just validating that the thing was called correctly

client/src/pages/DemonstrationDetail/DemonstrationDetail.tsx
```diff
@@ -229,6 +230,7 @@ export const DemonstrationDetail: React.FC = () => {
                 demonstrationId={demonstration.id}
                 medicaidId={demonstration.medicaidId}
                 extensions={demonstration.extensions}
+                selectedExtensionId={extensionParam || undefined}
```
connor-parke-icf, line 233: feels weird to cast null to undefined, but the param semantically makes more sense as undefined than null to me /shrug

james-becwar-cms (approved):

PR #1446, opened 2026-06-15
DEMOS-2221 — Comments panel spacing/alignment

nunesmlucas:
The Comments panel stretched to fill the full parent height, and the textarea section absorbed the leftover space — leaving a large gap between the **Add Comment** button and **Comment History**.

## Fix
- Removed `min-h-full` from the panel container → panel is now auto-height and wraps its content.
- Removed `flex-1` from the textarea wrapper → input section sizes to its content instead of expanding.
- Added `self-start` to the panel's wrapper `<div>` so it hugs content height instead of being stretched by the flex row.

image

### DIFF
--- a/client/src/pages/deliverables/DeliverableDetailsManagementPage.tsx
+++ b/client/src/pages/deliverables/DeliverableDetailsManagementPage.tsx
@@ -328,7 +328,7 @@ export const DeliverableDetailsManagementPage: React.FC<{
           <div className="flex-1">
             <FileAndHistoryTabs deliverable={data.deliverable} />
           </div>
-          <div>
+          <div className="self-start">
             <CommentBox deliverableId={data.deliverable.id} />
           </div>
         </div>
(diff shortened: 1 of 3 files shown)


dustbuster (approved):
Looks great!

client/src/pages/deliverables/DeliverableDetailsManagementPage.tsx
```diff
@@ -328,7 +328,7 @@ export const DeliverableDetailsManagementPage: React.FC<{
           <div className="flex-1">
             <FileAndHistoryTabs deliverable={data.deliverable} />
           </div>
-          <div>
+          <div className="self-start">
```
connor-parke-icf, line 331: interesting! never seen the self-start, but looking at it it could be pretty useful. Ran into some odd looking alignment of icons and text which this may come in handy for

PR #1454, opened 2026-06-16
Demos 2198 application intake phase adjustments

connor-parke-icf:
Adjustments in wording for the Application Intake phase. As i was going through this, i found some inconsistencies of styles between this phase and the others. I think this is because we're defining a set of styles for each phase rather than drawing on a common style. Its too big a task for the moment, so  ive stuck largely to just updating the text (and moving the finish button, as specified in the ticket), but it would be fantastic to get all of the phase style updates and knock them all out at once with a refactor. 

image

### DIFF
--- a/client/src/components/tags/ApplicationHealthTypeTags.tsx
+++ b/client/src/components/tags/ApplicationHealthTypeTags.tsx
@@ -31,8 +31,6 @@ export const GET_APPLICATION_TAG_OPTIONS: TypedDocumentNode<
 export interface ApplicationHealthTypeTagsProps {
   applicationId: string;
   selectedTags: Tag[];
-  title: string;
-  description?: string;
   suggestedTags?: TagName[];
   onRemoveTag: (tag: string) => void;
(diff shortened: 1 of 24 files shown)


client/src/components/tags/ApplicationHealthTypeTags.tsx
```diff
@@ -65,11 +61,7 @@ export const ApplicationHealthTypeTags = ({
   };
 
   return (
-    <div aria-labelledby="state-application-tags-title">
-      <h4 id="state-application-tags-title" className={STYLES.stepThree}>
-        {title}
-      </h4>
-      {description && description.trim() !== "" && <p className={STYLES.helper}>{description}</p>}
```
connor-parke-icf, line 72: moved the title and description sections to the applicationIntakePhase to keep the main layout features there, leaving the listing and suggestions here

james-becwar-cms (approved):

nunesmlucas (approved):

PR #1458, opened 2026-06-16
DEMOS-2200: updated text for federal comment phase

connor-parke-icf:
yet more text updates. Trying to limit my scope as i go, but have a number of ideas to try and wrangle the diverging patterns/styles in these phases. Will relegate to a refactor item of its own. 
image

### DIFF
--- a/client/src/components/application/phase-selector/PhaseSelector.test.tsx
+++ b/client/src/components/application/phase-selector/PhaseSelector.test.tsx
@@ -3,7 +3,7 @@ import React from "react";
 import { TestProvider } from "test-utils/TestProvider";
 import { describe, expect, it, vi } from "vitest";
 
-import { render, screen, waitFor } from "@testing-library/react";
+import { render, screen, waitFor, within } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 
 import { PhaseSelector, getDisplayedPhaseStatus, getDisplayedPhaseDate } from "./PhaseSelector";
@@ -65,6 +65,8 @@ describe("PhaseSelector", () => {
         <PhaseSelector application={demonstration} workflowApplicationType="demonstration" />
       </TestProvider>
     );
+    const phaseSelectorGrid = screen.getByTestId("phase-selector-grid");
+
     [
       "Concept",
       "Application Intake",
@@ -75,7 +77,7 @@ describe("PhaseSelector", () => {
       "Approval Package",
       "Approval Summary",
     ].forEach((name) => {
-      expect(screen.getByText(name)).toBeInTheDocument();
+      expect(within(phaseSelectorGrid).getByText(name)).toBeInTheDocument();
     });
   });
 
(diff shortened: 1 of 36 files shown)


client/src/components/application/phase-selector/PhaseSelector.test.tsx
```diff
@@ -75,7 +77,7 @@ describe("PhaseSelector", () => {
       "Approval Package",
       "Approval Summary",
     ].forEach((name) => {
-      expect(screen.getByText(name)).toBeInTheDocument();
+      expect(within(phaseSelectorGrid).getByText(name)).toBeInTheDocument();
```
connor-parke-icf, line 85: because federal comment phase was being used as the initial selected phase, it would render two on the page. So i just wrapped the grid in a testid and used within to filter the search.

client/src/components/application/phases/FederalCommentPhase.tsx
```diff
@@ -138,27 +156,24 @@ export const FederalCommentPhase: React.FC<FederalCommentPhaseProps> = ({
...
 
       <section className={STYLES.pane}>
         <div className={STYLES.grid}>
           <span aria-hidden className={STYLES.divider} />
           <VerifyCompleteSection />
           <ApplicationUploadSection
-            title="STEP 2 - UPLOAD"
-            helperText="Upload the Internal Analysis Document (Optional)"
+            title="Step 2 - Upload"
+            helperText={FEDERAL_COMMENT_PHASE_STEP_TWO_DESCRIPTION.text}
```
connor-parke-icf, line 176: i really like this component for the individual upload section. Would love to, as a future refactor, adjust the other phases to utilize this; didnt know we had it.

client/src/components/application/phase-selector/PhaseSelector.tsx
```diff
@@ -139,7 +139,7 @@ export const PhaseSelector = ({
 
   return (
     <>
-      <div className="grid grid-cols-8 gap-md mb-2">
+      <div className="grid grid-cols-8 gap-md mb-2" data-testid="phase-selector-grid">
```
tresdonicf: Always a nice enhancment! I would recommend while you're at it throwing this in an `export const` at the top of the file and using that so it's not two magic strings that need to match up but one code symbol  that can be re-used

client/src/components/application/phases/FederalCommentPhase.tsx
tresdonicf: FWIW I don't know that this specific component will be "generalizable" to the point of re-using the actual component but the pattern of creating larger named "sections" instead of a single huge component with a bunch of inline `<div>`s is super desirable and worth improving the codebase towards

tresdonicf (approved):

client/src/components/application/phases/FederalCommentPhase.tsx
connor-parke-icf: hm... that may be the case. torn between just moving its content to the federal comment phase file (since this is the only place that uses it) or try to make it somewhat generalized so it can be used elsewhere.  I definitely have cornered myself with over-generalization before haha.  Either way; its a later decision, once we have some space to do some more general cleanup/refactor we can revisit

client/src/components/application/phase-selector/PhaseSelector.tsx
connor-parke-icf: ah good catch, will do!

PR #1465, opened 2026-06-17
DEMOS-2287: added basic support for new date type

connor-parke-icf:
Covers Tickets https://github.com/Enterprise-CMCS/demos/pull/1465#pullrequestreview-4517788961 and https://jiraent.cms.gov/browse/DEMOS-2290

This adds a new date type "Application Approval Date" to the database, API, Reporting, and frontend on the first section of phase 8. This date must be non-null for the application to be approved and exists for All three of Demonstrations, Amendments, and Extensions. Largely, works the same as any other Application Date. 

Because the only place this field can be edited is on phase 8 itself, its logic diverges a little bit from the other fields to allow for updating it, until the section is readonly from approval; one can still "uncomplete" the section and resave, up until the application itself is approved. 

https://github.com/user-attachments/assets/4f8b5e0b-cdcb-4b1d-898e-a8bab80fddce


client/src/components/application/phases/approval-summary/ApprovalSummaryPhase.tsx
```diff
@@ -94,6 +94,10 @@ const getReadonlyFields = (data: ApplicationDetailsFormData) => ({
 export const getDemonstrationApprovalSummaryFormData = (
   demonstration: ApplicationWorkflowDemonstration
 ): ApplicationDetailsFormData => {
+  const applicationApprovalDate = demonstration.phases
```
connor-parke-icf, line 97: as i have been thinking about this, perhaps instead of an array of Phases the API could return an object with a field for every phase. Then we wouldnt have to do all this phase hunting stuff, we could just query for the phase we want... refactor wishlist perhaps.

client/src/components/application/phases/approval-summary/ApprovalSummaryPhase.tsx
```diff
@@ -111,6 +115,9 @@ export const getDemonstrationApprovalSummaryFormData = (
     description: demonstration.description,
     sdgDivision: demonstration.sdgDivision,
     signatureLevel: demonstration.signatureLevel,
+    applicationApprovalDate: applicationApprovalDate
+      ? (formatDateForServer(applicationApprovalDate) as LocalDate)
```
connor-parke-icf, line 119: kinda goes with what we discussed, but formatting the date here is kinda awkward since it comes in as a iso date, then we need to format it as a LocalDate for the datepicker, but also format it as a display date for the text field version. This issue doesnt appear for the other dates on this page because theyre never actually used in the YYYY-mm-dd format; if they exist already they will be displayed in plaintext. 

Because this field is _only_ used on this page, i opted to make it editable so long as the section is not complete, nor the phase complete. So one could still "uncomplete" the section and edit it if they wanted to or made a mistake.

client/src/components/application/phases/approval-summary/ApprovalSummaryPhase.tsx
```diff
@@ -330,6 +344,12 @@ export const ApprovalSummaryPhase = ({
         projectOfficerUserId: formData.projectOfficerId,
       };
 
+      await setApplicationDate({
```
connor-parke-icf, line 347: not a great way to transactionalize this without major changes to the API. However, it aligns with how we set the completed date when completing the tab so i figure its fine

server/src/onDemandReports/configs/demonstrationOverviewReportQueries.ts
```diff
@@ -177,7 +177,7 @@ LEFT JOIN
     demos_app.application_date AS app_date
     ON
         demo.id = app_date.application_id
-        AND app_date.date_type_id = 'Approval Summary Completion Date';
+        AND app_date.date_type_id = 'Application Approval Date';
```
connor-parke-icf, line 180: everything else in these reports was already completed; just needed to hotswap the placeholder date with this new one.

james-becwar-cms (approved):

PR #1467, opened 2026-06-18
Demos 2314 remove ability to edit document type

connor-parke-icf:
We are moving the ability to edit the documentType across the application. 
For the backend, the mutators/resolvers/validation was updated to not allow documentTypes to be changed. 

The frontend was a bit trickier, but ultimately decided that since the edit document dialog only contained two scalar fields, it no longer needed to be combined with the default DocumentDialog. So the edit parts of the DocumentDialog were trimmed away and moved into the EditDocumentDialog.  In doing so, a number of props could be removed from the DocumentDialog and the signatures of both reduced. 

https://github.com/user-attachments/assets/41363402-3164-4568-977e-f76c9ee3c9ab

### DIFF
--- a/client/src/components/dialog/document/DocumentDialog.tsx
+++ b/client/src/components/dialog/document/DocumentDialog.tsx
@@ -17,8 +17,6 @@ import { UploadButton } from "./UploadButton";
 import { BNPreValidationState, useBNWorkbookPreValidation } from "./useBNWorkbookPreValidation";
 import { DocumentChip } from "components/document/documentChip";
 
-type DocumentDialogType = "add" | "edit";
-
 // BN Workbook uploads require the user to attest to the content before submitting.
 const ATTESTATION_DOCUMENT_TYPES: DocumentType[] = ["BN Workbook"];
 
@@ -69,7 +67,7 @@ const SUCCESS_MESSAGES = {
   fileDeleted: "Your document has been removed.",
 };
 
-const TitleInput: React.FC<{ value: string; onChange: (value: string) => void }> = ({
+export const TitleInput: React.FC<{ value: string; onChange: (value: string) => void }> = ({
   value,
   onChange,
 }) => (
@@ -89,7 +87,7 @@ type DescriptionInputProps = {
   error?: string;
 };
 
-const DescriptionInput: React.FC<DescriptionInputProps> = ({ value, onChange, error }) => {
+export const DescriptionInput: React.FC<DescriptionInputProps> = ({ value, onChange, error }) => {
   const validationMessage = error ?? "";
   return (
     <div className="flex flex-col gap-sm">
@@ -315,14 +313,11 @@ export const checkFormHasChanges = (
 
 export type DocumentDialogProps = {
   onClose?: () => void;
-  mode: DocumentDialogType;
-  documentTypeSubset?: DocumentType[];
+  applicableDocumentTypes?: DocumentType[];
   onSubmit: (dialogFields: DocumentDialogFields) => Promise<DocumentUploadResult>;
   initialDocument?: DocumentDialogFields;
   titleOverride?: string;
   cancelButtonIsDisabled?: boolean;
-  canEditDocumentType?: boolean;
-  hideDocumentType?: boolean;
 };
 
 // Sets the default document type if a subset is provided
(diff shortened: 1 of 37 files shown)


server/src/model/documentPendingUpload/documentPendingUploadSchema.ts
```diff
@@ -47,12 +47,6 @@ export const documentPendingUploadSchema = gql`
     deliverableId: ID!
   }
 
-  input UpdateDocumentInput {
-    name: NonEmptyString
-    description: String
-    documentType: DocumentType
-  }
```
connor-parke-icf, line 54: this is not supposed to be here, probably leftover from documentPendingUpload refactors

client/src/components/dialog/document/EditDocumentDialog.tsx
```diff
@@ -1,64 +1,100 @@
...
+export const hasChanges = (initialDocument: Document, activeDocument: Document): boolean => {
+  return (
+    activeDocument.name !== initialDocument.name ||
+    activeDocument.description !== initialDocument.description
+  );
+};
+
+export const isValid = (document: Document): boolean => !!document.name.trim();
+
 export const EditDocumentDialog: React.FC<{
```
connor-parke-icf, line 37: with the removal of all things documentType in the editDocumentDialog, it had diverged from the default DocumentDialog enough to warrant just breaking it out into its own component. Now its just a simple dialog with two fields.

client/src/components/dialog/document/DocumentDialog.tsx
```diff
@@ -339,18 +334,15 @@ const setDefaultDocumentType = (
 
 export const DocumentDialog: React.FC<DocumentDialogProps> = ({
   onClose = () => {},
-  mode,
-  documentTypeSubset,
+  applicableDocumentTypes,
```
connor-parke-icf, line 337: because this prop was being treated as an override rather than a filter, i changed the name to reflect that

client/src/components/dialog/document/DocumentDialog.tsx
```diff
@@ -339,18 +334,15 @@ const setDefaultDocumentType = (
...
   onSubmit,
   initialDocument,
   titleOverride,
-  canEditDocumentType = true,
-  hideDocumentType = false,
```
connor-parke-icf, line 348: the hideDocumentType prop was only being used in the case of an edit instance, so because edits dont show documentTypes anyway, this could safely be removed. The canEditDocumentType was also moved to be dependent on the number of documentTypes input to the field itself: if theres only one option, its disabled.

client/src/components/dialog/document/EditDocumentDialog.tsx
tresdonicf, line 1: Been waiting a long time for this! Great to see this broken out

client/src/components/dialog/document/DocumentDialog.tsx
tresdonicf, line 1: NB: It might be worthwhile to rename this to `AddDocumentDialog` now since that's the only case it addresses now. And later down the line we can probably move the components into their own folder here to make the sharing a bit more "shared"

tresdonicf (approved):
Looks good!

client/src/components/dialog/document/DocumentDialog.tsx
dustbuster: I feel like this is a long time coming. We have an Edit and RemoveDocumentDialog too. But i can see where it could be out of scope.

client/src/components/dialog/document/DocumentDialog.tsx
```diff
@@ -461,23 +450,19 @@ export const DocumentDialog: React.FC<DocumentDialogProps> = ({
...
-            file={file}
-            onRemove={() => setFile(null)}
-            fileInputRef={fileInputRef}
-            uploadStatus={uploadStatus}
-            uploadProgress={uploadProgress}
-            handleFileChange={handleFileChange}
-          />
-        ) : (
-          ""
-        )}
```
dustbuster, line 480: Nice!

client/src/components/dialog/document/DocumentDialog.tsx
connor-parke-icf: hmm... i do want to, but im afraid it'd blow up the git diff too much for this PR. And i dont want to rename the DocumentDialog to the AddDocumentDialog while the EditDocumentDialog is still pulling from it (its not quite enough of a "sibling" component imo).  So im going to leave it as is for now, but definitely a great suggestion and an easy win for the refactor wishlist

PR #1468, opened 2026-06-18
Demos 2305 allow deletion of unsubmitted documents

connor-parke-icf:
Documents should be able to be deleted until they are submitted, regardless of the status of the deliverable. 
Splitting this ticket into two parts in order to keep the reviewable chunks small.  
This part includes:
Backend was adjusted to only check for the submission status of the document itself. 
Frontend was adjusted to make the document tables' delete button only active if none of the selected documents are part of a deliverable submission. 

https://github.com/user-attachments/assets/cc73659c-c4f1-4093-9c35-6216f7f9b94a

The next part will include a rework of the disabled flags, as currently they overencapsulate some logic. For instance, these flags are still preventing the deletion of documents in the finalized states, when that should be covered by the document's deliverable submission status. this should be trimmed up to keep the logic consistent.


tresdonicf (approved):
LGTM!

james-becwar-cms (approved):

PR #1469, opened 2026-06-18
DEMOS-2286: componentized the "create new menu" and moved it to only the demos table 

dustbuster:
In DefaultHeaderLower there's the createNew buttons. They inc the create demo ext and amendment buttons, but this is on the wrong line in the header. Was able to reduce the header lower page and moved to Header file

Screenshot 2026-06-18 at 14 22 51

### DIFF
--- a/client/src/components/header/CreateNewButton.tsx
+++ b/client/src/components/header/CreateNewButton.tsx
@@ -1,0 +1,78 @@
+import React, { useEffect, useRef, useState } from "react";
+
+import { IconButton } from "components/button";
+import { AddNewIcon } from "components/icons";
+import { PersonType } from "demos-server";
+import { getCurrentUser } from "components/user/UserContext";
+import { useDialog } from "components/dialog/DialogContext";
+
+const CREATE_PERSON_TYPES: ReadonlySet<PersonType> = new Set(["demos-admin", "demos-cms-user"]);
+
+export const CreateNewButton: React.FC = () => {
+  const [showDropdown, setShowDropdown] = useState(false);
+  const dropdownRef = useRef<HTMLDivElement>(null);
+  const { currentUser } = getCurrentUser();
+  const { showCreateDemonstrationDialog, showCreateAmendmentDialog, showCreateExtensionDialog } =
+    useDialog();
+
+  useEffect(() => {
+    const handleClickOutside = (e: MouseEvent) => {
+      if (dropdownRef.current && !dropdownRef.current.contains(e.target as Node)) {
+        setShowDropdown(false);
+      }
+    };
+    document.addEventListener("mousedown", handleClickOutside);
+    return () => document.removeEventListener("mousedown", handleClickOutside);
+  }, []);
+
+  if (!CREATE_PERSON_TYPES.has(currentUser.person.personType)) {
+    return null;
+  }
+
+  return (
+    <div ref={dropdownRef} className="relative">
+      <IconButton
+        icon={<AddNewIcon />}
+        name="create-new"
+        onClick={() => setShowDropdown((prev) => !prev)}
+      >
+        Create New
+      </IconButton>
+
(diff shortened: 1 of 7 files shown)


client/src/components/header/CreateNewButton.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+  return (
+    <div ref={dropdownRef} className="relative">
+      <IconButton
+        icon={<AddNewIcon />}
+        name="create-new"
+        onClick={() => setShowDropdown((prev) => !prev)}
+      >
+        Create New
+      </IconButton>
```
dustbuster, line 40: At first i saw this, And was no big deal. 
But really the entirety of DefaultHeaderLower was the button. not thinking about the action of the button, But really it's a menu. 
I could even see where we'd call this CreateNewMenu. But pretty sure we refer to it as a button and this is why i named it as such. 

But in the need this is pretty novel functionality and it think it make sense as it's own Component. 
Image

client/src/pages/DemonstrationsPage.tsx
```diff
@@ -82,16 +82,14 @@ export const DemonstrationsPage: React.FC = () => {
 
   return (
     <div className="shadow-md bg-white p-[16px]">
-      <h1 className="text-[20px] font-bold mb-[24px] text-brand uppercase border-b-1 pb-[8px]">
-        Demonstrations
-      </h1>
+      <div className="flex items-center justify-between mb-[24px] border-b-1 pb-[8px]">
+        <h1 className="text-[20px] font-bold text-brand uppercase">Demonstrations</h1>
+        <CreateNewButton />
```
dustbuster, line 87: Still need to do some checking around and see all the pages this lands on.

client/src/components/header/DefaultHeaderLower.test.tsx
```diff
@@ -67,66 +31,13 @@ describe("DefaultHeaderLower", () => {
...
-        person: {
-          ...mockUsers[0].person,
-          personType: "demos-state-user",
-        },
-      },
-    });
 
     render(<DefaultHeaderLower />);
 
     expect(screen.queryByText("Create New")).not.toBeInTheDocument();
```
connor-parke-icf, line 41: i suspect we can just remove this test altogether, since there are many things that could not be in the document that would make this test pass

client/src/components/header/CreateNewButton.tsx
```diff
@@ -0,0 +1,78 @@
...
+    const handleClickOutside = (e: MouseEvent) => {
+      if (dropdownRef.current && !dropdownRef.current.contains(e.target as Node)) {
+        setShowDropdown(false);
+      }
+    };
+    document.addEventListener("mousedown", handleClickOutside);
+    return () => document.removeEventListener("mousedown", handleClickOutside);
+  }, []);
+
+  if (!CREATE_PERSON_TYPES.has(currentUser.person.personType)) {
```
connor-parke-icf, line 28: NB: im hesitant to include in this permission check here, since the state users dont have access to this page anyway and we're probably going to want to push away from using the IDM user roles/persontypes for determining user capability on frontend.

client/src/pages/DemonstrationsPage.tsx
dustbuster: Actually this button is only on Demos home page!

connor-parke-icf (approved):
LGTM! love wrapping the button in its own tidy component with its own dedicated tests and whatnot. 

I would remove that test which checks for the absence of something; sure it proves that the task was done, but it doesnt tell us much about the current behavior of the app in isolation. For the permissions thing, i think we should settle on a comprehensive way to determine user permissions on the frontend, but that may be a post-mvp convo, since everywhere except the deliverables page is fair game to non-state users. but i dont think thats a blocker

PR #1471, opened 2026-06-22
DEMOS-2210: Updates to SubmitButton for 508

tresdonicf:
Per DEMOS-2210, fixes the call to `aria-label` for this button, in the capital B button I guess it's `ariaLabel`. I looked into ways to have typescript flag this but it didn't seem to be possible since tsc treats `aria` props differently.

The fix would be just moving to `aria-label` for every prop, added to the refactors list here: https://confluenceent.cms.gov/x/ApwtU

firefox_10_39_09


james-becwar-cms (approved):

client/src/components/dialog/demonstration/DemonstrationDialog.tsx
```diff
@@ -53,6 +55,28 @@ const DemonstrationDescriptionTextArea: React.FC<{
...
+      onClick={onClick}
+      ariaLabel={"Create New Demonstration"}
+      disabled={disabled || isSubmitting}
+    >
+      {isSubmitting && <Spinner />}
+      {isSubmitting ? "Loading" : "Submit"}
+    </Button>
+  );
+};
+
```
dustbuster, line 79: i like this. to my point on Thursday on my own PR. If wrap up button into a minimal wrapper than we are just writing <SpecificUseCaseButton /> which is getting overly componentized.

dustbuster (approved):

connor-parke-icf (approved):

PR #1473, opened 2026-06-22
DEMOS-2322 Download Reports

JeffreyM-GA:
This wires in the download report buttons


dustbuster (approved):
LGTM!

connor-parke-icf (approved):
LGTM! these some debate im having with myself about whether mocking the dependencies of a child of a component is better or worse than mocking the child itself, as in the ReportsPage.test.tsx.  On one hand, mocking only the dependencies lets us get a more hollistic picture of the behavior, on the other, we're mocking something that doesnt actually get pulled into the parent file (its just a passthrough component). just food for thought, no idea what style i prefer

PR #1475, opened 2026-06-22
DEMOS-2285: made deleting BN workbook also delete associated bn data

connor-parke-icf:
A good be simpler than i was worrying about. Just need to check the document type and if its a BN workbook, also delete the BN data within the same transaction. Because deleting a document is only happening via one spot thats already validated, just needed to add on the delete. Went ahead and built out with the common pattern of utilizing a query, then made the frontend use the same const we're using in backend where applicable. 

https://github.com/user-attachments/assets/09aae059-8316-40e9-b912-9b47aa126095

### DIFF
--- a/client/src/components/dialog/document/AddDocumentToDeliverableDialog.tsx
+++ b/client/src/components/dialog/document/AddDocumentToDeliverableDialog.tsx
@@ -16,8 +16,7 @@ import { tryUploadingFileToS3 } from "./tryUploadingFileToS3";
 import { useBNValidationStatus } from "./useBNValidationStatus";
 import { useDocumentPassedVirusScan } from "./useDocumentPassedVirusScan";
 import { useUploadDocument } from "./useUploadDocument";
-
-const BN_WORKBOOK_DOCUMENT_TYPE: DocumentType = "BN Workbook";
+import { BN_WORKBOOK_DOCUMENT_TYPE } from "demos-server-constants";
 
 export const UPLOAD_DOCUMENT_TO_DELIVERABLE_STATE_FILES_MUTATION: TypedDocumentNode<
   {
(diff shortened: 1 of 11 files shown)


client/src/components/dialog/document/AddDocumentToDeliverableDialog.tsx
```diff
@@ -16,8 +16,7 @@ import { tryUploadingFileToS3 } from "./tryUploadingFileToS3";
 import { useBNValidationStatus } from "./useBNValidationStatus";
 import { useDocumentPassedVirusScan } from "./useDocumentPassedVirusScan";
 import { useUploadDocument } from "./useUploadDocument";
-
-const BN_WORKBOOK_DOCUMENT_TYPE: DocumentType = "BN Workbook";
+import { BN_WORKBOOK_DOCUMENT_TYPE } from "demos-server-constants";
```
connor-parke-icf, line 19: because we needed a const like this in the server, went ahead and looked for the places we use "BN Workbook" and made them pull from a common const.

server/src/model/document/handleDeleteDocument.ts
```diff
@@ -2,12 +2,17 @@ import { Prisma, Document as PrismaDocument } from "@prisma/client";
...
+import { deleteBudgetNeutralityWorkbook } from "../budgetNeutralityWorkbook/queries";
 
 export async function handleDeleteDocument(
   where: Prisma.DocumentWhereUniqueInput,
   tx: PrismaTransactionClient
 ): Promise<PrismaDocument> {
   const document = await deleteDocument(where, tx);
+  if (document.documentTypeId === BN_WORKBOOK_DOCUMENT_TYPE) {
+    await deleteBudgetNeutralityWorkbook({ id: document.id }, tx);
+  }
```
connor-parke-icf, line 15: this is the main change. Its already always in a transaction cause of s3, so we just need to delete the BN data while we're at it if we need to

dustbuster (approved):
Looks fine to me.

PR #1476, opened 2026-06-22
(1/2) DEMOS-2218: Select Cleanup

tresdonicf:
Looking at the dialog dropdown story for DEMOS-2218 and noticed our `components/input/select` folder could use a bit of love so went ahead and did some cleanup work. The dropdown stuff is also a bit more involved then it seemed so didn't want to muddy the signal there

- Better semantics, combobox for the autocomplete selects is more accurate
- Update tests to use `getByTestId` - much more reliable
- Inline props if possible, avoid drift / separation by other code
- Remove an unused select


client/src/components/input/select/SelectDemoStatuses.tsx
tresdonicf, line 1: unused

client/src/components/input/select/AutoCompleteMultiselect.tsx
```diff
@@ -120,6 +120,11 @@ export const AutoCompleteMultiselect: React.FC<AutoCompleteMultiselectProps> = (
           data-testid={id ?? "input-autocomplete-multiselect"}
           id={id}
           type="text"
+          role="combobox"
+          aria-expanded={isOpen}
+          aria-haspopup="listbox"
+          aria-autocomplete="list"
+          aria-controls={id ? `${id}-listbox` : "multiselect-listbox"}
```
tresdonicf, line 127: better semantics for this

client/src/components/input/select/AutoCompleteMultiselect.test.tsx
```diff
@@ -49,30 +49,30 @@ describe("AutoCompleteMultiselect", () => {
 
   it("opens dropdown on input focus", async () => {
     render(<AutoCompleteMultiselect options={options} onSelect={onSelect} />);
-    const input = screen.getByRole("textbox");
+    const input = screen.getByTestId(AUTOCOMPLETE_MULTISELECT_TEST_ID);
```
tresdonicf, line 52: A bunch of moves to using the named exported variables from the tests, less brittle

client/src/components/dialog/BaseDialog.tsx
```diff
@@ -6,25 +6,13 @@ import { ConfirmCancellationDialog } from "./ConfirmCancellationDialog";
...
-  name?: string;
-  title: string;
-  onClose: () => void;
-  children: React.ReactNode;
-  actionButton?: React.ReactNode;
-  dialogHasChanges?: boolean;
-  maxWidthClass?: string;
-  hideHeader?: boolean;
-  cancelButtonIsDisabled?: boolean;
-}
```
tresdonicf, line 19: If the interface isn't exported (or exported but not used) then it's simpler to keep it inline with the props that way they're directly next to each other, can't get separated by any other code

client/src/components/input/select/SelectDemoStatuses.tsx
connor-parke-icf: ah this mustve been from back when we could change the status in the dialog. Nice!

client/src/components/input/select/AutoCompleteSelect.tsx
```diff
@@ -121,7 +123,7 @@ export const AutoCompleteSelect: React.FC<AutoCompleteSelectProps> = ({
       return filteredOptions.map((option, i) => {
         const isActive = i === activeIndex;
         return (
-          <li key={option.value}>
+          <li key={option.value} role="option" aria-selected={option.value === value}>
```
connor-parke-icf, line 126: nice! i guess with this component there wasnt a good way to tell from a screenreader perspective what options were selected in the dropdown other than examining a nearby checkbox or something

connor-parke-icf (approved):
This all looks great to me. Love the changes to inline the props with the function def; certainly much easier to read instead of having to scroll halfway up a page to edit, especially when most of the time we need to edit both in tandem. Also like the usage of testId; certainly a good habit to use those whenever possible.  

Looks like both of those changes, some slight styling changes, insertion of accessibility features and props are the main things here. All good by me!

client/src/components/input/select/AutoCompleteSelect.tsx
tresdonicf: Yeah pretty much, interesting that in this case it might be harder for sighted users to tell what's selected at a glance!

PR #1477, opened 2026-06-22
DEMOS-NO TICKET YET; stopped ext and amend w/o approval

dustbuster:
So if there's no approval you cannot make extension/amend for demo
this is a quick draft.


dustbuster:
Moving to real branch

PR #1479, opened 2026-06-23
Remove  `addTypename` on MockedProvider

tresdonicf:
Apollo has deprecated `addTypename` on MockedProvider, this defaults to `true` if it's not included, but it never needs to be `false` since apollo will always return the `_typename` field


dustbuster (approved):
looks straightforward!

connor-parke-icf (approved):
nice! this is something that had been on my radar for a bit but never got around to it. lgtm

PR #1481, opened 2026-06-23
DEMOS-2219 docTitle for pdf

nunesmlucas:
### DEMOS-2219: Show document title in the file preview (not a UUID)

On the State/CMS Files preview, the browser's native PDF viewer toolbar showed a
random UUID instead of the document title. We embedded a `blob:` URL, the PDF had
no embedded title, so the viewer fell back to the blob URL's generated UUID.

Give the viewer a real title via the file's `Content-Disposition`, and embed the
URL that carries it.

downloads now save with the document title instead of the UUID.

https://github.com/user-attachments/assets/88ed64ca-d7a2-455a-ad41-16d2b5d2926f


client/src/pages/DocumentDetails/DocumentPreview.tsx
connor-parke-icf, line 29: does the issue occur still with non-pdf files? when there are non-pdfs, the direct download link may still call it the documentId; i remember reading something about how 3rd party sources may not respect the download={filename} but i hadnt looked into it myself.

server/src/adapters/s3/AwsS3Adapter.ts
```diff
@@ -51,10 +51,13 @@ export function createAWSS3Adapter(): S3Adapter {
       });
     },
 
-    async getPresignedDownloadUrl(key: string): Promise<string> {
+    async getPresignedDownloadUrl(key: string, fileName?: string): Promise<string> {
       const getObjectCommand = new GetObjectCommand({
         Bucket: cleanBucket,
         Key: key,
+        ResponseContentDisposition: fileName
+          ? `inline; filename="${fileName.replace(/"/g, "")}"; filename*=UTF-8''${encodeURIComponent(fileName)}`
```
connor-parke-icf: we may want a more rigorous or library-based sanitization for this; we dont have any restrictions i dont think on what a document may be called, so i wonder if it may try to download as a name of wacky characters.

connor-parke-icf (approved):
Nice, i had ran into this behavior a couple times and from what i was reading handling this as a param within the url's query string was the way to go. Curious, does this gracefully handle file extensions?

I would verify the case of non-embedded files whether or not those download with the same expected names. I dont _think_ the timing out of the s3 link should be a problem since the embed runs basically right after the gql fetch, and for the direct download button the blob already has all the data so it doesnt need to download anything. 

It might be good to more thoroughly sanitize the documents, or maybe just leave them alone since files can have spaces in their names. Not sure the requirements on this, but it might be the users are expecting to get back that awkward, multi-worded name

nunesmlucas:
`content-disposition` safely turns a doc's title into the "save-as" filename header — titles are free text and hand-rolling that header is error-prone/injection-prone. Tiny, no sub-deps, the same lib Express uses.

nunesmlucas:
Pinned `fast-uri` to `4.0.1` via `overrides` to clear a Snyk high (prisma dev dep). Verified prisma still works — no ignore needed.

If you guys have any issues with this please let me know...

PR #1483, opened 2026-06-23
DEMOS-2319-clean-up-unused-auth-functions

TARehman:
This is the initial PR relating to changes to the user / person system that are necessary to support the data migration. In this case, however, I am just trying to clean up and simplify some aspects of the auth module in preparation for needing to expand its functionality to handle associating migrated users to new Cognito subjects. In this PR, I do the following.

- Remove the `auth.plugin.ts` and `cookie.plugin.ts` files. `cookie.plugin.ts` was entirely unused and not wired to anything. `auth.plugin.ts` was wired up but because of how the code in `server.ts` and `local-server.ts` was written, it was never being exercised.
- Remove `claim-utils.ts` as this was not in use at all.
- Extract `getPersonTypeFromClaims` into its own testable unit. Then, place this in both `local-server.ts` and `server.ts`. Doing this validates that the claims have one of the three valid person types that are allowed to access DEMOS. Previously, this was buried fairly deeply into the code that was also used to provision new users, etc, which made it hard to trace. Now it is checked at the server level in a transparent fashion.


james-becwar-cms (approved):

server/src/local-server.ts
```diff
@@ -69,6 +63,9 @@ const { url } = await startStandaloneServer<GraphQLContext>(server, {
 
       const decodedToken = await decodeToken(token);
       const claims = extractClaimsFromDecodedToken(decodedToken);
+      // Validate that the claims map to a valid person type (role gate);
+      // throws and rejects the request if the role is missing or ambiguous.
+      getPersonTypeFromClaims(claims);
```
connor-parke-icf, line 68: hmm, not huge into running this function both here in local-server and again in getPersonTypeFromClaims for the same purpose. Might be we should take the return from this call and save it to the claims object so we dont need to re search for the applicable role. not a bit deal either way, just seems redundant as it is now.

connor-parke-icf (approved):
lgtm. I like moving the getPersonTypeFromClaims out to its own function; however, i think calling it from two places for different purposes might be a bit redundant. Itd be nice if we could call it once, and it returns the valid role, and we pass that through with the claims and not have to search for it again later. 

That being said, this all looks fine to me and the early validation is probably good

server/src/local-server.ts
TARehman: Discussed with Connor a bit; this may be changed somewhat as I work through the rest of 2319.

PR #1485, opened 2026-06-23
DEMOS-2321: added a refetch queries to EditDelivble

dustbuster:
The history was not being reactive. So after a deliverable update is done we just fetch deliverable actions to get the items we need.  

```JavaScript
+ deliverableActions {
+         id
+         actionType
+         actionTimestamp
+         userFullName
+         details
+ }
```
Tho i changed pattern the screen recording does not change! 

https://github.com/user-attachments/assets/35230d08-716f-4aa4-89a3-73ec6aad028f

### DIFF
--- a/client/src/pages/deliverables/DeliverableDetailsManagementPage.test.tsx
+++ b/client/src/pages/deliverables/DeliverableDetailsManagementPage.test.tsx
@@ -1,9 +1,8 @@
 import React from "react";
 import { Route, Routes } from "react-router-dom";
-import { fireEvent, render, screen, waitFor } from "@testing-library/react";
+import { render, screen, waitFor } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import { describe, expect, it, vi } from "vitest";
-import { InMemoryCache } from "@apollo/client";
 import {
   DeliverableDetailsManagementPage,
   DELIVERABLE_DETAILS_QUERY,
@@ -20,12 +19,7 @@ import {
 import { FILE_AND_HISTORY_TABS_NAME } from "./sections/FileAndHistoryTabs";
 import { REQUEST_EXTENSION_BUTTON_NAME } from "./sections/DeliverableButtons";
 import { DialogProvider } from "components/dialog/DialogContext";
-import {
-  EDIT_DELIVERABLE_DIALOG_TITLE,
-  EDIT_DELIVERABLE_REASON_FIELD_NAME,
-  EDIT_DELIVERABLE_SAVE_BUTTON_NAME,
-  UPDATE_DELIVERABLE_MUTATION,
-} from "components/dialog/deliverable/EditDeliverableDialog";
+import { EDIT_DELIVERABLE_DIALOG_TITLE } from "components/dialog/deliverable/EditDeliverableDialog";
 import {
   DELIVERABLE_REVIEW_NOTICE_NAME,
   START_REVIEW_BUTTON_NAME,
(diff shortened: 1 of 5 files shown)


connor-parke-icf (approved):
I think this is fine, though i think routing the refetchQueries prop through the component stack might be over generalizing since this dialog is only accessible from one location anyway; Might can just add the refetch queries in the dialog itself and not make it a prop. 

Alternatively, we could just add the deliverableActions to the return of the UpdateDeliverable mutation
```
export const UPDATE_DELIVERABLE_MUTATION = gql`
  mutation UpdateDeliverable($id: ID!, $input: UpdateDeliverableInput!) {
    updateDeliverable(id: $id, input: $input) {
      id
      name
      dueDate
      cmsOwner {
        id
        person {
          id
          fullName
        }
      }
      demonstrationTypes {
        tagName
        approvalStatus
      }
      deliverableActions {
        id
        actionType
        actionTimestamp
        userFullName
        details
      }
    }
  }
`;
```

client/src/pages/deliverables/DeliverableDetailsManagementPage.tsx
```diff
@@ -219,7 +215,9 @@ export const DeliverableDetailsManagementPage: React.FC<{
   const handleDeleteDeliverable = useCallback(() => {}, []);
   const handleEditDeliverable = useCallback(() => {
     if (data?.deliverable) {
-      showEditDeliverableDialog(data.deliverable);
+      showEditDeliverableDialog(data.deliverable, {
+        refetchQueries: [GET_DELIVERABLE_DETAILS_QUERY_NAME],
+      });
```
tresdonicf: In this case we don't really gain by passing through a parameterized version of `refetchQueries` - it would be a bit better to add this directly to the `useMutation` call and cut down on the lines of code changed here to `1`, in a [YAGNI ](https://lawsofsoftwareengineering.com/laws/yagni/) sort of consideration.

client/src/pages/deliverables/DeliverableDetailsManagementPage.test.tsx
```diff
@@ -283,6 +289,79 @@ describe("DeliverableDetailsManagementPage", () => {
     expect(screen.getByText(EDIT_DELIVERABLE_DIALOG_TITLE)).toBeInTheDocument();
   });
 
+  it("refreshes history after saving a due date edit from the route", async () => {
+    const user = userEvent.setup();
+    const updatedDeliverable = {
+      ...MOCK_DELIVERABLE_1,
+      dueDate: new Date("2026-07-20"),
+      deliverableActions: [
```
tresdonicf: This test mocks a bit too much to provide value for what we're asking of it. Better to add a test of the form we have to test this in other places.:
```
    expect(mockMutation).toHaveBeenCalledWith({
      variables: {
        id: "deliverable-1",
        finalStatus: "Accepted",
      },
      refetchQueries: [
        { query: DELIVERABLE_DETAILS_QUERY, variables: { id: "deliverable-1" } },
      ],
      awaitRefetchQueries: true,
    });
```

The point is that if this refetchQueries changes the test will provide a signal and catch any unexpected changes here in the minimal way to do so

tresdonicf (dismissed):
I think it would be worth simplifying this a bit, particularly the test for it, the parameterized `refetchQueries` isn't a deal breaker but not needed either

client/src/pages/deliverables/DeliverableDetailsManagementPage.tsx
tresdonicf: Oh I see this is also pretty much exactly what Connor was saying, but yeah totally agree there

dustbuster:
> export const UPDATE_DELIVERABLE_MUTATION = gql`
>   mutation UpdateDeliverable($id: ID!, $input: UpdateDeliverableInput!) {
>     updateDeliverable(id: $id, input: $input) {
>       id
>       name
>       dueDate
>       cmsOwner {
>         id
>         person {
>           id
>           fullName
>         }
>       }
>       demonstrationTypes {
>         tagName
>         approvalStatus
>       }
>       deliverableActions {
>         id
>         actionType
>         actionTimestamp
>         userFullName
>         details
>       }
>     }
>   }
> `;

This is a great suggestion. let me just try this out.

client/src/pages/deliverables/DeliverableDetailsManagementPage.test.tsx
dustbuster: yeah. I agree. Resetting and re-applying.

PR #1486, opened 2026-06-24
Demos 2336: Prevent frontend for creating extension and amendments on unapproved demos

dustbuster:
We'll take a couple approches. 
Prevent the UI from showing unapproved demos in the select when you "create new"
And finding places where ext/amd are creating and limiting functionality 

i test this all and it works great. 
Screenshot 2026-06-24 at 15 37 05

Screenshot 2026-06-22 at 16 32 12


client/src/components/header/CreateNewButton.tsx
```diff
@@ -53,21 +55,23 @@ export const CreateNewButton: React.FC = () => {
...
               showCreateAmendmentDialog();
             }}
-            className="w-full text-left px-1 py-[10px] hover:bg-gray-100"
+            className="w-full text-left px-1 py-[10px] hover:bg-gray-100 disabled:text-text-placeholder disabled:cursor-not-allowed disabled:hover:bg-white"
           >
             Amendment
           </button>
           <button
             data-testid="button-create-new-extension"
+            disabled={!hasApprovedDemonstrations}
```
connor-parke-icf: might can add a tooltip to explain why: "No Approved Demonstrations Exist" maybe?

client/src/components/input/select/SelectDemonstration.tsx
```diff
@@ -32,13 +33,17 @@ export const SelectDemonstration: React.FC<{
   const demonstrations = demonstrationsData?.demonstrations;
   if (demonstrationsError) return <p>Error retrieving list of demonstrations.</p>;
   if (!demonstrations || demonstrations.length < 1) return <p>No demonstrations found.</p>;
+  const approvedDemonstrations = demonstrations.filter(
+    (demonstration) => demonstration.status === "Approved"
+  );
+  if (approvedDemonstrations.length < 1) return <p>No demonstrations found.</p>;
```
connor-parke-icf: we can probably shift some of this around so we only check for the list length once. Maybe:
```
if (demonstrationsError || !demonstrations) return <p>Error retrieving list of demonstrations.</p>;
const approvedDemonstrations = demonstrations.filter(
    (demonstration) => demonstration.status === "Approved"
  );
if (approvedDemonstrations.length < 1) return <p>No approved demonstrations found.</p>;
```
if its not loading and `demonstrations` is still undefined, we can safely assume it broke with or without an associated error

connor-parke-icf (approved):
this all looks great to me. Couple of nitpicks, but it looks like we're 
* blocking the create modifications from the listing page if no approved demos exist 
* filtering the list of applicable demonstrations in the demo select
* blocking the create buttons for modifications on the demo details pages 
* blocking the create buttons on the modifications tabs

i think that covers everywhere? at least that i can think of.

i would validate if theres any weird behavior between "my demonstrations" and "all demonstrations", like blocking create cause there isnt a demo in my demos when there are in all demos. 

maybe add some tooltips to the disabled buttons where applicable to say something like "cant create amendment on unapproved demonstration" or simply "demonstration not approved". 

NJ!

client/src/components/header/CreateNewButton.tsx
```diff
@@ -53,21 +59,25 @@ export const CreateNewButton: React.FC = () => {
           </button>
           <button
             data-testid="button-create-new-amendment"
+            disabled={!hasApprovedDemonstrations}
```
dustbuster: Tested this and it works great!

Image

dustbuster: Running this by Michael.

PR #1487, opened 2026-06-24
DEMOS-2319-refactor-existing-user-creation-code

TARehman:
This is a fairly large refactor of how the auth module currently works, which will support adding in the ability to link PMDA users (coming in a future PR). I did a number of things.

- Added two new fields and some check constraints to the `users` table in the database. This is necessary to support future functionality with PMDA users.
- Cleaned up imports from the `auth` module across the codebase; we were all over the place on these before.
- Removed ghost mocks and code that was entirely useless in some tests.
- Moved a bunch of code into `auth/user` to try and isolate concerns. It's still not perfect; the way we are handling the claims in particular is misleading and we do a lot of processing at the lowest level which is odd to me. However, I want to try and keep the scope of this manageable given the timeframe, so I'm limiting the amount of stuff I refactor.
  - The code breaks a lot of the stuff that was all in one place in `findOrCreateContextUserFromClaims` into separate testable units. We now have `createNewUserFromClaims` and `findUserByCognitoSubject`, which in turn use a number of new query functions that each were added with proper tests.
  - The previous `getPersonTypeFromClaims` function was moved to `auth/user` and then a semantic `validatePersonTypeInClaim` wrapper was added at the `auth` level. This helps to avoid some circular import stuff. I'm still not 100% in love with this, but its currently the best way I can see to structure things given how the code was originally generated.
  - `findOrCreateContextUserFromClaims` now orchestrates the entire process within a transaction. Using a null coalesce, we try to find the user by its Cognito subject, and if that fails, we then create the new user. In the future, this is where we'll add the third path of trying to link together with migrated users, but I wanted to get this reviewed first.


server/src/auth/user/createNewUserFromClaims.ts
```diff
@@ -0,0 +1,61 @@
...
+  const newPersonTypeId = getPersonTypeFromClaims(claims);
+  const person = await insertPerson(
+    {
+      personTypeId: newPersonTypeId,
+      email: claims.email,
+      firstName: claims.givenName,
+      lastName: claims.familyName,
+    },
+    tx
+  );
```
james-becwar-cms, line 31: Is there a chance that we end up with duplicate Persons here?

server/src/auth/user/createNewUserFromClaims.ts
```diff
@@ -0,0 +1,61 @@
...
+    {
+      id: person.id,
+      personTypeId: newPersonTypeId,
+      cognitoSubject: claims.sub,
+      username: claims.externalUserId,
+      isMigratedFromPmda: false,
+      hasLoggedIn: true,
+    },
+    tx
+  );
```
james-becwar-cms, line 42: Could we get a record for the migrated user and a non migrated user from the same user?

server/src/auth/user/userContext.ts
```diff
@@ -0,0 +1,17 @@
+import type { AuthorizationClaims } from "..";
+import type { ContextUser } from ".";
+import { prisma } from "../../prismaClient";
+import { upsertUserSession } from "../../model/userSession/queries";
+import { createNewUserFromClaims, findUserByCognitoSubject } from ".";
+
+export async function findOrCreateContextUserFromClaims(
+  claims: AuthorizationClaims
+): Promise<ContextUser> {
+  return await prisma().$transaction(async (tx) => {
```
james-becwar-cms, line 10: Since you are returning a promise you don't need to await  the Prisma transaction.  As written we are returning a promise that is always resolved.

server/src/auth/user/userContext.ts
```diff
@@ -0,0 +1,17 @@
...
+    const returnedUser =
+      (await findUserByCognitoSubject(claims.sub, tx)) ??
+      (await createNewUserFromClaims(claims, tx));
```
james-becwar-cms, line 13: This code is syntactically correct, but I would rewrite it to be more explicit and not rely on the short circuit of the nullish coalescing operator.

server/src/auth/validatePersonTypeInClaim.ts
```diff
@@ -0,0 +1,8 @@
+import type { AuthorizationClaims } from ".";
+import { getPersonTypeFromClaims } from "./user";
+
+// Note: this is an extremely lightweight semantic wrapper
+// It exists to clarify what is happening in the server when this is called
+export function validatePersonTypeInClaim(claims: AuthorizationClaims): void {
+  getPersonTypeFromClaims(claims);
+}
```
james-becwar-cms, line 8: Personal opinion is that this adds cognitive load and couples this function the getPersonTypeFromClaims().  I also don't like how getPersonTypeFromClaims() isn't idempotent, but thats a bigger issue then this PR.

I would consider removing this file.

server/src/seeder.ts
```diff
@@ -874,6 +876,8 @@ async function seedDatabase() {
         personTypeId: person.personTypeId,
         cognitoSubject: faker.string.uuid(),
         username: faker.internet.username(),
+        isMigratedFromPmda: false,
+        hasLoggedIn: true,
```
james-becwar-cms, line 880: Should add a few fake migrated users?

james-becwar-cms (approved):
Left some style comments/questions about defensive code.  Nothing that would stop a merge.

server/src/auth/user/createNewUserFromClaims.ts
```diff
@@ -0,0 +1,61 @@
...
+  "demos-cms-user": "CMS User",
+  "demos-state-user": "State User",
+};
+
+export async function createNewUserFromClaims(
+  claims: AuthorizationClaims,
+  tx: PrismaTransactionClient
+): Promise<ContextUser> {
+  const newPersonTypeId = getPersonTypeFromClaims(claims);
+  const person = await insertPerson(
```
connor-parke-icf, line 23: oh nice, i see; this uses the same structure of query functions we have for the resolver flows. nice!

server/src/auth/user/createNewUserFromClaims.ts
```diff
@@ -0,0 +1,61 @@
...
+      hasLoggedIn: true,
+    },
+    tx
+  );
+  const systemRoleAssignment = await insertSystemRoleAssignment(
+    {
+      personId: person.id,
+      grantLevelId: "System",
+      personTypeId: newPersonTypeId,
+      roleId: initialUserTypeRoles[newPersonTypeId],
```
connor-parke-icf, line 48: one part of the migration thats concerning me is the influx of new roles; no idea how we are going to handle that, whether or not we're going to create a bunch of system-level roles for the new ones. If so, this may be a little complicated to default

server/src/auth/user/findUserByCognitoSubject.ts
```diff
@@ -0,0 +1,31 @@
...
+  if (!user) {
+    return null;
+  }
+  const systemRoles = await selectManySystemRoleAssignments({ personId: user.id }, tx);
+  const rolePermissions = await selectManyRolePermissions(
+    {
+      roleId: { in: systemRoles.map((sr) => sr.roleId) },
+    },
+    tx
+  );
```
connor-parke-icf, line 22: aside: one of the limiting things about our current query function structure is it makes joins fairly rigid, so getting the user and their permissions in one query is more difficulty. might be worth thinking about passing in optional includes params. But that may break the intended permissions flow.

server/src/model/deliverableTypeDocumentType/selectDocumentTypesForDeliverableType.test.ts
connor-parke-icf, line 1: assuming we just didnt have a test for this before?

server/src/model/user/user.prisma
```diff
@@ -24,5 +26,6 @@ model User {
   @@id([id])
   @@unique([id, personTypeId]) // Not required by Postgres, but Prisma requires this notation
   @@unique([cognitoSubject])
+  @@unique([username])
```
connor-parke-icf, line 29: i imagine this is how we'll be matching up the pmda users before they have a cognito subject?

connor-parke-icf (approved):
This all looks fine to me. a lot of different moving parts, but i dont see anything that really gives me pause that hasnt already been commented on. I like the instantiation of query methods for the different user/person operations and the barrelling up of the different auth files where applicable. The user db changes make sense to me as well.

server/src/auth/user/createNewUserFromClaims.ts
TARehman: In what way do you mean? The DB has a unique identifier so in that sense, no, its impossible (Prisma generates it). In the sense of a concurrency issue, its not _entirely_ impossible but I'm hoping the DB transaction will block most cases.

server/src/auth/user/createNewUserFromClaims.ts
TARehman: How exactly would that happen? These users aren't migrated ones, and to be clear, we don't actually have that code in place yet. The point is that a user coming in who falls through the "already has a user" check goes right now to the "make a new user" one, and any user we are making in that fashion aren't migrated from PMDA by definition because we're making them.

server/src/auth/user/userContext.ts
TARehman: I thought about that and went back and forth about it, to be honest. The issue is that if you do it that way it looks really odd. You have to create something called `existingUser` and then check if it exists or not and so on. This was one rare case where I thought the nullish coalesce was actually fairly readable. When I do the third option for the migrated users I can write it without the coalesce instead, if you prefer.

server/src/auth/validatePersonTypeInClaim.ts
TARehman: I don't understand how `getPersonTypeFromClaims` isn't idempotent, but I share your skepticism of this. I don't love that it's basically a wrapper around the function and serves only as a semantic guidepost. The _underlying_ smell is not actually this code, however - it's the general way we've set up the `auth` module. It really needs a full refactor from the `AuthorizationClaims` object all the way down, so that things like `validateClaims` actually validate the claims, etc. But given times and constraints I was trying to draw a decent boundary around things.

For this one, if we get rid of this, we'll just have a bare `getPersonTypeFromClaims` call in both of the servers. We can do it, but it will just move this call to two places where we put an identical comment and say "Hey this is why we're doing this random thing." At least in this case the function name tells you what it's being used for in that context.

server/src/seeder.ts
TARehman: I didn't even seed, so I don't think so for now. I'll maybe add them once I do the next PR.

server/src/auth/user/userContext.ts
TARehman: Right, will fix.

TARehman: Discussed with James; going to merge as-is as future changes are coming from the next PR.

server/src/auth/user/userContext.ts
TARehman: Discussed with James; going to merge as-is as future changes are coming from the next PR.

server/src/auth/validatePersonTypeInClaim.ts
TARehman: Discussed with James; going to merge as-is as future changes are coming from the next PR.

server/src/auth/user/createNewUserFromClaims.ts
TARehman: Discussed with @james-becwar-cms, no issue.

server/src/auth/user/createNewUserFromClaims.ts
TARehman: Discussed with @james-becwar-cms, no issue.

server/src/auth/user/createNewUserFromClaims.ts
TARehman: All the users from PMDA will be mapped into our roles, so hopefully this isn't super painful.

server/src/auth/user/findUserByCognitoSubject.ts
TARehman: It's also a pain to do in TS because it mutates the return signature. I debated whether to make the queries more specific or just to run a few extra ones and settled on it being fine to run extra ones.

server/src/model/deliverableTypeDocumentType/selectDocumentTypesForDeliverableType.test.ts
TARehman: Yeah I just stuck it in while I was at it.

server/src/model/user/user.prisma
TARehman: We'll actually be using email most likely, but more to come in the next PR.

PR #1488, opened 2026-06-24
Demos 2330 backend no submitted state files for finalization

connor-parke-icf:
to mirror the frontend, the api also now blocks deliverables from being finalized when it has unsubmitted documents. This change was done in both the api (hooking into the validator struct already there) and the database (with a trigger, similar to how we check against active extension requests). To demonstrate, i could use the frontend without the frontend validation. here i show just the db validation (by commenting out the api validation), then the api validation itself.

Note: merge this only after the frontend validation has been merged to prevent errors from propagating to the frontend.  

https://github.com/user-attachments/assets/1e8b773d-c87e-4249-8a1d-66425507ae37


server/src/model/deliverable/checkDeliverableInputFunctions.ts
```diff
@@ -43,6 +43,19 @@ export function checkDeliverableHasStatus(
   }
 }
 
+export async function checkDeliverableHasNoUnsubmittedStateDocuments(
```
TARehman, line 46: Should this function be wrapped / part of a transaction?

server/src/model/deliverable/completeDeliverable.ts
```diff
@@ -18,7 +18,7 @@ export async function completeDeliverable(
   validateUserPersonTypeAllowed(context, "completeDeliverable", ["demos-admin", "demos-cms-user"]);
   return await prisma().$transaction(async (tx) => {
     const incompleteDeliverable = await selectDeliverableOrThrow({ id: deliverableId }, tx);
-    validateCompleteDeliverableInput(incompleteDeliverable);
+    await validateCompleteDeliverableInput(incompleteDeliverable);
```
TARehman: Same transaction question. I think its actually probably really important to do it in the same TX that you're inside.

server/src/model/deliverable/validateDeliverableInputs.ts
```diff
@@ -31,6 +31,7 @@ import {
 import { GraphQLContext } from "../../auth";
 import { PersonType } from "../../types";
 import { ACTIVE_DELIVERABLE_STATUSES } from "../../constants";
+import { checkDeliverableHasNoUnsubmittedStateDocuments } from "./checkDeliverableInputFunctions";
```
TARehman: We should import this through the barrel like the other eleventyfive million check* functions.

server/src/model/deliverable/validateDeliverableInputs.ts
```diff
@@ -161,11 +159,14 @@ export function validateStartDeliverableReviewInput(deliverable: PrismaDeliverab
...
 
-export function validateCompleteDeliverableInput(deliverable: PrismaDeliverable): void {
+export async function validateCompleteDeliverableInput(
+  deliverable: PrismaDeliverable
+): Promise<void> {
   const errors: (string | undefined)[] = [];
 
   // Deliverables may only be completed from review status
   errors.push(checkDeliverableHasStatus(deliverable, ["Under CMS Review"]));
+  errors.push(await checkDeliverableHasNoUnsubmittedStateDocuments(deliverable));
```
TARehman: Probably want a comment above this since otherwise it looks confusing now that there's two checks.

TARehman (changes requested):
I think you need to wrap some stuff in transactions that isn't right now.

server/src/model/deliverable/checkDeliverableInputFunctions.ts
connor-parke-icf: good catch; propagating the transaction param through the validation

server/src/model/deliverable/completeDeliverable.ts
connor-parke-icf: agreed; updating!

server/src/model/deliverable/validateDeliverableInputs.ts
connor-parke-icf: done!

server/src/model/deliverable/validateDeliverableInputs.ts
connor-parke-icf: done!

server/src/model/deliverable/validateDeliverableInputs.test.ts
```diff
@@ -831,21 +835,21 @@ describe("validateDeliverableInputs", () => {
         statusId: "Under CMS Review",
       };
 
-      validateCompleteDeliverableInput(testInput as PrismaDeliverable);
+      validateCompleteDeliverableInput(testInput as PrismaDeliverable, mockTransaction);
```
TARehman: I think you need to await here.

TARehman (approved):
I think you're still missing one await, but overall the issues are addressed.

server/src/model/deliverable/validateDeliverableInputs.test.ts
connor-parke-icf: fixed

PR #1489, opened 2026-06-24
DEMOS-2218: AutoCompleteSelects don't get cut off by dialog

tresdonicf:
per DEMOS-2218, this fixes the problem of dialogs cutting off the text in our custom combo box implementations. Took a while to come to this solution, a lot of editing but it's looking good now

## Problem
image

## Fixed

firefox_14_00_04
firefox_14_00_11

### DIFF
--- a/client/src/components/dialog/BaseDialog.tsx
+++ b/client/src/components/dialog/BaseDialog.tsx
@@ -6,7 +6,7 @@ import { ConfirmCancellationDialog } from "./ConfirmCancellationDialog";
 
 export const DIALOG_CANCEL_BUTTON_NAME = "button-dialog-cancel";
 
-const DIALOG = tw`bg-surface-white text-text-font w-full rounded shadow-md p-md relative max-h-[85vh] overflow-y-auto space-y-sm backdrop:bg-black/40`;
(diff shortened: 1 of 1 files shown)


dustbuster (approved):
OMG SO APPROVED!

connor-parke-icf (approved):
looks great! curious if a dialog was tall enough, whether the overflow would push beyond the bottom of the page's height since the max height is removed.  Might not ever be a real issue though so long as the combo-box is only so tall

tresdonicf:
> looks great! curious if a dialog was tall enough, whether the overflow would push beyond the bottom of the page's height since the max height is removed. Might not ever be a real issue though so long as the combo-box is only so tall

Yeah I ran through the dialogs in our dialog debug sandbox and they all looked good, pretty much all the same height as they were before(most of them weren't hitting that max-h anyhow it seemed). I did remove it because when it was there it it was actually putting a scrollbar in the dialog and just a weird interaction going on there. And yeah like you mentioned the combobox itself has a max height so it's not a problem (otherwise state / territory would always be going off the dialog)

firefox_14_46_37

PR #1490, opened 2026-06-24
Demos 2334 -  BN Workbook validation updates

james-becwar-cms:
Vivian and team created a realistic example BN workbook which is now included.  The new file exposed some issues that required some code changes.  

The changes:
- Excel doesn't always save empty rows, so if a row is empty treat the value as null (before we would throw an out of index error)
- On the C report page, The rows that say:  "Federal Share" "ADM Waivers" "Total Computable" have these values in all cells, but they look blank on the screen.  We ignore them now.
- Validation Error 7 had a typo where it was looking at k3 when it should have been f3

### DIFF
--- a/shared_library/src/BN/index.ts
+++ b/shared_library/src/BN/index.ts
@@ -64,10 +64,10 @@ export function excelColumnRow(
 
   const columnLetters = match[1]!;
   const rowNumber = parseInt(match[2]!, 10) - 1;
-  
+
   const row = sheetData.data[rowNumber];
   if (!row) {
-    throw new Error(`Row ${rowNumber + 1} not found`);
+    return null;
   }
 
   return row[excelColumnToNumber(columnLetters)] as CellValue;
--- a/shared_library/src/BN/rulesets/v1/index.ts
+++ b/shared_library/src/BN/rulesets/v1/index.ts
@@ -65,10 +65,21 @@ export const validations: ValidationFunction[] = [
         row += 1;
         continue;
       }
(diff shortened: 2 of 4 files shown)


shared_library/src/BN/rulesets/v1/index.ts
```diff
@@ -65,10 +65,21 @@ export const validations: ValidationFunction[] = [
...
+      for (
+        let colNumber = excelColumnToNumber(columnStart);
+        colNumber <= excelColumnToNumber(columnEnd);
+        colNumber++
+      ) {
         const col = numberToExcelColumn(colNumber);
         const cellValue = excelColumnRow(`${col}${row}`, "C Report", data);
-        if (cellValue !== "" && cellValue !== null) {
+        if (
+          cellValue !== "" &&
```
connor-parke-icf, line 76: wonder if theres a build-in "is falsey" or empty or whatever function that can catch this

shared_library/src/BN/index.ts
```diff
@@ -64,10 +64,10 @@ export function excelColumnRow(
 
   const columnLetters = match[1]!;
   const rowNumber = parseInt(match[2]!, 10) - 1;
-  
+
   const row = sheetData.data[rowNumber];
   if (!row) {
-    throw new Error(`Row ${rowNumber + 1} not found`);
+    return null;
```
connor-parke-icf, line 70: > Excel doesn't always save empty rows, so if a row is empty treat the value as null (before we would throw an out of index error)

will it just count like [1, 2, 3, 4, 6, 7, 8] sometimes if row 5 was blank? does the same thing happen with columns?

connor-parke-icf (approved):

PR #1493, opened 2026-06-25
Demos 2337 no modification on non approved demos

connor-parke-icf:
DO NOT MERGE without consideration/discussion around db reset

Ok this one is pretty big, but there are three main components:
* database updates - now we enforce that an extension or amendment can only be on an approved demonstration, similarly to deliverables
* api validation - now checks before creating that the associated demonstration is indeed approved 
* seeder updates - the seeder used to just create amendments and extensions randomly; now it does so only on approved demonstrations. The seeder also generates more "approved" demonstrations.  

With this, we now block invalid amendment/extension creation. Here i demonstrate in the API, once with no API validation (relying on the db constraints only), and another time with API validation (with the common error pattern).  

https://github.com/user-attachments/assets/da7fda37-33af-4b45-b6ab-5838b427aa94


data/docs/DEMOS_Data_Model.mmd
```diff
@@ -329,10 +335,6 @@ erDiagram
     text id PK, FK "FK126: id ∈ deliverable_extension_status.id"
   }
 
-  deliverable_demonstration_status_limit:::typeLimiter {
```
connor-parke-icf, line 332: because the type limiter for Deliverable and amendments/extensions was the same (demo must be approved), i decided to combine that into the same limit. I renamed the limiter to match this.

server/src/errors/cleanErrorsAndThrow.ts
```diff
@@ -0,0 +1,17 @@
+import { GraphQLError } from "graphql";
+
+export function cleanErrorsAndThrow(
```
connor-parke-icf, line 3: for validating the input of the amendments/extensions, i copied the structure from deliverables. This was a helper that was defined there, so i exported it and moved here. Theres no directly associated test file, but it is exercised in all the tests that utilize it. This could be refactored, but its used a lot and this PR is already really big

data/docs/DEMOS_Data_Model.mmd
```diff
@@ -369,7 +371,8 @@ erDiagram
   amendment:::dataTable {
     uuid id PK, FK, UK "FK054: (id, application_type_id) ∈ application.(id, application_type_id) DEFERRED<br>UK: (id, application_type_id)"
     text application_type_id FK, UK "FK054: (id, application_type_id) ∈ application.(id, application_type_id) DEFERRED<br>UK: (id, application_type_id)<br>FK057: application_type_id ∈ amendment_application_type_limit.id"
-    uuid demonstration_id FK "FK055: demonstration_id ∈ demonstration.id"
+    uuid demonstration_id FK "FK055: (demonstration_id, demonstration_status_id) ∈ demonstration.(id, status_id)"
+    text demonstration_status_id FK "FK055: (demonstration_id, demonstration_status_id) ∈ demonstration.(id, status_id)<br>FK159: demonstration_status_id ∈ approved_application_status_limit.id"
```
connor-parke-icf, line 375: same structure as what we do for deliverables. Join on both the demo id and the demo status, then join to a limit to ensure only Approved can exist for demonstration status here.

server/src/model/amendment/amendmentResolvers.ts
```diff
@@ -28,36 +27,7 @@ import { getManyDocuments } from "../document";
 import { selectManyApplicationTagAssignments } from "../applicationTagAssignment/queries";
 import { selectManyApplicationTagSuggestions } from "../applicationTagSuggestion/queries";
 import { selectManyApplicationPhases } from "../applicationPhase/queries";
-
-const amendmentApplicationType: ApplicationType = "Amendment";
-const conceptPhaseName: PhaseName = "Concept";
-const newApplicationStatusId: ApplicationStatus = "Pre-Submission";
-
-export async function __createAmendment(
```
connor-parke-icf, line 36: as part of the validation, i went ahead and popped this into its own file to mirror the structure of the createDeliverable and associated validation. Now amendments, extensions, and deliverables have similar structure in their create methods. At the moment, we can completely block off the permission for creating amendments, so we dont need to route this through a dedicated data function.

server/src/model/amendment/createAmendment.test.ts
```diff
@@ -0,0 +1,113 @@
...
+  prisma: vi.fn(),
+}));
+
+vi.mock(".", () => ({
+  validateCreateAmendmentInput: vi.fn(),
+}));
+
+import { prisma } from "../../prismaClient";
+
+describe("createAmendment", () => {
```
connor-parke-icf, line 17: this is close to a perfect copy/paste from what used to be in amendmentResolvers.test.ts

server/src/model/amendment/validateAmendmentInputs.test.ts
```diff
@@ -0,0 +1,99 @@
...
+}));
+
+vi.mock("../demonstration", () => ({
+  checkDemonstrationStatus: vi.fn(),
+}));
+
+import { getApplication } from "../application";
+import { checkDemonstrationStatus } from "../demonstration";
+
+describe("validateAmendmentInputs", () => {
```
connor-parke-icf, line 24: this is largely a copy/paste from the validateDeliverableInputs test file, just trimmed down to only care about the demonstration status check

server/src/model/applicationStatus/applicationStatus.prisma
```diff
@@ -2,10 +2,10 @@
...
   id String
 
-  amendments                          Amendment[]
-  demonstrations                      Demonstration[]
-  extensions                          Extension[]
-  deliverableDemonstrationStatusLimit DeliverableDemonstrationStatusLimit[]
+  amendments                     Amendment[]
+  demonstrations                 Demonstration[]
+  extensions                     Extension[]
+  approvedApplicationStatusLimit ApprovedApplicationStatusLimit[]
```
connor-parke-icf, line 8: im not sure the cardinality here; feel like it should be an optional, single join instead of a one to many. But thats the way it was before.

server/src/model/deliverable/index.ts
```diff
@@ -7,7 +7,6 @@ export {
   checkDeliverableHasNoComments,
   checkDeliverableHasNoDocuments,
   checkDeliverableHasStatus,
-  checkDemonstrationStatus,
```
connor-parke-icf, line 10: moved this function to the Demonstration folder, since its not used in deliverables, amendments, and extensions (with slight edit)

server/src/model/extension/createExtension.test.ts
```diff
@@ -0,0 +1,113 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
```
connor-parke-icf, line 1: everything here about extensions should be identical to amendments, since theres no real difference between them

server/src/model/migrations/20260624205341_prisma_no_modifications_on_approved_demonstrations/migration.sql
```diff
@@ -0,0 +1,53 @@
+/*
+  Warnings:
+
+  - You are about to drop the `deliverable_demonstration_status_limit` table. If the table is not empty, all the data it contains will be lost.
+  - Added the required column `demonstration_status_id` to the `amendment` table without a default value. This is not possible if the table is not empty.
+  - Added the required column `demonstration_status_id` to the `extension` table without a default value. This is not possible if the table is not empty.
+
+*/
+-- DropForeignKey
```
connor-parke-icf, line 9: So this is kinda the tough part; im not sure how we should go about handling cases where there exists invalid data already. We could try to make every associated demonstration be approved, but i think thatd just lead to way more complicated scenarios (demo must have demo types, demo must have xyz fields, etc). 

So ive build this under the assumption that a db reset will be needed.

server/src/seeder.ts
```diff
@@ -1212,60 +1278,10 @@ async function seedDatabase() {
...
-      input: updatePayload,
-    };
-    await __updateExtension(undefined, updateInput);
-  }
-
   await seedTagsAndStatuses();
-  await seedApprovedDemonstration();
+  await seedApprovedDemonstrations();
+  await seedAmendments();
+  await seedExtensions();
```
connor-parke-icf, line 1293: to summarize these changes: 
Now seeds multiple approved demonstrations, based on which ones got demo types applied. In that subset of demonstrations, amendments and extensions can be created safely. 

Ill note that the seeded approved demonstrations are not "correct", they simply have the final phase completed. While we could go through and make a "correct" approved demonstration, i dont think its all that important at the moment and would significantly blow up the scope of the seeder. So just going with the fastest path to resolution here.

server/src/model/applicationStatus/applicationStatus.prisma
TARehman: That's a Prisma-ism; it gets all weird if you try to do a 1:1 because you have to make it optional then and it looks weird.

server/src/model/migrations/20260624205341_prisma_no_modifications_on_approved_demonstrations/migration.sql
TARehman: Well, it will certainly be easier to do with a DB reset. Depends on how much that would mess up QA however.

server/src/errors/cleanErrorsAndThrow.ts
TARehman: I think I skipped making a test for this because we just test it directly in the checking of what gets thrown (e.g. it is not mocked out in the test files). I think it's fine to leave it without an explicit test right now, but we should probably stick one in here just to be comprehensive at some point.

server/src/model/amendment/amendmentResolvers.test.ts
```diff
@@ -38,6 +32,7 @@ import { selectManyApplicationTagAssignments } from "../applicationTagAssignment
 import { ApplicationTagAssignmentQueryResult } from "../applicationTagAssignment/queries";
 import { selectManyApplicationTagSuggestions } from "../applicationTagSuggestion/queries";
 import { selectManyApplicationPhases } from "../applicationPhase/queries";
+import { createAmendment } from "./createAmendment";
```
TARehman: Should we barrel this?

server/src/model/amendment/createAmendment.test.ts
```diff
@@ -0,0 +1,113 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { createAmendment } from "./createAmendment";
+import { ApplicationStatus, ApplicationType, PhaseName } from "../../types";
+import { Prisma } from "@prisma/client";
+
+const testDemonstrationId = "518aa497-d547-422e-95a0-02076c7f7698";
```
TARehman: Nit: This doesn't really matter but we usually put these inside the describe() block.

server/src/model/amendment/amendmentResolvers.test.ts
```diff
@@ -279,63 +259,29 @@ describe("amendmentResolvers", () => {
...
+    const testSignatureLevel = "OA" satisfies SignatureLevel;
+    it("should call the createAmendment function with the right arguments", async () => {
       const testInput: { input: CreateAmendmentInput } = {
         input: {
           demonstrationId: testDemonstrationId,
-          name: testAmendmentName,
-          description: testAmendmentDescription,
+          name: testName,
+          description: testDescription,
+          signatureLevel: testSignatureLevel,
```
TARehman, line 273: Didn't we remove the signature level once upon a time here because the F/E isn't usually sending one? This is fine, its just an optional argument, I am trying to remember why we didn't put it in there before? Might have been just drift.

server/src/model/amendment/amendmentResolvers.ts
```diff
@@ -28,36 +27,7 @@ import { getManyDocuments } from "../document";
...
-        name: input.name,
-        description: input.description,
-        statusId: newApplicationStatusId,
-        currentPhaseId: conceptPhaseName,
-        signatureLevelId: input.signatureLevel,
-      },
-    });
-  });
-}
+import { createAmendment } from "./createAmendment";
```
TARehman: See other comment re: barreling.

server/src/model/amendment/createAmendment.test.ts
```diff
@@ -0,0 +1,113 @@
...
+const testAmendmentTypeId: ApplicationType = "Amendment";
+const testAmendmentStatusId: ApplicationStatus = "Pre-Submission";
+const testAmendmentPhaseId: PhaseName = "Concept";
+
+vi.mock("../../prismaClient", () => ({
+  prisma: vi.fn(),
+}));
+
+vi.mock(".", () => ({
+  validateCreateAmendmentInput: vi.fn(),
```
TARehman, line 11: Don't you need to actually exercise this mock by asserting that it was called with the expected data?

server/src/model/amendment/createAmendment.test.ts
```diff
@@ -0,0 +1,113 @@
...
+          signatureLevelId: "OA",
+        },
+      },
+    ];
+    await createAmendment(testInput);
+    expect(transactionMocks.application.create).toHaveBeenCalledExactlyOnceWith(expectedCalls[0]);
+    expect(transactionMocks.amendment.create).toHaveBeenCalledExactlyOnceWith(expectedCalls[1]);
+  });
+
+  it.each(["OA", "OCD"] as const)(
```
TARehman: I'm not sure what we're testing at this point with this.

server/src/model/deliverable/validateDeliverableInputs.test.ts
```diff
@@ -88,6 +90,7 @@ import {
   selectDeliverableOrThrow,
 } from ".";
 import { ACTIVE_DELIVERABLE_STATUSES } from "../../constants";
```
TARehman: Nit: We should maybe move this up if its not part of the mocks but not worth changing on its own.

server/src/model/extension/createExtension.test.ts
TARehman: I remember literally copying and pasting between the two resolver files and then doing string substitutions. 🤣

server/src/seeder.ts
TARehman, line 1: Just to confirm, did you try running the seeder without a reset to make sure that it still works properly? It should, just confirming.

server/src/model/extension/createExtension.test.ts
connor-parke-icf: this is the way

server/src/model/amendment/amendmentResolvers.test.ts
connor-parke-icf: Im not entirely sure which way is better, since i think of the barrel files as an interface for importing when outside of the module. So, following that, this import should remain as ./createAmendment, but uses of createAmendment outside this module should go through the barrel. But i dont think we're consistently applying that rule enough to warrant pushing for it, so ill go ahead and import from the barrel

server/src/model/amendment/createAmendment.test.ts
connor-parke-icf: makes sense, thats where to put the transactionMocks and whatnot. Fixed

server/src/model/amendment/amendmentResolvers.test.ts
connor-parke-icf: there was definitly drift in a couple places. I figured that while we're testing some of the args, might as well test them all in case something does something wacky. Besides, this tests the mapping from "signatureLevel" to "signatureLevelId"

server/src/model/amendment/amendmentResolvers.ts
connor-parke-icf: barrelled

server/src/model/amendment/createAmendment.test.ts
connor-parke-icf: ahh, yes youre right. I spent so much time refactoring this that i forgot to actually test main thing in this ticket was getting called 😅.
Fixed here and in extensions as well.

server/src/model/amendment/createAmendment.test.ts
connor-parke-icf: youre right. I left this in cause it was on the resolver tests before, but its outdated. It used to accompany an explicit validation that has since been removed. I will remove this test and its counterpart on extensions.

server/src/model/deliverable/validateDeliverableInputs.test.ts
connor-parke-icf: good call, moved!

server/src/seeder.ts
connor-parke-icf: i did not, actually.  standby...
actually found i had completely missed to update after a couple of signature/location changes. With those changes, validated that seed:reset and just seed works!

server/src/model/amendment/amendmentResolvers.test.ts
connor-parke-icf: did the same for extensions

TARehman (approved):
Addressed concerns and questions, thanks @connor-parke-icf 👍🏻

PR #1494, opened 2026-06-25
DEMOS-2270: format date before sending to date picker

connor-parke-icf:
issue here was that we were passing the full ISO string into the datepicker, rather than the YYYY-mm-dd date, so nothing would display. Made it format when it generates the form data, and made the hasChanges pass the original modification through that same formatter.

### DIFF
--- a/client/src/components/dialog/modification/ModificationForm.test.tsx
+++ b/client/src/components/dialog/modification/ModificationForm.test.tsx
@@ -323,16 +323,16 @@ describe("ModificationForm", () => {
     describe("hasChanges", () => {
       it("returns false when no fields have changed", () => {
         const formData: ModificationFormData = {
-          demonstrationId: "demo-1",
           name: "Test",
           description: "Description",
           signatureLevel: "OA",
+          effectiveDate: "2024-02-01" as LocalDate,
         };
         const original: Partial<Modification> = {
-          demonstration: { id: "demo-1" },
           name: "Test",
           description: "Description",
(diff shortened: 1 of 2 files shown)


client/src/components/dialog/modification/ModificationForm.test.tsx
```diff
@@ -398,7 +398,7 @@ describe("ModificationForm", () => {
           name: "Test",
           description: "Description",
           signatureLevel: "OA",
-          effectiveDate: "2024-01-01" as LocalDate,
+          effectiveDate: "2024-06-01T04:00:00.000Z",
```
connor-parke-icf, line 401: when the original modification is passed into the hasChanges method, its date has yet to be formatted. So tried to make that explicit here

client/src/components/dialog/modification/ModificationForm.tsx
```diff
@@ -49,13 +50,13 @@ export const hasChanges = (
   createModificationFormData: ModificationFormData,
   initialModification: Partial<Modification>
 ): boolean => {
-  const initialEffectiveDate = initialModification.effectiveDate ?? undefined;
+  const initialModifiactionFormData = getFormDataFromModification(initialModification);
```
connor-parke-icf, line 53: passing the initialModification to getFormDataFromModification lets us compare apples to apples, since the createModificationFormData has itself already been passed through this method.

tresdonicf (approved):

PR #1495, opened 2026-06-25
DEMOS-2343: validated that submission is more meaningful

connor-parke-icf:
DO NOT MERGE until the associated [frontend component](https://jiraent.cms.gov/browse/DEMOS-2342) is completed, else error messages will appear seemingly for no reason. 

This PR is the backend component required to validate that a submission can only be made if it is meaningful. This comes in the form of two scenarios:
* there exists at least one unsubmitted document on the deliverable
* the deliverable has a state of "Upcoming" or "Past Due"

The existing check for at least one document on the deliverable remains. 

With this, the state user can upload a doucment, hit submit, then they wont be able to submit again until they either upload a new document or CMS takes an action to put the status back in upcoming (like a resubmission request). After that, the State User can hit submit again without uploading a document to signify "no, what i uploaded before is correct" without blocking any part of the process. 

https://github.com/user-attachments/assets/23fccc17-9bbf-468f-9a07-f7b709a3af08


server/src/constants.ts
```diff
@@ -185,7 +190,7 @@ export const DOCUMENT_TYPES = [
 ] as const;
 export type DocumentType = (typeof DOCUMENT_TYPES)[number];
 
-export const BN_WORKBOOK_DOCUMENT_TYPE = 'BN Workbook' as const satisfies DocumentType;
+export const BN_WORKBOOK_DOCUMENT_TYPE = "BN Workbook" as const satisfies DocumentType;
```
connor-parke-icf, line 193: i have since reverted my linter settings...

dustbuster (approved):
Look good!

PR #1496, opened 2026-06-25
QOL: More concise messages for `getBy` and `findBy` errors

tresdonicf:
A small change I made for another piece of work that I figured I'd break out into its own piece, the error messages before would dump a pretty large DOM tree so I this just helps scroll through the errors a bit easier

https://github.com/user-attachments/assets/e46e6bd8-5b13-4b73-afdf-2cd3b0fab3c6

### DIFF
--- a/client/vitest.setup.ts
+++ b/client/vitest.setup.ts
@@ -1,5 +1,6 @@
 // Import DOM testing library for Jest, to be used by all test files
 import "@testing-library/jest-dom";
+import { configure } from "@testing-library/dom";
 
 import React from "react";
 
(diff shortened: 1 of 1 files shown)


connor-parke-icf (approved):
i like it! its annoying especially when you have a big DOM area and it truncates the html before reaching the critical piece you need to look at🤦(and then you add a screen.debug(undefined, 999999), but forget to remove it and have to make another commit to remove one line 🤦 🤦 )

curious, does this impact the errors that get thrown from getByRole? or when theres multiple of the same selector and you need to specify/change the selector? the ones that list out elements that could be what you meant to choose

tresdonicf:
> i like it! its annoying especially when you have a big DOM area and it truncates the html before reaching the critical piece you need to look at🤦(and then you add a screen.debug(undefined, 999999), but forget to remove it and have to make another commit to remove one line 🤦 🤦 )
> 
> curious, does this impact the errors that get thrown from getByRole? or when theres multiple of the same selector and you need to specify/change the selector? the ones that list out elements that could be what you meant to choose

Totally! It's such a bummer that it usually doesn't include the relevant part of the tree anyhow 😩 This should match all `getBy`, `findBy`, `getAllBy` and `findAllBy` errors as they all go through `getElementError`. This does give me some idea about some complimentary tooling that could help for a more verbose output (not in the terminal) - I'll look into this when I have some free cycles!

This won't effect comparison type errors though `expected x but got y`

PR #1499, opened 2026-06-26
Demos 2359 phase 8 section completion

connor-parke-icf:
Issue was that we werent passing along the signature level to the update mutators. this worked fine for demonstrations since we dont save that field there anyway, but for amendments and extensions we need that field. While i was at it, i added a few tests to validate that the data is making it to the mutations

### DIFF
--- a/client/src/components/application/phases/approval-summary/ApprovalSummaryPhase.test.tsx
+++ b/client/src/components/application/phases/approval-summary/ApprovalSummaryPhase.test.tsx
@@ -1,20 +1,22 @@
 import "@testing-library/jest-dom";
 import React from "react";
 import { describe, it, expect, vi, beforeEach } from "vitest";
-import { render, screen, within } from "@testing-library/react";
+import { render, screen, waitFor, within } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 
-import { ApprovalSummaryPhase } from "./ApprovalSummaryPhase";
+import {
+  ApprovalSummaryPhase,
+  UPDATE_AMENDMENT_MUTATION,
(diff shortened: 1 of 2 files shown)


client/src/components/application/phases/approval-summary/ApprovalSummaryPhase.tsx
```diff
@@ -357,12 +364,13 @@ export const ApprovalSummaryPhase = ({
...
       });
     }
-    const updateInput: UpdateDemonstrationInput = {
+    const updateInput: UpdateAmendmentInput | UpdateExtensionInput = {
       name: formData.name,
       description: formData.description,
       effectiveDate: formData.effectiveDate
         ? formatDateForServer(parseFormDateString(formData.effectiveDate))
         : null,
+      signatureLevel: formData.signatureLevel,
```
connor-parke-icf, line 373: main fix, we werent saving the signature level, so even though the form was filled saving it wouldnt do anything. now it does

client/src/components/application/phases/approval-summary/ApprovalSummaryPhase.test.tsx
```diff
@@ -320,22 +329,81 @@ describe("ApprovalSummaryPhase", () => {
     expect(screen.getByText(/verify amendment/i)).toBeInTheDocument();
   });
 
+  it("calls updateDemonstration for demonstration", async () => {
```
connor-parke-icf, line 332: added some tests to validate that the form data was actually making it to the mutator. Did this with demos, amendments, and extensions.

dustbuster (approved):
Looks gud

PR #1500, opened 2026-06-26
DEMOS-2357: changed up some triggers/behaviors

connor-parke-icf:
we were seeing some weird edge case behavior in the autocomplete selects, mainly that when you selected an option you had to defocus then refocus to open the menu again. This stops that from happening. In addition, i found that selecting the same already-selected option didnt close the menu, so made it do that while i was at it. 

https://github.com/user-attachments/assets/9b65565c-722e-41f6-82e3-e31653e14716

### DIFF
--- a/client/src/components/input/select/AutoCompleteSelect.tsx
+++ b/client/src/components/input/select/AutoCompleteSelect.tsx
@@ -82,7 +82,14 @@ export const AutoCompleteSelect = ({
     setIsOpen(true);
   };
 
+  const activateSelect = () => {
+    if (!isDisabled) {
+      openSelect();
+    }
+  };
+
   const handleFilterChange = (newFilterValue: string) => {
+    activateSelect();
     setFilterValue(newFilterValue);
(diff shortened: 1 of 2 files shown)


client/src/components/input/select/AutoCompleteSelect.tsx
```diff
@@ -107,6 +114,11 @@ export const AutoCompleteSelect = ({
       openSelect();
       return;
     }
+    if (!isOpen && e.key.length === 1 && !e.altKey && !e.ctrlKey && !e.metaKey) {
```
connor-parke-icf, line 117: not a fan of this section, but needed some way to signify that this is a key press meant for for the search input

dustbuster: There's nothing with a good ole standard "IF" conditional. 
React made it weird, not you! And sometimes that's just what needs to happen.

dustbuster (approved):

PR #1501, opened 2026-06-26
DEMOS-1934: passed in user from context

connor-parke-icf:
checks the usercontext for the current logged in user and uses it to populate the ProjectOfficerUserId field on load.

### DIFF
--- a/client/src/components/dialog/demonstration/CreateDemonstrationDialog.tsx
+++ b/client/src/components/dialog/demonstration/CreateDemonstrationDialog.tsx
@@ -5,15 +5,7 @@ import { useMutation } from "@apollo/client";
 import { CreateDemonstrationInput, Demonstration } from "demos-server";
 import { gql } from "@apollo/client";
 import { DEMONSTRATIONS_PAGE_QUERY } from "pages/DemonstrationsPage";
-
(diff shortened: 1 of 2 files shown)


client/src/components/dialog/demonstration/CreateDemonstrationDialog.tsx
```diff
@@ -70,7 +67,14 @@ export const CreateDemonstrationDialog: React.FC<{
...
       mode="create"
-      initialDemonstration={DEFAULT_DEMONSTRATION_DIALOG_FIELDS}
+      initialDemonstration={{
+        name: "",
+        effectiveDate: "",
+        expirationDate: "",
+        description: "",
+        stateId: "",
+        projectOfficerId: userContext.currentUser.id,
+      }}
```
dustbuster, line 77: I guess this is not re-used then eh?

dustbuster (approved):
Assuming this i not reused and it borked other places, which doesn't look like it is, then GTG!

client/src/components/dialog/demonstration/CreateDemonstrationDialog.tsx
connor-parke-icf: it wasnt, so i figured why not just inline it. IMO its a little easier to parse this way, since one of the fields is dynamically set.

PR #1502, opened 2026-06-26
DEMOS-2342 update submission enabled logic for deliverables

JeffreyM-GA:
This updates the submission enabled logic for deliverables.

- When status is in Past Due or Upcoming, submission is enabled (unless no state files exist)
- When status is in Submitted or Under Review:
- - If there are unsubmitted files, submission is enabled
- - If there are no unsubmitted files, submission is disabled
- When status is in Accepted, Approved, or Received and filed, submission is disabled

### DIFF
--- a/client/src/pages/deliverables/fileAndHistoryTabs/FileAndHistoryTabs.tsx
+++ b/client/src/pages/deliverables/fileAndHistoryTabs/FileAndHistoryTabs.tsx
@@ -41,6 +41,11 @@ export const SUBMIT_DELIVERABLE_MUTATION = gql`
   }
 `;
 
+export const SUBMISSION_ALWAYS_ENABLED_STATUSES: ReadonlySet<DeliverableStatus> = new Set([
+  "Upcoming",
+  "Past Due",
+]);
+
 export const RESUBMISSION_DISABLED_STATUSES: ReadonlySet<DeliverableStatus> = new Set([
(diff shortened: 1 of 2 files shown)


client/src/pages/deliverables/fileAndHistoryTabs/FileAndHistoryTabs.tsx
```diff
@@ -149,7 +161,11 @@ export const FileAndHistoryTabs: React.FC<{
     }
   };
 
-  const isSubmitDisabled = isFinalized || stateFiles.length === 0 || submitLoading;
+  const isSubmitDisabled =
+    isFinalized
```
connor-parke-icf, line 165: technically, with the rules set up the way they are, i dont think we need to check if its finalized or not. 
you can only finalize or be in the finalized status if there are no unsubmitted files, and you can only submit if there are unsubmitted files, and you can only finalize from the Under CMS Review status, so there shouldnt be a scenario where we need to check against status here.  That being said, its not wrong; it may just obfuscate the logic

connor-parke-icf (approved):
LGTM! there might be a way to simplify some of the logic here to better reflect the rules enforced by the API; for instance the check for is finalized or not may be redundant or better replaced with a check that we're _in_ one of the 4 ACTIVE_DELIVERABLE_STATUSES

PR #1503, opened 2026-06-29
DEMOS-2019: defaulted to current user from context

connor-parke-icf:
Similar to creating a new demonstration, the CMS Owner should be defaulted to the current user. Because CMS users are the only ones who can create a new deliverable anyway, we can safely error if something happens when fetching the current user. 

https://github.com/user-attachments/assets/446bd754-7508-4c99-a8af-ea4161e13b68


client/src/components/dialog/deliverable/AddDeliverableSlotDialog.tsx
```diff
@@ -149,11 +146,26 @@ export const AddDeliverableSlotDialog = ({
...
+
+  const initialFormData: AddDeliverableSlotFormData = {
+    deliverableName: "",
+    deliverableType: "" as DeliverableType,
+    scheduleType: "Single",
+    dueDate: "",
+    quarterlyDueDates: ALL_QUARTERS.map(() => ""),
+    demonstrationTypes: [],
+    cmsOwnerUserId: contextUser.currentUser.id,
+  };
```
connor-parke-icf, line 162: instead of this being constant, we can dynamically build this up with the default userId to use as the initialFormData. Then we just pass this into wherever the initial form data was being used before.

dustbuster (approved):
Sure. Sure.

PR #1504, opened 2026-06-29
(1/2) DEMOS-2358: Break components out in `CompletenessPhase`

tresdonicf:
While working on DEMOS-2358 I noticed this component could use some modernization of the sub-components. Broke these out into their own components and made them a bit more pure regarding taking parameters rather than using implicit values in scope - this should benefit both the reliability and maintainability of the component.

### DIFF
--- a/client/src/components/application/phases/completeness/CompletenessNotice.tsx
+++ b/client/src/components/application/phases/completeness/CompletenessNotice.tsx
@@ -1,0 +1,54 @@
+import React, { useMemo, useState } from "react";
+
+import { Notice, NoticeVariant } from "components/notice/Notice";
+import { formatDateForDisplay } from "util/formatDate";
+import { differenceInCalendarDays, parseISO } from "date-fns";
+
+export const CompletenessNotice = ({
+  completenessReviewDate,
+  completenessComplete,
+}: {
+  completenessReviewDate?: string;
+  completenessComplete: boolean;
+}) => {
+  const [isNoticeDismissed, setNoticeDismissed] = useState(
+    !(completenessReviewDate && !completenessComplete)
+  );
+
+  const noticeContent = useMemo(() => {
+    if (!completenessReviewDate) return null;
+    const noticeDueDateValue = parseISO(completenessReviewDate ?? "");
+    const daysLeft = differenceInCalendarDays(noticeDueDateValue, new Date());
+    if (daysLeft > 1) {
+      return {
+        title: `${daysLeft} days left`,
+        description: `This Demonstration must be declared complete by ${formatDateForDisplay(noticeDueDateValue)}`,
+        variant: "warning" as NoticeVariant,
+      };
+    }
+    if (daysLeft === 1) {
+      return {
+        title: "1 day left in Completion Period",
+        description: `This Demonstration must be declared complete by ${formatDateForDisplay(noticeDueDateValue)}`,
+        variant: "error" as NoticeVariant,
+      };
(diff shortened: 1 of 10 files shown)


client/src/components/application/phases/completeness/CompletenessPhase.tsx
```diff
@@ -388,8 +255,21 @@ export const CompletenessPhase = ({
...
+            />
+            <VerifyCompleteSection
+              stateDeemedComplete={stateDeemedComplete}
+              federalStartDate={federalStartDate}
+              federalEndDate={federalEndDate}
+              completenessComplete={completenessComplete}
+              finishIsEnabled={!!finishIsEnabled()}
+              onDateChange={setUserSelectedStateDeemedCompleteDate}
+              onDeclareIncomplete={handleDeclareIncomplete}
+              onFinish={handleFinishCompleteness}
```
connor-parke-icf, line 271: any more dates used in this component and id suggest bundling them together in a records object or extracting the date values from the application in the subcomponent itself (in the effort of separating the "data used in the component" from the "component configuration"), but it isnt an overwhelming number of props as it is

james-becwar-cms (approved):

client/src/components/application/phases/completeness/CompletenessNotice.tsx
```diff
@@ -0,0 +1,54 @@
...
+  if (isNoticeDismissed || !noticeContent) return null;
+
+  return (
+    <Notice
+      title={noticeContent.title}
+      description={noticeContent.description}
+      variant={noticeContent.variant}
+      onDismiss={() => setNoticeDismissed(true)}
+    />
+  );
```
connor-parke-icf, line 53: NB/suggestion: instead of checking for the validity of the input within this component, we might can delegate that to the caller component. That way, this component's job is always just to format the Notice and can assume that if its called, it has valid data to format. 

Might just be {completenessReviewDate && <CompletenessNotice />}

client/src/components/application/phases/completeness/UploadSection.tsx
```diff
@@ -0,0 +1,41 @@
+import React from "react";
+
+import { SecondaryButton } from "components/button";
+import { ExportIcon } from "components/icons";
+import { ApplicationWorkflowDocument } from "components/application";
+import { DocumentList } from "../sections";
+import {
+  COMPLETENESS_PHASE_STEP_ONE_DESCRIPTION,
+  COMPLETENESS_UPLOAD_BUTTON_NAME,
```
connor-parke-icf, line 9: NB: since these constants are only used here and the associated test (i think), might can move their definitions to this file.

connor-parke-icf (approved):
so much better; i definitely felt the size of this component before so its nice that its broken up a good bit now. Couple of organizational suggestions, but certainly nothing blocking (i suspect some of my thoughts here are designated to be in the next part, so this PR is just moving things around)

tresdonicf:
> so much better; i definitely felt the size of this component before so its nice that its broken up a good bit now. Couple of organizational suggestions, but certainly nothing blocking (i suspect some of my thoughts here are designated to be in the next part, so this PR is just moving things around)

All good ideas! I really did just want to move things in the most direct way to not affect any functionality at this point (so these were pretty much just lifted + passing in everything the component needed) but I agree big time! After fixing the core issue I'll look to clean up some of the individual components, it's likely they can do some of the calculation of the props they're currently getting passed in

PR #1505, opened 2026-06-29
Demos 2323: Set MinDate attribute to allow for only today or Greater dates for delv create - TEMP WORK AROUND

dustbuster:
Added F/E error messages and submit disabled handling for past date deliverables.
So when adding a deliverable it must be today or later
Okay. so in the course of doing [DEMOS-2323](https://jiraent.cms.gov/browse/DEMOS-2323) (Adding deliverable slots and validating dates are in the future)

We have discovered a couple things in relation to dates the frontend. [@Tresdon Jones](https://globalalliantinc.slack.com/team/U097PJYDXK5) of course was someone i could run this by and verify that i'm not crazy! (Spoiler i am, but not in this case)

If your component is set to use minDate={today} , where the date cannot be before today,
the expectation is that it will not allow you to set a date in the past.

But what we found is that if you set that date, and then change it. It only validates the first date and it actually submits that first date because it's not communicating the changes with the parent component...

But i did find a fix.
A very simple change to datepicker. However. Since datepicker is used 75 times in 22 files. I do not want to change it until i'm 100% sure this does not have cascading effects.

this:
`if (newDate === "" || (newDate >= minDate && newDate <= maxDate)) {`
becomes that:
`if (newDate === "" || newDate >= FULL_YEAR_DATE) {`

But for now i get a fix in there. We just lose the disable past date. as i can't have in there dates that are not what the user is seeing! 

But if typed in that datepicker will throw the warning message and tell you that you need a date >= today!

Screenshot 2026-06-29 at 10 45 19


client/src/components/dialog/deliverable/fields/schedule-type/QuarterlyDeliverableSchedule.tsx
```diff
@@ -97,6 +108,7 @@ export function QuarterlyDeliverableSchedule({
           value={getQuarterDueDate(3)}
           onChange={(newDate: string) => onSelectQuarterDate?.(3, newDate)}
           isRequired={true}
+          getValidationMessage={() => getQuarterValidationMessage(3)}
```
dustbuster: Just so you know. the numbers equate to quarters. 

```JavaScript
quarterlyDueDates[0] // 1st Quarter
quarterlyDueDates[1] // 2nd Quarter
quarterlyDueDates[2] // 3rd Quarter
quarterlyDueDates[3] // 4th Quarter
```

client/src/components/dialog/deliverable/fields/schedule-type/QuarterlyDeliverableSchedule.tsx
```diff
@@ -36,8 +39,13 @@ export function QuarterlyDeliverableSchedule({
   onSelectQuarterDate?: (quarterIndex: number, dueDate: string) => void;
 }) {
   const [selectedYear, setSelectedYear] = useState(1);
+  const today = getTodayEst();
```
james-becwar-cms: Nice

james-becwar-cms (approved):

client/src/components/dialog/deliverable/fields/schedule-type/QuarterlyDeliverableSchedule.tsx
```diff
@@ -86,6 +90,7 @@ export function QuarterlyDeliverableSchedule({
           value={getQuarterDueDate(2)}
           onChange={(newDate: string) => onSelectQuarterDate?.(2, newDate)}
           isRequired={true}
+          minDate={today}
```
dustbuster: Simplier!

dustbuster: This will come back after completing [DEMOS-2366](https://jiraent.cms.gov/browse/DEMOS-2366)

PR #1506, opened 2026-06-30
Demos 2369 aria label and testid

connor-parke-icf:
simple cleanup. Replaces instances of ariaLabel or similar spellings in our components' props with aria-label.  Then does similar with data-testId, making it all lowercase and where used in a prop like testId, changed it to data-testid as well. 

Its kinda strange, for most props typescript will throw an error if the prop doesnt exist on the called component. However, these hyphenated props dont flag. This leads to scenarios where we might attempt to pass in a "aria-label" but the prop is dropped without any indication that something is wrong. 

I had tried to figure out if there was a way to make TS be a little more explicit, and perhaps there is, but given the scope of this cleanup id rather just do the string substitution. So looked at wherever i could find we were using these attributes and made them consistent with the aria/html standard. 

the other one, data-testId, had a similar problem, but we were getting console errors because of it. apparently, it needs to be spelled "data-testid", all lowercase, even though the testing framework doesnt care. 

Finally, im not entirely sure if this is the best way to go; personally i kinda like making it explicitly not a hyphenated prop, so that the typechecking is enforced. But without custom linting rules or something, theres no way to enforce "dont use the hyphenated prop here". So rather than go down that rabbithole, lets just make it consistent for now

### DIFF
--- a/client/src/components/table/PaginationControls.test.tsx
+++ b/client/src/components/table/PaginationControls.test.tsx
@@ -159,7 +159,7 @@ describe("PaginationControls", () => {
       const data = createTestData(100);
       render(<TestWrapper data={data} />);
 
-      const prevButton = screen.getByRole("button", { name: /Go to previous page/i });
+      const prevButton = screen.getByRole("button", { name: /No previous page/i });
       expect(prevButton).toBeDisabled();
     });
 
@@ -244,7 +244,7 @@ describe("PaginationControls", () => {
       const data = createTestData(100);
       render(<TestWrapper data={data} />);
 
-      const currentPageButton = screen.getByRole("button", { name: "Go to page 1" });
+      const currentPageButton = screen.getByRole("button", { name: "Page 1, current page" });
       expect(currentPageButton).toBeInTheDocument();
       expect(currentPageButton).toHaveClass("bg-action");
     });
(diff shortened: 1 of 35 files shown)


james-becwar-cms (approved):

client/src/components/table/PaginationControls.test.tsx
tresdonicf, line 1: I'm not certain why this test file is changed here (unless perhaps we change them to use `data-testId` instead of `getByRole`

tresdonicf: I just imagine this test might fail unless there are accompanying changes to the component and since we're talking about the eslint 10 exception don't want this to fall through the cracks if we bypass checks and merge

connor-parke-icf: good catch; i shoulda mentioned that change. 
```
         <Button
              key={page}
              name={"Go to page " + (page + 1)}
              onClick={() => setPageIndex(page)}
              aria-label={`Page ${page + 1}, current page`}
            >
```
previously, the aria-label here was being dropped. With the updates to Button though, it gets picked up as the name of the button in the test. So updated the test here to point to the new label. 

_could_ update it to use data-testid now, but if im doing that id rather do that in a dedicated pass for this component.

tresdonicf (approved):

PR #1508, opened 2026-06-30
DEMOS-2366 & DEMOS-2372 DatePicker updates and testing datepicker and safari adjustments

dustbuster:
I feel very confident about this after a solid 12 hrs of testing. 

But i made the changes to Datapicker and then ran thru this with a few browsers and ran into issues with mostly Safari. So both tickets ran into this. Which is just a consequence of the the first ticket. 

Before: 
Screenshot 2026-06-30 at 10 05 59

After: 
Screenshot 2026-06-30 at 11 32 27

Safari datepicker disabled:
Screenshot 2026-06-30 at 16 54 12

Chrome:
Screenshot 2026-06-30 at 16 55 05

### DIFF
--- a/client/src/components/application/phase-selector/PhaseBox.tsx
+++ b/client/src/components/application/phase-selector/PhaseBox.tsx
@@ -6,7 +6,7 @@ import { PhaseDate } from "./PhaseDate";
 import { PhaseName } from "demos-server";
 
 const BASE_STYLES = {
-  PHASE_BOX: tw`relative flex flex-col items-center justify-center rounded-sm hover:cursor-pointer aspect-2/1 p-1`,
+  PHASE_BOX: tw`relative flex flex-col items-center justify-center rounded-sm hover:cursor-pointer aspect-2/1 w-full min-w-0 p-1`,
   PHASE_NUMBER: tw`flex items-center justify-center text-lg w-3 h-3 my-1 rounded-full font-bold`,
   PHASE_NAME: tw`text-[12px] font-bold truncate max-w-full`,
   AI_SUGGESTION_ICON: tw`absolute right-[5px] top-[5px] text-purple-700 size-[1.8em]`,
@@ -57,7 +57,7 @@ export const PhaseBox = (props: PhaseBoxProps) => {
   const showSuccessIcon = isCompletionStatus(props.phaseStatus);
 
   return (
-    <div className="flex flex-col justify-center col-span-1">
+    <div className="flex flex-col justify-center col-span-1 min-w-0">
       <div
         key={props.phaseName}
         className={`${BASE_STYLES.PHASE_BOX} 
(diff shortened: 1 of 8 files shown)


client/src/components/application/phase-selector/PhaseBox.tsx
```diff
@@ -57,7 +57,7 @@ export const PhaseBox = (props: PhaseBoxProps) => {
   const showSuccessIcon = isCompletionStatus(props.phaseStatus);
 
   return (
-    <div className="flex flex-col justify-center col-span-1">
+    <div className="flex flex-col justify-center col-span-1 min-w-0">
```
dustbuster, line 60: Tested in Safari, Firefox, and Chrome. These changes only affect Safari.

dustbuster: Image

dustbuster: Firefox!

Image

client/src/components/dialog/deliverable/fields/schedule-type/QuarterlyDeliverableSchedule.tsx
```diff
@@ -103,7 +102,7 @@ export function QuarterlyDeliverableSchedule({
           value={getQuarterDueDate(3)}
           onChange={(newDate: string) => onSelectQuarterDate?.(3, newDate)}
           isRequired={true}
-          getValidationMessage={() => getDueDateValidationMessage(getQuarterDueDate(3))}
+          minDate={today}
```
dustbuster, line 105: This. Is how datepicker SHOULD be. Forget all that custom logic.

client/src/components/input/date/DatePicker.tsx
```diff
@@ -46,9 +46,9 @@ export const DatePicker = ({
     const newDate = e.target.value;
     setDisplayedDate(newDate);
 
-    // Only call onChange if the date is valid (empty or within range).
-    // (newDate === "" || newDate >= FULL_YEAR_DATE) <-- change that fixes min date validation.
-    if (newDate === "" || (newDate >= minDate && newDate <= maxDate)) {
+    // Browser date inputs can emit partial-year values while editing the year subfield.
+    // Complete dates should propagate even when out of range so callers can validate current state.
+    if (newDate === "" || newDate >= FULL_YEAR_DATE) {
```
dustbuster, line 51: I've been testing this for like a day, it works great!

tresdonicf (approved):

client/src/components/input/date/DatePicker.tsx
connor-parke-icf: guessing that full year date here is just the furthest back in time you can go and still have a four digit year? like 1000-01-01. Seems like a clever solution, with this im guessing so long as its a valid date format, it will process the on change event and then any validation can happen

connor-parke-icf (approved):
lgtm!

client/src/components/input/date/DatePicker.tsx
dustbuster: This says, as long as this date is AFTER 1/1/1000, then lets pass it to the parent. i lowered the bar for passing to parent. That's it. And it has the appearance of being more reactive and i think it is. This just removes a barrier for entry!

PR #1510, opened 2026-06-30
(2/3) DEMOS-2358: Further cleanup / organization of completeness phase

tresdonicf:
Working through DEMOS-2358 and bringing this component to the modern era, this one breaks out props that can be calculated on their own in the component itself as well as shifting testing around to match what each component is responsible for now

### DIFF
--- a/client/src/components/application/phases/completeness/CompletenessPhase.tsx
+++ b/client/src/components/application/phases/completeness/CompletenessPhase.tsx
@@ -1,18 +1,9 @@
-import React, { useState } from "react";
+import React from "react";
 
 import { tw } from "tags/tw";
-import { formatDateForServer, getDateEst } from "util/formatDate";
-import { addDays, parseISO } from "date-fns";
+import { getDateEst } from "util/formatDate";
 import { WorkflowApplication, ApplicationWorkflowDocument } from "components/application";
-import { useToast } from "components/toast";
-import { ApplicationDateInput, DateType, LocalDate, PhaseName, PhaseStatus } from "demos-server";
-import { useDialog } from "components/dialog/DialogContext";
-import { useSetApplicationDates } from "components/application/date/dateQueries";
-import { getPhaseCompletedMessage, SAVE_FOR_LATER_MESSAGE } from "util/messages";
-import {
-  useCompletePhase,
-  useDeclareCompletenessPhaseIncomplete,
-} from "components/application/phase-status/phaseCompletionQueries";
+import { DateType, PhaseName, PhaseStatus } from "demos-server";
 import { CompletenessNotice } from "./CompletenessNotice";
 import { UploadSection } from "./UploadSection";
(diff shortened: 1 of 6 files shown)


client/src/components/application/phases/completeness/UploadSection.tsx
```diff
@@ -12,30 +13,32 @@ import {
 export const UploadSection = ({
   applicationId,
   completenessDocuments,
-  onUploadClick,
 }: {
   applicationId: string;
   completenessDocuments: ApplicationWorkflowDocument[];
-  onUploadClick: (applicationId: string) => void;
```
tresdonicf, line 19: For instance, we can just have this component open up the dialog itself rather than receive a callback to open it up

client/src/components/application/phases/completeness/VerifyCompleteSection.tsx
tresdonicf, line 1: Moved the main logic of dates / button states to this component

client/src/components/application/phases/completeness/CompletenessPhase.tsx
tresdonicf, line 1: Able to remove a lot of stuff from the top-level component, now it concerns itself mostly with initializing sub-components

client/src/components/application/phases/completeness/VerifyCompleteSection.tsx
```diff
@@ -12,83 +23,205 @@ import {
...
-      </div>
+  applicationId: string;
+  stateDeemedCompleteDate: string;
+  completenessPhaseStatus: PhaseStatus;
+  completenessDocuments: ApplicationWorkflowDocument[];
+  applicationIntakeComplete: boolean;
+  setSelectedPhase: (phase: PhaseName) => void;
+}) => {
+  const completenessComplete = completenessPhaseStatus === "Completed";
+  const completenessIncomplete = completenessPhaseStatus === "Incomplete";
```
connor-parke-icf, line 76: might can remove these variables since their def is so small, and theyre mutually exclusive to eachother. might can just inline the comparison where used

connor-parke-icf (approved):
not much to say here, this all looks great! big fan of reducing the props and passing through only whats needed and moving around the tests to match.  Maybe some of the helper functions can get inlined since many of them are so lightweight.  Maybe also we could pass in a Pick<Application, "field" | "field" | "field"> object to the subcomponents instead of separate props for the phase status, documents, previous phase status, etc. but honestly, not sure if itd make it easier or harder to parse

client/src/components/application/phases/completeness/VerifyCompleteSection.tsx
tresdonicf: totally agree!

tresdonicf: I looked into it a bit more and it seems like there's a bit more usages for these than I initially thought. I might hold off on it for now but overall good insight!

PR #1511, opened 2026-06-30
DEMOS-2319-migrated-user-path

TARehman:
This PR adds a new path for migrated users to be linked when they come to the application from IDM / Cognito. The workflow is as below, which starts in `findOrCreateContextUserFromClaims` in `userContext.ts`.

- Initially, we attempt to find a claim from a user based on their Cognito subject (`sub`). If this works, the process ends - the user is already associated with a Cognito subject so no action is needed.
- If we cannot find the user by Cognito subject, we attempt to locate a user record by email among users who a) are migrated from PMDA and b) have not logged in before.
  - If no users are found, we move on to create a new user like we did before.
  - If more than one user is found, we throw an error and log it.
  - If exactly one user is found, we verify that the person_type_id matches.
    - If it does, we update the DB to associate the incoming claims with that user, including updating the stored data to match the claims.
    - If it does not, we throw an error and log it.

The code also updates how throwing happens within the `auth` module. It seems that because the logging plugin isn't fully loaded at the time this code works, the thrown errors are not automatically logged out like the rest of the codebase. So I found everywhere in the `auth` module where we were throwing, and ensured that we explicitly logged the error messages. I also updated the codes throughout to mask what was happening.

There's also a bunch of small stuff like query functions that all exists to support the broader workflow.


server/src/auth/user/userContext.ts
```diff
@@ -2,16 +2,17 @@ import type { AuthorizationClaims } from "..";
...
 ): Promise<ContextUser> {
   return await prisma().$transaction(async (tx) => {
-    const returnedUser =
-      (await findUserByCognitoSubject(claims.sub, tx)) ??
-      (await createNewUserFromClaims(claims, tx));
-    await upsertUserSession(returnedUser.id, claims.authTime, tx);
-    return returnedUser;
+    let result: ContextUser | null;
+    result = await findUserByClaims(claims, tx);
+    result ??= await linkNewlyMigratedUserFromClaims(claims, tx);
```
connor-parke-icf, line 13: nullish coalescing assignment operator in the wild!

server/src/auth/auth.util.ts
```diff
@@ -18,30 +20,59 @@ export type AuthorizationClaims = {
...
   claims: Partial<AuthorizationClaims>
 ): asserts claims is AuthorizationClaims {
+  let errorMessage: string;
+  let errorCode: CustomInternalErrorCode;
   if (!claims.email) {
-    throw new Error("Authorizer claims missing required 'email' field");
+    errorCode = "CLAIM_VALIDATION_EMAIL_ERROR";
+    errorMessage = `Error Code ${errorCode}: Authorizer claims missing required 'email' field.`;
+    log.error(errorMessage);
+    throwCustomGQLError(errorMessage, errorCode);
```
connor-parke-icf, line 29: nice, glad to see these logged and reported back to the user better

server/src/model/deliverable/validateDeliverableInputs.ts
```diff
@@ -155,10 +155,12 @@ export async function validateCompleteDeliverableInput(
...
-  // Deliverables may only be completed from review status
-  errors.push(checkDeliverableHasStatus(deliverable, ["Under CMS Review"]));
-  // Deliverables may only be completed if there are no unsubmitted state documents
-  errors.push(await checkDeliverableHasNoUnsubmittedStateDocuments(deliverable, tx));
+  errors.push(
+    // Deliverables may only be completed from review status
+    checkDeliverableHasStatus(deliverable, ["Under CMS Review"]),
+    // Deliverables may only be completed if there are no unsubmitted state documents
+    await checkDeliverableHasNoUnsubmittedStateDocuments(deliverable, tx)
+  );
```
connor-parke-icf, line 163: nice, might as well

connor-parke-icf (approved):
lgtm. Discussed the error logging enhancements directly. Other than that, i like the separation of the findNewlyMigrateUserByEmail return values; makes the calling function very simple. 

I assume that because the personType of the user shouldnt change when they log in, this shouldnt prompt any cascading affects like making a CMS user no longer able to be a project officer or something

server/src/model/deliverable/validateDeliverableInputs.ts
TARehman: Was a "code smell" when I ran through SonarQube manually.

PR #1513, opened 2026-07-01
DEMOS-2356 add loading spinner for report downloads

JeffreyM-GA:
Adds a spinner for report downloads

https://github.com/user-attachments/assets/656629af-0d36-41a0-9f43-ae65692e54a1

### DIFF
--- a/client/src/components/table/tables/ReportsTable.tsx
+++ b/client/src/components/table/tables/ReportsTable.tsx
@@ -24,13 +24,13 @@ export const AVAILABLE_REPORT_TYPES = [
 ] as const;
 
 export const ReportsTable: React.FC = () => {
-  const [currentDownload, setCurrentDownload] = useState<string | null>(null);
+  const [downloadingReports, setDownloadingReports] = useState<Set<string>>(new Set());
   const { triggerDownload } = useTriggerDownload();
   const { showSuccess, showError } = useToast();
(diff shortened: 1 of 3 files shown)


client/src/components/table/tables/ReportsTable.tsx
```diff
@@ -24,11 +24,14 @@ export const AVAILABLE_REPORT_TYPES = [
 ] as const;
 
 export const ReportsTable: React.FC = () => {
+  const [currentDownload, setCurrentDownload] = useState<string | null>(null);
```
connor-parke-icf: if one were to download the reports in quick succession, would this cause issues since we're only tracking one report at a time for the whole table? 

Looks like we disable the other buttons when a download is in progress, so i imagine its not an issue in the current design. Perhaps we could move all the downloading logic into the download button itself, making a component for that. But as it works now i wouldnt change it, just a possible area for improvement if we have a bunch of reports the user might want to download all at once.

connor-parke-icf (approved):
LGTM!

client/src/components/table/tables/ReportsTable.tsx
JeffreyM-GA: Yeah actually I had restricted it to only allow one download at a time, I just updated the logic to track the downloads separately so they can be downloaded synchronously

PR #1532, opened 2026-07-06
DEMOS-2021_added_set_person_states_mutator

connor-parke-icf:
post mvp feature for admin functionalility. This includes the ability for an admin to assign people to states by introducing a setPersonStates mutator. Includes typical validation to prevent failing on FK constraints in the database. Only State Users can have their states modified, as the other user types are already automatically assigned to all states. Only admin can access this mutation. Also, prevents the unassignment of a state that is used in a demonstrationRoleAssignment. 

https://github.com/user-attachments/assets/178d946f-0032-42e5-996c-e75b5025c1dd

EDIT: updated branch to be post-launch


server/src/model/person/personSchema.ts
```diff
@@ -16,9 +16,14 @@ export const personSchema = gql`
     states: [State!]!
     roles: [DemonstrationRoleAssignment!]!
   }
+
   type Query {
     people: [Person!]! @auth(requires: ["Access CMS Query"])
   }
+  type Mutation {
+    setPersonStates(personId: ID!, stateIds: [String!]!): Person!
```
connor-parke-icf, line 24: thought about making the stateIds input a more formal scalar in the gql, but decided to hold off on that until we address a refactor on how we handle state objects in the server. Ideally, we could replace all instances of the state object with a StateId scalar. For now, just using string.

server/src/model/personState/validatePersonNotAssignedToDemonstrationOfStates.ts
```diff
@@ -0,0 +1,26 @@
+import { PrismaTransactionClient } from "../../prismaClient";
+import { selectManyDemonstrationRoleAssignments } from "../demonstrationRoleAssignment/queries";
+import { State } from "../state/stateSchema";
```
TARehman: Shouldn't we export this through `types.ts` and then import it from there, rather than directly like this? Also, not anything that blocks but I've been slowly trying to move toward importing types using `import type {}` and using `import {}` on the non-type stuff to help make it more clear what's going on, we might want to make a guideline at some point.

server/src/model/personState/validatePersonNotAssignedToDemonstrationOfStates.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export async function validatePersonNotAssignedToDemonstrationOfStates(
```
TARehman: I think we would probably call this a "check" function and then the "validate" is the thing that stitches them all together, but that's nits, not worth changing. To me the only item here is that this function's name takes a pass or two to read - but I don't know what else you'd do.

server/src/model/personState/validatePersonNotAssignedToDemonstrationOfStates.test.ts
```diff
@@ -0,0 +1,56 @@
...
+      },
+      mockTransaction
+    );
+  });
+
+  it("should return an error message if the person is assigned to demonstrations of the states", async () => {
+    vi.mocked(selectManyDemonstrationRoleAssignments).mockResolvedValue([
+      { stateId: "NY" },
+      { stateId: "CA" },
+    ] as never);
```
TARehman: This feels weird. Why not mock this properly? Or do a `Partial` like we do elsewhere?

server/src/model/personState/validatePersonIsStateUser.ts
```diff
@@ -0,0 +1,13 @@
+import { PrismaTransactionClient } from "../../prismaClient";
+import { PersonType } from "../../types";
+import { selectPersonOrThrow } from "../person/queries";
+
+export async function validatePersonIsStateUser(
+  personId: string,
+  tx: PrismaTransactionClient
+): Promise<string | undefined> {
+  const person = await selectPersonOrThrow({ id: personId }, tx);
```
TARehman: Not worth changing, but couldn't you have just selected from the `person` table and filtered by the person type, and then returned the message if the result was empty?

server/src/model/personState/validatePersonIsStateUser.test.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  beforeEach(() => {
+    vi.resetAllMocks();
+  });
+
+  it("should return undefined if the person is a state user", async () => {
+    vi.mocked(selectPersonOrThrow).mockResolvedValue({
+      id: testPersonId,
+      personTypeId: "demos-state-user",
+    } as never);
```
TARehman: Same, as my other comment; this feels weird.

server/src/model/personState/setPersonStates.ts
```diff
@@ -0,0 +1,66 @@
...
+import { validatePersonIsStateUser } from "./validatePersonIsStateUser";
+
+export const setPersonStates = async (
+  personId: string,
+  stateIds: State["id"][]
+): Promise<PrismaPerson> => {
+  return prisma().$transaction(async (tx) => {
+    const personTypeValidationResult = await validatePersonIsStateUser(personId, tx);
+    if (personTypeValidationResult) {
+      cleanErrorsAndThrow(
```
TARehman, line 18: NB: We should probably integrate the clean and throw stuff with the custom error stuff added elsewhere, but not important for this PR.

server/src/model/personState/setPersonStates.ts
```diff
@@ -0,0 +1,66 @@
...
+      },
+    }).then((states) => states.map((state) => state.id));
+
+    const statesToRemove = currentStateIds.filter((stateId) => !stateIds.includes(stateId));
+    const statesToAdd = stateIds.filter((stateId) => !currentStateIds.includes(stateId));
+
+    const demonstrationAssignmentValidationResult =
+      await validatePersonNotAssignedToDemonstrationOfStates(personId, statesToRemove, tx);
+    if (demonstrationAssignmentValidationResult) {
+      cleanErrorsAndThrow(
```
TARehman, line 39: I think the bigger item I see is that strictly speaking, we intended to have that function get called once with a list of all the validations that failed. _That being said_, it doesn't make sense to keep going when someone isn't a state user, so I get it.

server/src/model/personState/setPersonStates.ts
```diff
@@ -0,0 +1,66 @@
...
+      );
+    }
+
+    const currentStateIds = await selectManyStates({
+      personStates: {
+        some: {
+          personId,
+        },
+      },
+    }).then((states) => states.map((state) => state.id));
```
TARehman, line 31: This usage of `.then()` is redundant I think, because you're already awaiting the result above.

server/src/model/personState/setPersonStates.ts
```diff
@@ -0,0 +1,66 @@
...
+import { selectPersonOrThrow } from "../person/queries";
+import { selectManyStates } from "../state/queries";
+import { State } from "../state/stateSchema";
+import { deleteManyPersonStates, insertManyPersonStates } from "./queries";
+import { prisma } from "../../prismaClient";
+import { validatePersonNotAssignedToDemonstrationOfStates } from "./validatePersonNotAssignedToDemonstrationOfStates";
+import { cleanErrorsAndThrow } from "../../errors/cleanErrorsAndThrow";
+import { validatePersonIsStateUser } from "./validatePersonIsStateUser";
+
+export const setPersonStates = async (
```
TARehman, line 11: Because of the complexity of state assignments in terms of downstream constraints, I think I probably wouldn't have done a `set` operator here, and instead, I would have done a simple add and remove mutator. This is also true because in the overwhelming majority of cases a person will have one and only one state (if they're a state user), so the ability to do a set operation on a bunch of stuff isn't going to be helpful usually.

server/src/model/personState/setPersonStates.ts
```diff
@@ -0,0 +1,66 @@
...
+    if (statesToAdd.length > 0) {
+      await insertManyPersonStates(
+        statesToAdd.map((stateId) => ({
+          personId,
+          stateId,
+        }))
+      );
+    }
+
+    return selectPersonOrThrow({ id: personId });
```
TARehman, line 64: NB: Might be nice to toss a warning or info message if they send a payload that doesn't change anything.

server/src/model/personState/queries/insertManyPersonStates.ts
```diff
@@ -0,0 +1,10 @@
+import { Prisma } from "@prisma/client";
+import { prisma, PrismaTransactionClient } from "../../../prismaClient";
+
+export async function insertManyPersonStates(
+  data: Prisma.PersonStateCreateManyInput[],
+  tx?: PrismaTransactionClient
+): Promise<void> {
+  const prismaClient = tx ?? prisma();
```
TARehman, line 8: NB: I ran your stuff through SQ and technically because you don't have a path that tests the creation of the client if one isn't passed in, you don't have full coverage. It's boilerplate elsewhere so I don't think its the end of the world but more an FYI.

TARehman (approved):
There's some stuff I would have done a little differently but it all seems logically sound, no concerns. 👍🏻

server/src/model/personState/validatePersonNotAssignedToDemonstrationOfStates.ts
connor-parke-icf: theres actually an eslint rule about this somewhere; i thought i had implemented that at some point, but maybe not? either way nice catch; ill update

server/src/model/personState/validatePersonNotAssignedToDemonstrationOfStates.ts
connor-parke-icf: i struggled with the name for a little bit haha. went with a more explicit one.  I can rename this to check, that convention makes sense to me.

server/src/model/personState/validatePersonNotAssignedToDemonstrationOfStates.test.ts
connor-parke-icf: good catch, that as never doesnt make sense

server/src/model/personState/validatePersonIsStateUser.ts
connor-parke-icf: oh yeah... yeah i like that better, ill change it. then the validation for whether or not the person exists in the first place gets absorbed more steamlined

server/src/model/personState/setPersonStates.ts
connor-parke-icf: oh, i thought this was the new stuff. Will followup with you on the new means, since i want to implement this with the more fleshed out version

server/src/model/personState/setPersonStates.ts
connor-parke-icf: yeah, kinda awkward since the validation checks kinda feed into eachother here, rather than in parallel. Might refactor

server/src/model/personState/setPersonStates.ts
connor-parke-icf: its the equivalent of wrapping the await and the function together in parentheses. I think i prefer the .then in this case though since it leads to less nesting

```
  const currentStateIds = await selectManyStates({
      personStates: {
        some: {
          personId,
        },
      },
    }).then((states) => states.map((state) => state.id));
```
is the same as 
``` 
const currentStateIds = (await selectManyStates({
      personStates: {
        some: {
          personId,
        },
      },
    })).map((state) => state.id));
```
the postfix operation takes precedence over the await

server/src/model/personState/setPersonStates.ts
connor-parke-icf: hmm... i think youre probably right. in 95% of the cases, we're probably just adding a state, removing a state, or changing a state. Its odd cause the concept of belonging to multiple states is itself shakey. Im not entirely sure which is best. I think for now, im going to leave it as is, but im certainly open to refactoring this later when we have a more concrete usecase/consumer

server/src/model/personState/setPersonStates.ts
connor-parke-icf: i dont think itd hurt anything to do so, but i dont think its worth adding at this point; after all, a no-op is perfectly reasonably given the idempotent semantic of the endpoint.  It wont get flagged by the redundant update catch though, so maybe that diverges from behavior we have elsewhere? idk

server/src/model/personState/queries/insertManyPersonStates.ts
connor-parke-icf: oh, nice catch; i can add that

server/src/model/personState/setPersonStates.ts
connor-parke-icf: opting to hold off on this, since the scope surrounds all of our common validation pattern, and should have its own story for it

server/src/model/personState/setPersonStates.ts
connor-parke-icf: eh, didnt like going forward with a selectManyStates call if they werent a state user and piping that through a filter operation. So going to leave this as is, at least for now.

server/src/model/personState/validatePersonNotAssignedToDemonstrationOfStates.test.ts
connor-parke-icf: resolved to cast to the actual query result Type

server/src/model/personState/validatePersonNotAssignedToDemonstrationOfStates.ts
connor-parke-icf: renamed this to be "check" instead of "validate" to hold the convention

server/src/model/personState/validatePersonNotAssignedToDemonstrationOfStates.ts
connor-parke-icf: went through each of the files i touched and inserted "type" to the imports where i could, then made sure i was importing "State" from types.ts

server/src/model/personState/validatePersonIsStateUser.ts
connor-parke-icf: updated to use selectPerson, passing in personTypeId as a filter, then checking whether theres a person or not after

PR #1534, opened 2026-07-07
DEMOS-2378 Generate sanitized, human-readable download file names

nunesmlucas:
Files are stored in S3 under their UUID, but users expect a human-readable name
when downloading. This wires the presigned **download** URL to return a file with
a clean, valid name and the correct extension — for documents, references,
reference agreements, and on-demand reports (not just the document preview page).

## What it does

1. **Human-readable names via Content-Disposition** — `getPresignedDownloadUrl`
   accepts a desired file name and sets `ResponseContentDisposition` so the
   download comes back named, instead of a raw UUID / S3 object key.
2. **Sanitized names** — the name is cleaned with `sanitize-filename` before use.
   If a name is comprised entirely of invalid characters, it falls back to the
   object's UUID/key.
3. **Correct extension** — the extension is derived from the object's stored S3
   `Content-Type` (via `mime-types`) and appended, so names passed in are always
   extension-less and the download still opens correctly.

## Call pattern

Every caller passes a bare, extension-less name; the adapter adds the extension
from Content-Type. The only thing that varies is the disposition:

| Caller | Disposition | Notes |
| --- | --- | --- |
| Documents | `inline` (default) | Keeps the in-page PDF `<embed>` preview working |
| References / agreements / reports | `attachment` | Direct downloads |

## Key changes

- `adapters/s3/sanitizeDownloadFileName.ts` *(new)* — shared sanitize + UUID-fallback helper
- `adapters/s3/AwsS3Adapter.ts` — sanitize → read Content-Type → append extension → Content-Disposition
- `adapters/s3/S3Adapter.ts` — `GetPresignedDownloadUrlOptions` (`disposition`) + updated signature
- `adapters/s3/LocalS3Adapter.ts` — signature kept in sync with the AWS adapter
- `model/reference/getReferenceDownloadUrl.ts`, `getReferenceAgreementDownloadUrl.ts` — pass the reference/agreement name
- `model/onDemandReport/onDemandReportResolvers.ts` — pass the report type as the name

## Notes / edge cases

- **Document preview page** was verified for collisions: it never appends an
  extension, so the new server-side extension logic won't produce `name.pdf.pdf`.
- Document names are stored **without** an extension (the client strips it on
  upload), which is why the extension is derived from Content-Type here.
- A user *could* manually type an extension into a document name (free-text
  field), which would produce `report.pdf.pdf`. This is the rare, accepted edge
  the ticket calls out.

## Dependencies added

- `sanitize-filename`
- `mime-types` + `@types/mime-types`

## Testing

- `tsc --noEmit`, ESLint clean
- Unit tests added/updated for the sanitizer, both S3 adapters, and the
  reference/report resolvers — all passing

### DIFF
--- a/server/package-lock.json
+++ b/server/package-lock.json
@@ -28,8 +28,10 @@
         "graphql-tools": "^9.0.28",
         "jsonwebtoken": "^9.0.3",
         "jwks-rsa": "^3.2.2",
+        "mime-types": "^3.0.2",
         "pg": "^8.21.0",
         "pino": "^10.3.1",
+        "sanitize-filename": "^1.6.4",
         "zod": "^4.4.3"
       },
       "devDependencies": {
@@ -40,6 +42,7 @@
         "@faker-js/faker": "^9.9.0",
         "@graphql-tools/merge": "^9.1.9",
         "@parcel/watcher": "^2.5.6",
+        "@types/mime-types": "^3.0.1",
         "@types/node": "^22.19.20",
         "@vitest/coverage-v8": "^4.1.8",
         "@vitest/ui": "^4.1.8",
@@ -66,7 +69,7 @@
         "read-excel-file": "^8.0.3"
       },
       "devDependencies": {
-        "@types/node": "^25.9.3",
+        "@types/node": "^25.9.4",
         "typescript": "^5.9.3",
         "vitest": "^4.1.9"
       }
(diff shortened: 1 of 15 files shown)


connor-parke-icf (approved):
lgtm; this all aligns with what we discussed. nice catch updating the reference materials downloads as well. One thing im not sure about is the distinction between "inline" and "attachment" disposition types. on the document preview page, if the document is unable to be rendered as a PDF it will render as regular download button. In those cases, wouldnt the download need to be "attachment" instead of "inline"? (It might not show as an issue since we may still be doing the blob download there, but if possible itd be great if all the file downloads were using the same pattern. 

It may be that it works fine, but im also curious what the difference is between attachment and inline. We had some issues with CSP and CORS on the deployed environments and had to make a couple of exceptions/rules for it (see thread in the main channel, related to https://github.com/Enterprise-CMCS/demos/pull/1535) which may involve the disposition type. When this is merged, we should verify that it all works fine on DEV since we had issues there which didnt appear on local.

nunesmlucas:
> lgtm; this all aligns with what we discussed. nice catch updating the reference materials downloads as well. One thing im not sure about is the distinction between "inline" and "attachment" disposition types. on the document preview page, if the document is unable to be rendered as a PDF it will render as regular download button. In those cases, wouldnt the download need to be "attachment" instead of "inline"? (It might not show as an issue since we may still be doing the blob download there, but if possible itd be great if all the file downloads were using the same pattern.
> 
> It may be that it works fine, but im also curious what the difference is between attachment and inline. We had some issues with CSP and CORS on the deployed environments and had to make a couple of exceptions/rules for it (see thread in the main channel, related to #1535) which may involve the disposition type. When this is merged, we should verify that it all works fine on DEV since we had issues there which didnt appear on local.

Good catch. The doc preview stays `inline` because the PDF `<embed>` needs it to render, and the non-PDF "Download File" button uses the blob download (with its own `download` attr), so disposition doesn't matter there. References/reports navigate straight to S3, so those use `attachment` to force the named download. So no change needed — but agreed, let's verify downloads + the PDF preview on DEV once this merges given the #1535 CSP/CORS stuff.

PR #1535, opened 2026-07-07
DEMOS-2403: Update CSP and embedder policy to allow presigned url embeds

cms-jesse:
This change updates `Cross-Origin-Embedder-Policy` to `unsafe-none` since S3 cannot set the CORP header and we are now directly embedding S3 presigned urls.

This also updates the CSP to allow the embeds.

An additional commit was added to fix the existing snyk vuln in aws-cdk-lib


dustbuster (approved):

connor-parke-icf (approved):
may or may not involve some of the changes with https://github.com/Enterprise-CMCS/demos/pull/1534, where we are specifying a download url as "inline" or "attachment"

PR #1536, opened 2026-07-07
DEMOS-2371-public-private-comments-for-deliverable-report

TARehman:
This turned out not to be quite as bad as I was fearing. However, it did require a gentle refactor of the code to support an optional column width parameter in order to override the default "auto-fit" behavior. That, in turn, led down a bit of a rabbit hole in TS generics, but I think the ultimate solution is okay. Columns are now configured with a columnName and an optional columnWidth; if the columnWidth is present, we override the default auto-fit behavior, otherwise, we don't touch it. We also explicitly set the row height to 12 - note that this is the behavior which results in a row height of 16 in the file, because I _believe_ the row height parameter is expressed in typographic points. A little confusing.

All the rest was just some SQL to concatenate everything we wanted together, then using `string_agg` to combine those with carriage returns in between them (I didn't try regular `\n` newlines because its Excel, so if we want to see if that works we can always do that). Everything "seems" to be working from the testing I could do, and I use `left` in the SQL to avoid anything more than 32000 character which I think we agreed was the size limit in Excel.


connor-parke-icf (approved):
this all seems to make sense to me. The bulk of the changes are due to replacing the column configuration name string with an object that can contain the name and an optional width param. The types are a little confusing, but as far as typical development on the reports (adding columns, making new ones, etc.) i dont think the type defs will need changing much. Beyond that the inserting of the two private and public comment column in the deliverable status report looks good. 

I imagine without the word wrap, line height configs, and column width, the column would grow much wider than the other columns, making it difficult to work with. On one hand, if we were using csv the formatting responsibility would be on the consumer; but, since we are using excel we have the ability to bake this into the report generation.  All looks good to me

TARehman:
Tagged as `DO NOT MERGE` for the moment because team is getting customer feedback next week.

PR #1541, opened 2026-07-08
Demos 2405 refactor layout

connor-parke-icf:
this is part one of maybe three PRs for updating the root of the client. Initially, the goal was to make the Header less context-aware, and able to be configured at the Route level. But as i was going through that, i found that we had parts of user authentication/authorization further down the stack than should be needed. For instance, in the LayoutProvider we handled redirecting based on withAuthenticationRequired, when this step could be taken care of long before determining a page's layout.  This PR attempts to adjust some of that before tackling the layout/routes themselves.


client/src/components/user/UserProvider.tsx
```diff
@@ -23,49 +25,20 @@ export const GET_CURRENT_USER_QUERY = gql`
...
-        email={auth.user?.profile.email}
-        errorMessage={errorParts.length > 0 ? errorParts.join("; ") : undefined}
-      />
-    );
+  if (!data?.currentUser) {
+    return <UserAuthenticationFailed errorMessage={NO_USER_FOUND_ERROR_MESSAGE} />;
   }
 
   // Provide the current user data to the rest of the app if auth succeeds
   return <Ctx.Provider value={{ currentUser: data.currentUser }}>{children}</Ctx.Provider>;
```
connor-parke-icf, line 43: reduced the scope of this component to handle just populating the context with current user data from gql as we can assume at this point that authentication has already occurred

client/src/router/DemosBootstrap.tsx
```diff
@@ -0,0 +1,20 @@
...
+  return (
+    <DemosAuthProvider>
+      <RequireAuthentication>
+        <DemosApolloProvider>
+          <UserProvider>
+            <DemosRouter />
+          </UserProvider>
+        </DemosApolloProvider>
+      </RequireAuthentication>
+    </DemosAuthProvider>
```
connor-parke-icf, line 18: goal with this was to separate the non-layout "bootstrapping" operations out of the main routing/layout. This includes authenticating the user, initializing Apollo, and creating the user context.

client/src/router/DemosRouter.tsx
```diff
@@ -34,60 +31,54 @@ const HomePage = () => {
...
-                  path="reports"
-                  element={
-                    <RequireRole allowedRoles={["demos-admin", "demos-cms-user"]}>
-                      <ReportsPage />
-                    </RequireRole>
-                  }
-                />
-                <Route path="admin" element={<AdminPage />} />
-                <Route path="references" element={<ReferencesPage />} />
+    <BrowserRouter>
```
connor-parke-icf, line 34: now the router only cares about its routes. This will be the focus of the next PR, where the routes themselves are adjusted to declare Layouts/headers/sidebar/authorization/debugmode more cleanly

client/src/router/RequireAuthentication.tsx
```diff
@@ -0,0 +1,31 @@
...
+const RenderChildrenWithAuth = withAuthenticationRequired(RenderChildren, {
+  OnRedirecting: () => <LoadingScreen />,
+  signinRedirectArgs: {
+    state: { returnUrl: window.location.pathname + window.location.search },
+  },
+});
+
+export const RequireAuthentication: React.FC<{ children: React.ReactNode }> = ({ children }) => {
+  const auth = useAuth();
+  const AuthenticatedChildren = shouldUseMocks() ? RenderChildren : RenderChildrenWithAuth;
```
connor-parke-icf, line 18: here is where we make the determination based on whether we're in mock mode whether to authenticate the user. This logic used to exist elsewhere, but now it has a dedicated component as an authorization gate, which is placed right after the AuthProvider

client/src/DemosApp.tsx
```diff
@@ -1,6 +1,6 @@
 import React from "react";
-import { DemosRouter } from "router/DemosRouter";
+import { DemosBootstrap } from "router/DemosBootstrap";
 
 export const DemosApp: React.FC = () => {
-  return <DemosRouter />;
+  return <DemosBootstrap />;
```
connor-parke-icf, line 5: instead of calling the router, we call the Bootstrap, which itself eventually calls the router. This way, we get all the things we need before thinking about how to direct the user

connor-parke-icf:
closing to break down task differently

PR #1542, opened 2026-07-08
DEMOS-2409: centralize logger code

cms-jesse:
## High-Level Changes
- Use proper service name for logs in deleteInfectedFile lambda
- Match logging format in BN lambda to other services
- Centralize logger in the shared_library
- Update shared library with exports to simplify import paths
- Update all code using shared library to use updated import paths
- Update esbuild configuration to alias the exports

## Overview
In the existing codebase, each service has its own log.ts file. For the most part, these are all supposed to be identical with the exception of the service name used to label the logs. At the time these files were added, having them duplicated was the simplest way to get the functionality added. Now that there is a shared library for other use cases, it makes sense to move the logger there as well and remove the duplicated code.

The actual "issues" fixed in this PR are that deleteInfectedFile was labeling itself in the logger as "fileprocess" and the BN lambda did not make use of the same log.ts file and had a different log format from everything else. If needed, a smaller PR with only those fixes could be made if the scope of changes is too large for this PR

## Details
Because of the way the logger works, the main change to the logger code is that there is now a wrapper function around the key components, which are all returned to the caller. This allowed for minimal changes to the application code while still providing the functionality.

This PR also includes changes to the shared library configuration. To improve (in my opinion) the usability of the shared library, the `exports` section was added to the package.json. This allows the packages to be imported like this:

```
import {createLambdaLogger} from "demos-shared-library/log"
```

Rather than:

```
import {createLambdaLogger} from "demos-shared-library/src/log/index"
```

I have also updated code that uses other imports to use the same update.

## Other Notes

- Where I could, `pino` and `pino-pretty` dependencies were removed since those are now inherited from the shared library. Some services still have it as a dependency where `pino` types are being used in the service code.


shared_library/package.json
```diff
@@ -5,8 +5,34 @@
...
+    },
+    "./BN/validation": {
+      "types": "./dist/src/BN/validation.d.ts",
+      "import": "./dist/src/BN/validation.js"
+    },
+    "./log": {
+      "types": "./dist/src/log/index.d.ts",
+      "import": "./dist/src/log/index.js"
+    }
+  },
```
cms-jesse, line 35: This is the added section that controls the names of the exports/imports. Not sure if `./BN/rulesets` would be better as `./BN/rulesets/v1` or if its ok to have it "default" to v1

(This exports object is also used in the CDK code to set all the proper aliases for esbuild)

lambdas/UIPath/log.ts
```diff
@@ -45,40 +46,28 @@ for (const logObject of REDACTED_LOG_OBJECTS) {
...
+  formatters: {
+    log: (obj) => {
+      // Extract top-level fields (eg. type to match default AWS lambda logs)
+      // All other fields go under `ctx`
+      const { type, ...rest } = obj;
+      // check `rest` length to prevent logging ctx:{}
+      return {
+        type,
+        ctx: Object.keys(rest).length ? serializeLogObject(rest) : undefined,
+      };
```
dustbuster, line 59: Will your logger have the redaction stuff? As this really tries very hard to share the tokens.

lambdas/UIPath/log.ts
```diff
@@ -45,40 +46,28 @@ for (const logObject of REDACTED_LOG_OBJECTS) {
...
-        };
-      },
-    },
-    // Disable pretty transport in bundled/worker contexts to avoid worker path issues.
-    transport: isNotOnAWS()
-    ? {
+  },
+  redact: {
+    paths: LOGGER_REDACTION_PATHS,
+    censor: REDACTED,
```
dustbuster, line 64: Nevermind!

dustbuster (approved):
This is fantastic! GET ALL THEM THINGS IN ONE PLACE!

lambdas/UIPath/log.test.ts
connor-parke-icf, line 1: what is this logger doing that is different from the other ones that warrants its own test file? could this one be removed?

pipelines/Jenkinsfile.cdk
```diff
@@ -95,6 +95,15 @@ pipeline {
...
               }
+              // dirCon('lambdas', 'node') {
+              //   sh '''
+              //     set -e
+              //     for dir in */; do
+              //       [ -d "$dir" ] || continue
+              //       (cd "$dir" && npm ci)
+              //     done
+              //   '''
+              // }
```
connor-parke-icf: is this intended to be here? i imagine it could be either developer artifacts or left in for ease of enable/disable

server/src/log.ts
```diff
@@ -1,51 +1,3 @@
...
-export const store = new Map();
-export const als = new AsyncLocalStorage<typeof store>();
-
-export const log = new Proxy(parentLogger, {
-  get(target, property, receiver) {
-    target = als.getStore()?.get("logger") ?? target;
-    return Reflect.get(target, property, receiver);
-  },
-});
+export const {log, reqIdChild, store, als } = createLambdaLogger("graphql")
```
connor-parke-icf, line 3: we have done some changes with the logging on server, but im not sure if this change and those actually interact with eachother. Worth looking at, but not a blocker

shared_library/package.json
connor-parke-icf: i think its fine, until we have a meaningful v2

connor-parke-icf (approved):
looks good to me. I believe i can follow along with the bulk of the changes. Most of the files touched are just moving around imports anyway. I like the removal of the different loggers in favor of a centralized one.  I do know we have some logging changes for server, but i dont know if that is going through a pathway that is impacted by the changes here.

lambdas/UIPath/log.test.ts
cms-jesse: The UI path logger builds on top of the basic logger to add some custom redaction. The tests are only testing the redaction functionality, but not duplicating the tests in the shared logger

pipelines/Jenkinsfile.cdk
cms-jesse: Oh, no I should remove this

PR #1544, opened 2026-07-09
Demos 2408 form validation messages

connor-parke-icf:
This PR introduces field validation messages to the document upload and edit modals. To do this, ive implemented a useValidation hook as a means of organizing field validations into a single struct such that both the overall form validation state and the messages displayed under each field are in sync; if an error message is displaying the button is disabled. It essentially moves the functions passed into the getValidationMessage into a mapping where they can be used for both displaying messages and determining button state. 

I applied this pattern to the Apply Demonstration Types dialog and the deliverable dialog as a proof of concept where it appeared to slot in without issue; but I excluded them from this PR to tighten the scope. 

https://github.com/user-attachments/assets/09542259-0490-4513-9201-dce6f591c5f2


client/src/components/dialog/document/DocumentDialog.tsx
```diff
@@ -333,6 +359,14 @@ const setDefaultDocumentType = (
   };
 };
 
+const formValidation: ValidationSchema<DocumentDialogFields> = {
+  file: [(formData) => (formData.file ? undefined : ERROR_MESSAGES.noFileSelected)],
+  name: [(formData) => (formData.name.trim() ? undefined : ERROR_MESSAGES.missingDocumentTitle)],
+  documentType: [
+    (formData) => (formData.documentType ? undefined : ERROR_MESSAGES.missingDocumentType),
+  ],
+};
```
connor-parke-icf: form validation is defined in this struct which maps a field name to a validation function. Though not utilized here, each field can take a multitude of validators, and the first failing validation will be returned as the message.

client/src/components/dialog/document/DocumentDialog.tsx
```diff
@@ -463,6 +495,7 @@ export const DocumentDialog: React.FC<DocumentDialogProps> = ({
           uploadStatus={uploadStatus}
           uploadProgress={uploadProgress}
           handleFileChange={handleFileChange}
+          validationMessage={errors.file}
```
connor-parke-icf: validationMessages are simply passed through as props.

client/src/components/dialog/document/DocumentDialog.tsx
```diff
@@ -448,7 +480,7 @@ export const DocumentDialog: React.FC<DocumentDialogProps> = ({
         actionButton={
           <UploadButton
             onClick={onUploadClick}
-            disabled={isMissing || !formHasChanges || isBNPreValidationBlocking}
+            disabled={!isValid || !formHasChanges || isBNPreValidationBlocking}
```
connor-parke-icf, line 493: isValid is true when there are no validation messages

client/src/components/input/Input.tsx
```diff
@@ -75,8 +76,11 @@ export const Input: React.FC<InputProps> = ({
         disabled={isDisabled ?? false}
         value={currentValue}
         onChange={handleChange}
+        onBlur={() => setTouched(true)}
```
connor-parke-icf, line 79: before this, we did not have any validation messages appearing when form elements were not filled (except for some checks for "has attempted submit" and an "isDirty" check i remember seeing). Without the touched check, the validation messages would appear on component render, before the user has had the chance to make changes. Now validation messages appear only after the user has interacted with the element.

client/src/hooks/useValidation.ts
```diff
@@ -0,0 +1,34 @@
...
+  formData: TFormData,
+  schema: ValidationSchema<TFormData>
+) => {
+  const errors: ValidationErrors<TFormData> = {};
+
+  for (const field of Object.keys(schema) as (keyof TFormData)[]) {
+    const rules = schema[field] ?? [];
+
+    for (const rule of rules) {
+      const message = rule(formData);
```
connor-parke-icf: intentionally, the validation takes in theentire formData object so that fields can reference data other than their own when determining validation. This is especially useful in the case of comparing dates or when validations do not apply when another field is set to specific values (in the case of deliverables)

client/src/hooks/useValidation.ts
```diff
@@ -0,0 +1,34 @@
...
+  [TField in keyof TFormData]?: ValidationRule<TFormData>[];
+};
+type ValidationErrors<TFormData> = {
+  [TField in keyof TFormData]?: ValidationMessage;
+};
+
+export const useValidation = <TFormData extends object>(
+  formData: TFormData,
+  schema: ValidationSchema<TFormData>
+) => {
```
tresdonicf: A generic that extends `object` is a bit loose of a contract to really bother with typing, I think if we wanted to move this direction then we might try to tighten up this contract a bit more to formalize what the shape of a `FormData` object looks like

client/src/components/input/Input.tsx
tresdonicf: Good idea here! One thing I think we don't want though is if this causes us to manually `fireEvent.blur()` in tests in order to test the validation messages as this can become quite a pitfall as why a message is not showing. I'm seeing this pattern in some tests after this change and it would be best to avoid it.

To this effect, how about we do this in the `handleChange` instead and that function will always set it to true, everything else could stay the same, what do you think?

client/src/components/dialog/document/DocumentDialog.tsx
```diff
@@ -394,14 +428,12 @@ export const DocumentDialog: React.FC<DocumentDialogProps> = ({
...
   const isBNPreValidationBlocking =
     !!file && (bnPreValidation.status === "validating" || bnPreValidation.status === "invalid");
 
   const onUploadClick = async () => {
     if (documentDialogState === "uploading") return;
-    if (isMissing) {
+    if (!isValid) {
       showError(ERROR_MESSAGES.missingField);
       return;
     }
```
tresdonicf, line 443: Should this even be clickable in this case? Seems like the disabled below kinda invalidates this block (no pun intended)

client/src/components/dialog/document/DocumentDialog.tsx
tresdonicf: I'm not a huge fan of this using the in-scope `formData` value since it makes the testing surface for a particular complex error message a bit more difficult.

Take this for instance versus if you have a function outside of the component like:
```
export const validationFunction = (param1, param2, param3) => {
  // Do some more complex validation that might throw many different messages
}
```

While you can implement the complexity in either case the first one is a bit easier to tease apart the logic in this function that is responsible for deciding which error message whereas here we would need to actually observe the error message under the input (which this component shouldn't be responsible for testing).

This can be addressed somewhat simply though by just moving these to their own functions outside of the component and still using them in this pattern.

```
// imagine some validation functions outside the component

const formValidation: ValidationSchema<DocumentDialogFields> = {
  file: [(formData) => isFileValid(formData)],
  name: [(formData) => isNameValid(formData)],
  documentType: [(formData) => isDocumentTypeValid(formData)],
};
```

The only problem is that then this seems like it stops being worth the squeeze much since we could just call something like `isFileValid(formData)` in the component itself and pass that into the `validationMessage` prop

client/src/components/dialog/document/DocumentDialog.tsx
```diff
@@ -68,17 +76,19 @@ const SUCCESS_MESSAGES = {
...
 };
 
-export const TitleInput: React.FC<{ value: string; onChange: (value: string) => void }> = ({
-  value,
-  onChange,
-}) => (
+export const TitleInput: React.FC<{
+  value: string;
+  onChange: (value: string) => void;
+  validationMessage?: string;
```
tresdonicf, line 79: I think this should call its own validation function based on passed-in props rather than receiving a message to show in order to make it stand on its own a bit more. If it needs to receive additional props to tease that out that's alright as well

tresdonicf (commented):
Some cool ideas here! I left some food for thought comments. Overall the changes I like the most are:

1. `touched` on the input
2. Passing just `validationMessage` instead of the function

I'm really not sure if the juice is worth the squeeze on the hook though, we already have what we need to group these if we wanted to but the testing surface stuff I mentioned means that we'd probably want to roll these into their own top-level functions anyhow

client/src/hooks/useValidation.ts
connor-parke-icf: thought some time about this, but i do think object here is appropriate, since we wont always know the exact shape of the data used in the form. This way, its more flexible and can be used to validate an object for validity, including those with nested fields or one constructed on-the-fly without requiring flattening. Updated the parameter names to reflect this intention

client/src/components/dialog/document/DocumentDialog.tsx
connor-parke-icf: yeah, this one is odd. Ideally, no it shouldnt be clickable and the user wouldnt ever be able to enter this function without the button being clickable.  But, we also include a state check for "uploading" just above this, which makes me a little nervous to completely rely on the button state.  I think to make this change we might want to do a more comprehensive look at this dialog as a whole, in a separate effort. For now, thinking ima error on the side of caution

client/src/components/dialog/document/DocumentDialog.tsx
connor-parke-icf: gave this some thought, and largely concluded that the form validation belongs in the scope of the form, not on the individual fields, since a) the entire form validation-state is used to determine the overall clickability of the submit button and b) form fields may rely on other fields for determining validation. For instance, if a field is required only when another field is filled, an individual field cannot determine whether its valid or not.

connor-parke-icf:
it wont let me reply directly under the comments, so ill reply here.
* https://github.com/Enterprise-CMCS/demos/pull/1544#discussion_r3559947114: at first, i had tried to pull this into the existing handleChange implementations, but found some issues with it. For one, it would flag before the user has concluded their edit of the field. For instance, if an validation were to validate against email regex, it would flag at the first character. Another issue is that if a user clicked on a field and moved on without any changes, they would not get any feedback to whether something was wrong; the submit button would be disabled but no field showing a reason why.

*  https://github.com/Enterprise-CMCS/demos/pull/1544#discussion_r3560021072: The main motivation behind these changes was to couple the overall validation of the form to the validation messages that appear underneath each field. Centralizing the form validation to the scope of the parent form accomplishes this and provides some additional benefits.  Some fields requires the context of other fields to determine their validity, so checking the entire form data as a whole is appropriate. Another boon is that all the form validation is located in one place (i did intentionally exclude any loading state behaviors, since that is a separate concern in my mind).  Another reason, and a motivation behind genericizing away from a "formData" naming, is that there are other elements that might effect form validation outside of form data. For instance if a demonstration is approved it has stricter validation rules. 

I did however refactor to pull the main validation configuration outside of the main components so they can be tested independently.  In addition, by moving towards a more generic validator, i removed the grouping of validation messages. Now it is defined by a collection of validation checks, each producing a single unique error message.  These can then be surfaced on the fields independently.  Furthermore, any other messages that might need to appear, such as within an eagerTooltip, can be surfaced in the same way.

client/src/hooks/useValidation.ts
tresdonicf, line 1: This code using lots of generics becomes difficult to understand what's expected based on the code alone, some comments would really help clear up how this hook is intended to be used.

client/src/components/input/Input.tsx
tresdonicf: This is still causing the same problems with needing to manually fire a `.blur` event on tests to see the validation message, bu we can avoid this by just setting `setTouched(true)` in the `handleChange` event, this will simplify what's needed in the tests dramatically!

Also it's a bit more correct because this would fire if you focus and then blur an input without adding anything to it (which we don't need to show the message for).

This is the only thing I think I would have to actually request changes for since it's an easy fix that means we don't need to inherit the `fireEvent.blur` pattern manually in the tests which is an anti-pattern

client/src/components/dialog/document/DocumentDialog.tsx
```diff
@@ -170,7 +187,10 @@ const DropTarget: React.FC<{
           ref={fileInputRef}
           className="hidden"
           accept={ACCEPTED_EXTENSIONS}
-          onChange={handleFileChange}
+          onChange={(e) => {
+            handleFileChange(e);
+            setTouched(true);
```
tresdonicf, line 192: Nice, yeah this is the place to do it instead of `onBlur`! Overall it's still weird that we would add that here at the same time we added the `useValidation` hook. If anything I think that code should be responsible for abstracting some of this away...

client/src/components/dialog/document/DocumentDialog.tsx
tresdonicf: This analysis is a bit too absolutist. We should ideally write the simplest code where it's needed and this also means that for fields that can show their own message (this for instance doesn't need anything else to determine which message to show) we should do that because we avoid "scope-pollution" of needing to hold the messages a scope higher than where they're used.

This still means we'll have the actual _check_ for if the field is valid at the form level but I don't think we need to make that so inherently tied to the message

tresdonicf (changes requested):
Overall I think this doesn't add much to the codebase. I think that the API of it is a bit too rigid and I think that we have everything in the features of the language to implement this in a bit more of a cleaner, easier to test way without the hook.

This is all to say I would still rather write up a few functions at the top of the file to do this validation and if anything I would compose them into a form-level `isValid` using something like:

```
// Assume the functions below
const formIsValid = Array.prototype.every([
fieldAIsValid(fieldA),
fieldBIsValid(fieldA, fieldB),
...])
```

Or something of this effect to simplify re-use of these functions (or can even do something like your thing of having them all return a message and checking that every one is equal to `""`.

The other thing that this undermines a bridge too far for me is testability for individual components. Now instead of testing the function directly in a test `formIsValid(...)` it increases the bootstrapping needed so that you have to grab the input, pretend to type in it, to then see the validation message (assuming it's not in a function already). But then if it is in a function then there wouldn't be any reason to use the hook as that's just extra code for no discernable gain.

Overall I won't stop you from adding this hook if you want but when I ask myself the golden question of code review "Does this make the codebase better?" I unfortunately have to say no for this one.

The only reason I did request changes here is that we should not use the `onBlur` event to determine the `touched` state but rather `onChange` if we're going to use that. This means we won't have to do `fireEvent.blur` in tests (which I just cleaned up in a recent PR to remove a bunch of these) and it's not semantically accurate since the message will show if they focus and then immediately blur with no input.

client/src/components/input/Input.tsx
connor-parke-icf: validated with HCD that the onBlur in indeed the intended behavior. In the event of clicking on an empty field and clicking away without making any changes, we _do_ want to show an error indicating that the field is required.  Triggering the action onChange does not accomplish this, and could lead to errors displaying on the page before the user has finished their field entry.

connor-parke-icf:
validated with Design team that the onBlur behavior is the behavior we're looking for here, not onChange

PR #1545, opened 2026-07-09
DEMOS-2378 Show the Document Title in the PDF preview

nunesmlucas:
## Problem

The PDF preview toolbar showed a UUID instead of the document's title. Chrome reads
the PDF's internal `/Title` metadata and falls back to the URL (our S3 key, a UUID)
when it's missing. `Content-Disposition` — added in DEMOS-2378 — only controls the
**download** file name, not what the viewer displays.

## Fix

Write the Document Title into the PDF's internal `/Title` metadata (`pdf-lib`), so
the viewer shows it. This overwrites any existing title, fixing both the missing-title
UUID case and PDFs that ship with a wrong internal title.

Applied in two places:

- **New uploads** — new `applyDocumentTitleMetadata` mutation, called once the virus
  scan clears (the first moment the file exists in the clean bucket).
- **Title edits** — `updateDocument` re-stamps the PDF, but only when `name` changes.

## Notes

- **PDFs only.** XLSX/DOCX store metadata differently and are left untouched.
- **`Content-Type` is preserved** on write-back — DEMOS-2378's download logic derives
  the file extension from it.
- **Best-effort.** Non-PDFs are skipped; unreadable PDFs and S3 errors are logged, not
  thrown. A cosmetic title never fails an upload or a rename.
- **User input is unchanged.** Titles are still stored exactly as typed and only
  sanitized when they become a download file name.
- **Not testable with `LOCAL_SIMPLE_UPLOAD=true`** (no real bytes stored locally) —
  use LocalStack or DEV.

## Dependencies

- `pdf-lib`

https://github.com/user-attachments/assets/59037b8f-e7c1-45d5-baa0-24451601384d

https://github.com/user-attachments/assets/066eaca5-4445-4785-95e1-25e1194fcef3


server/src/model/document/documentResolvers.ts
```diff
@@ -105,18 +106,32 @@ export const documentResolvers = {
...
           {
             name: input.name,
             description: input.description,
           },
           context.user
         );
+
+        if (input.name !== undefined) {
+          await applyPdfTitleMetadata(updatedDocument.s3Path, updatedDocument.name);
+        }
```
connor-parke-icf, line 120: here, we are letting anyone with access to the editDocument mutation edit any document arbitrarily (while a state user can access this mutation, they should not be able to edit a document on a demonstration they are not assigned to). we should move this to after the authorization gate in `/workspaces/demos/server/src/model/document/documentData.ts`

Also, any time we're needing to perform a s3 action and a database action in sync, we should wrap the operation in a transaction. That way, if the s3 operation fails for some reason, we can rollback the database operation.

server/src/model/document/documentResolvers.ts
```diff
@@ -105,18 +106,32 @@ export const documentResolvers = {
...
       } catch (error) {
         handlePrismaError(error);
       }
     },
+    applyDocumentTitleMetadata: async function applyDocumentTitleMetadata(
+      parent: unknown,
+      { documentId }: { documentId: string },
+      context: GraphQLContext
+    ): Promise<boolean> {
+      const document = await getDocument({ id: documentId }, context.user);
```
connor-parke-icf, line 132: same here; we need to have this operation behind the authorization gate so that users can only make modifications on the documents they are allowed to change

james-becwar-cms (dismissed):

connor-parke-icf (changes requested):
My main concerns are that with this we are opening up the API for any user with access to the mutation to make modifications to any document.  That could lead to a scenario where a state user modifies the title of a document on some other state's demonstration.  Also if we are modifying the s3 data in tandem with the database, we should be doing so in a transaction in case something fails and we need to rollback. 

Im not sure i completely understand the issue as well; if the user is uploading their own PDF, wouldnt it just have the title they set on it? i dont think we should be editing the metadata on a document that a user has uploaded. I feel like that could get hairy pretty quickly, since the document is no longer the same as when the user uploaded it, which could have legal implications. unless i am misunderstanding, and we're only editing storage metadata in s3

james-becwar-cms (commented):
Clearing my review.

PR #1549, opened 2026-07-10
Demos 2412 unselect table option after edit

connor-parke-icf:
For DEMOS-2412 we were seeing a weird inconsistent behavior when performing an edit action on a row in the table. On submit, the selected row would remain selected, but sometimes the pagination would reset the user back to page one. Because of this, to perform a subsequent edit the previous selection needed to be deselected first, which could mean fanning through pages to find the option. While the page resetting the pagination is an issue, we're opting to mitigate it by just forcing deselection upon saving an edit. The select column is used in these:
- DeliverableColumns: deliverable landing page, demonstration details page
- DocumentColumn: demonstration details->documents, 
- TypesColumns: demonstration details->Types, phase 8
- deliverableFileColumns: deliverable details page, cms and state files tabs.

For this, i implemented a "onSubmit" callback to the various edit dialogs across the app, feeding into it a call to resetRowSelection. Now, when submitting, the table resets to its unselected state. 

https://github.com/user-attachments/assets/3239635a-fa31-4135-ab82-775bf36f27f8

### DIFF
--- a/client/src/components/dialog/deliverable/EditDeliverableDialog.test.tsx
+++ b/client/src/components/dialog/deliverable/EditDeliverableDialog.test.tsx
@@ -62,6 +62,7 @@ const setup = (overrides?: {
   mocks?: React.ComponentProps<typeof TestProvider>["mocks"];
 }) => {
   const onClose = vi.fn();
+  const onSubmit = vi.fn();
   const deliverable = { ...TEST_DELIVERABLE, ...overrides?.deliverable };
 
   render(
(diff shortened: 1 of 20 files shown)


client/src/components/dialog/deliverable/EditDeliverableDialog.test.tsx
```diff
@@ -58,15 +57,11 @@ const MOCK_TAGS: Tag[] = [
   { tagName: "Annual Limits", approvalStatus: "Approved" },
 ];
 
-type OnSaveFn = (input: EditDeliverableInput, reasonForChange?: string) => Promise<void> | void;
-
```
connor-parke-icf, line 62: this onSave function callback wasnt being used anywhere. Verified that when removed, deliverable still saved as expected including dueDateChange request details

connor-parke-icf:
closing as moved branch to different ticket with different name. copied everything there

PR #1551, opened 2026-07-10
Demos 2416 unselect table option after edit

connor-parke-icf:
For DEMOS-2412 we were seeing a weird inconsistent behavior when performing an edit action on a row in the table. On submit, the selected row would remain selected, but sometimes the pagination would reset the user back to page one. Because of this, to perform a subsequent edit the previous selection needed to be deselected first, which could mean fanning through pages to find the option. While the page resetting the pagination is an issue, we're opting to mitigate it by just forcing deselection upon saving an edit. The select column is used in these:
- DeliverableColumns: deliverable landing page, demonstration details page
- DocumentColumn: demonstration details->documents, 
- TypesColumns: demonstration details->Types, phase 8
- deliverableFileColumns: deliverable details page, cms and state files tabs.

For this, i implemented a "onSubmit" callback to the various edit dialogs across the app, feeding into it a call to resetRowSelection. Now, when submitting, the table resets to its unselected state. 

https://github.com/user-attachments/assets/3239635a-fa31-4135-ab82-775bf36f27f8


tresdonicf (commented):
I get why it's not something we can solve at the table level but with all that the tanstack table offers I'm also kinda curious if there _might_ be some handle we're missing here that can help us with this. I guess as implemented now the issue will continue to crop up with new tables until this is added.

I'm gonna do some research here to see if there might be something else we can do to get this behavior from tanstack more directly!

tresdonicf (changes requested):
Okay we might be able to do this more simply! In our `Table` we have this block which we could probably extend just a bit to handle the case of "deleted" rows versus "edited" rows:

```
  // On data update, remove rows from selection if they no longer exist in the table
  React.useEffect(() => {
    setRowSelection((prev) => {
      const validRowIds = new Set(table.getRowModel().rows.map((r) => r.id));

      const filtered = Object.fromEntries(
        Object.entries(prev).filter(([key]) => validRowIds.has(key))
      );

      return filtered;
    });
  }, [data]);
```

We could for instance check if the _size_ of the new data array is different (indicating a deletion, or addition) or if the actual data is different! The challenge with this will be whether or not the `useEffect` will trigger here on a change in the actual data since it's an array, not sure if it'll do a deep inspection here. Regardless I think we should be able to fix this in the table itself! Maybe we just need a new prop or something to support it to compare the old data before change before the new data.

Could you give this a shot and see if it's possible?

connor-parke-icf:
i blew up git on this branch, so im just going to close this PR and open a new one where the suggested change has been implemented

PR #1554, opened 2026-07-10
Demos 2416 unselect table option after edit 2

connor-parke-icf:
Thanks to @tresdonicf for finding a massive simplification in how we can unselect the checkboxes on edit.  I blew up the git in my previous [PR](https://github.com/Enterprise-CMCS/demos/pull/1549) so i just recreated this one for it.

Instead of using a callback to trigger a cleaing of the checkboxes, which had to be routed to every relevent dialog, we can just modify the useEffect that we already had in Table.tsx.  Originally, that was scoped to just handling what should be selected after a delete operation, but it turns out after any change to the table data, we want to clear the selection. This vastly simplifies things, as we can just clear the selection whenever data changes. 

This did however unveil an elusive bug, where occasionally opening a dialog would clear the selection. This was happening on the Deliverables table on the demonstration details page because opening the modal would rerender the component which held the table. That component transformed the data before loading the table, so opening the dialog -> rerender tab component -> rebuild data -> unset row selections. Solution there was to memoize the data so its not rebuilt every render (we had a similar issue some time ago in one of the phases with respect to data not updating correctly on phase save).  

Anyway, this PR addresses both of those issues: checkbox behavior on save, and checkbox unintentionally getting unselected on open of the edit deliverable dialog. 

https://github.com/user-attachments/assets/524cc59b-056c-4578-8ba8-11049ad3634c

### DIFF
--- a/client/src/components/table/Table.tsx
+++ b/client/src/components/table/Table.tsx
@@ -190,15 +190,7 @@ export function Table<T extends { id: string }>({
 
   // On data update, remove rows from selection if they no longer exist in the table
   React.useEffect(() => {
-    setRowSelection((prev) => {
-      const validRowIds = new Set(table.getRowModel().rows.map((r) => r.id));
-
-      const filtered = Object.fromEntries(
-        Object.entries(prev).filter(([key]) => validRowIds.has(key))
-      );
-
-      return filtered;
-    });
+    setRowSelection({});
   }, [data]);
 
   return (
--- a/client/src/components/table/tables/DemonstrationDeliverableTable.tsx
+++ b/client/src/components/table/tables/DemonstrationDeliverableTable.tsx
@@ -47,12 +47,16 @@ export const DemonstrationDeliverableTable: React.FC<{
   );
   const actionButtons = viewMode === "demos-state-user" ? undefined : renderActionButtons;
 
-  const formattedDeliverables = sortDeliverablesByDefault(deliverables).map((deliverable) => ({
-    ...deliverable,
-    submissionDate: getLatestSubmissionDate(deliverable.deliverableActions),
-    combinedStatus: formatDeliverableStatus(deliverable),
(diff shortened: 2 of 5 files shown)


client/src/components/table/tables/DemonstrationDeliverableTable.tsx
```diff
@@ -47,12 +47,16 @@ export const DemonstrationDeliverableTable: React.FC<{
   );
   const actionButtons = viewMode === "demos-state-user" ? undefined : renderActionButtons;
 
-  const formattedDeliverables = sortDeliverablesByDefault(deliverables).map((deliverable) => ({
-    ...deliverable,
-    submissionDate: getLatestSubmissionDate(deliverable.deliverableActions),
-    combinedStatus: formatDeliverableStatus(deliverable),
-    combinedStatusFilter: formatDeliverableFilterStatus(deliverable),
-  }));
+  const formattedDeliverables = React.useMemo(
```
connor-parke-icf, line 50: turns out this was getting rerendered every time a dialog opened. this caused the data to get rebuilt, and triggered the useEffect.  We had see this manifest in an unsetting of selection on opening of the modal only sometimes, with inconclusive reasons. But, with the change to the useEffect, it happens everytime; so needed to memoize it so it only changes when it actually _was_ changed

client/src/components/table/Table.tsx
```diff
@@ -190,15 +190,7 @@ export function Table<T extends { id: string }>({
...
-    setRowSelection((prev) => {
-      const validRowIds = new Set(table.getRowModel().rows.map((r) => r.id));
-
-      const filtered = Object.fromEntries(
-        Object.entries(prev).filter(([key]) => validRowIds.has(key))
-      );
-
-      return filtered;
-    });
+    setRowSelection({});
```
connor-parke-icf, line 193: the core fix; just drop the selection every time data changes at the Table level

tresdonicf (approved):

PR #1556, opened 2026-07-13
Combine migration container

zoeelkins:
## Summary
Refactors migrate-local so the migration-owned database lives inside the DEMOS devcontainer's own Postgres instead of requiring a separate spin-up container. By default, the full pipeline now builds in a dedicated demos_migration database provisioned within the devcontainer, then ships only the demos_app data into the devcontainer's demos database. This removes the standalone scratch-Postgres dependency for local loads.

## Changes
- New default flow: provision a least-privilege scratch role + demos_migration DB inside the devcontainer Postgres (scripts/provision_scratch.sh), build there with SKIP_JSONSCHEMA, then dump/restore demos_app into the devcontainer.
- Added --external-scratch flag to opt back into building against a standalone scratch Postgres (--scratch-dsn / PG_URL).
- Env gains devcontainer_scratch_dsn() and scratch_pg_user/scratch_pg_password/scratch_pg_db settings (migration/lib.py).
- Updated Makefile, .env.example, environment reference doc, and the devcontainer-load runbook.
- Added tests for the new config resolution and plan (tests/test_lib.py, tests/test_migrate_local.py); refreshed uv.lock.

## Testing
Local make migrate-local verified through verify-prisma-local → provision-scratch → init → ddl. The run stops at load-full due to an environment MySQL PMDA connectivity issue (RDS handshake), unrelated to this change.


zoeelkins:
@TARehman @connor-parke-icf ~~marking this as a draft PR because I haven't fixed it so `make build` just writes to the devcontainer `demos_app` instead of the migration-owned `demos_app`.~~ Actually, it is smart to keep the migration PG db separate from the demos app db, even inside the local devcontainer, so we are not transforming the demos app db. To see the local dev commands & workflow, check out the `data-migration/runbooks/demos-devcontainer-load.md` file.

data-migration/pyproject.toml
```diff
@@ -14,6 +14,7 @@ dependencies = [
   "jsonschema>=4.21",
   "duckdb>=1.0",
   "boto3>=1.34",
+  "pre-commit>=4.6.0",
```
james-becwar-cms, line 17: Shouldn't matter, but I think pre-commit is install on the devcontainer by default, but this isn't hurting anything.

james-becwar-cms (approved):

PR #1557, opened 2026-07-13
task/duckdb-exporter-updates

TARehman:
This is an update to some of the existing data code in preparation for some of the work to be done on the data migration. This falls into basically three big tasks.

- Make the `devcontainer` a better location for Python-based development in particular. This involved setting up things like `ruff` and `mypy` to run dynamically to make the experience more similar to `TypeScript`, as well as futzing with test settings so that Test Explorer would work. This also was part of the next bullet.
  - Note that we just install the necessary packages during devcontainer build, rather than messing around with a virtual environment _within_ the container itself. I just moved the requirements file appropriately and added it to the devcontainer build for now.
  - I was sensitive to not wanting to nuke people's settings but I also wanted to enable some of the stuff that makes the development process easier, hence the copy-with-a-backup approach that I took.
- Reorganizing the code so that there is a single Python module in the project. VSCode was having many issues trying to cope with having many different Python modules in the project, and ultimately, it was simpler to just collapse them into one, `demos_data_tools`. Given that we are likely to have, at most, 10-12 Python files in that module, I think this is fine.
- Tweaking some small aspects of the DuckDB exporter. To avoid some namespace issues with Python, this was renamed `pmda_exporter.py`, but it basically works as it did before - there were some small tweaks to use the secret management features in DuckDB but otherwise it's more or less the same.
  - This still needs more attention; the tests are incomplete, and it needs additional work to productionize it, I think, but this was just a first pass.

I know the file count makes this look horrible, but fully 555 of those are just DDLs being moved from one place to another, and should hopefully not be an issue.


.devcontainer/Dockerfile
```diff
@@ -2,16 +2,4 @@ FROM mcr.microsoft.com/devcontainers/javascript-node:24-bookworm
 
 RUN apt-get update && \
   export DEBIAN_FRONTEND=noninteractive && \
-  apt-get -y install \
-  python3 \
-  python3-pip \
-  python-is-python3 \
-  awscli && \
-  pip3 install pre-commit --break-system-packages
```
james-becwar-cms, line 10: I think we need to install pre-commit so that the git checks work.

.devcontainer/post_create.sh
```diff
@@ -1,6 +1,11 @@
 #!/usr/bin/bash
+pip install --break-system-packages -r ./.devcontainer/python/requirements.txt
 pre-commit install
```
james-becwar-cms, line 3: Need to double check if pre-commit gets installed, I don't think its in the requirements.txt

james-becwar-cms (commented):
Just double checking about pre-commit

connor-parke-icf (approved):
nice; i was having some difficulty getting my python to run and had to do some workarounds to test things. Ill update what i have with these changes. Not too much to actually review here, like you said most of the changes are just moving to the target folder.

.devcontainer/Dockerfile
TARehman: https://github.com/Enterprise-CMCS/demos/blob/be01684d64b5602987de5b07f166e602a3cc33a9/.devcontainer/python/requirements.in#L4 It's here, but it'll be fixed more in my next PR anyway.

PR #1560, opened 2026-07-14
Demos 1900 - Emailer w/template engine 

dustbuster:
Introducing, EMAIL NOTIFICATIONS! 
Still a work in progress. 

But i just added a couple tables. (yes i know i need to update the MMD chart)

But I'll wanted to get a description up. It works! 

Deliverable created and Deliverable submitted. 

Screenshot 2026-07-30 at 11 23 50


.devcontainer/localstack/debug/runemailer.sh
```diff
@@ -0,0 +1,35 @@
...
+        "name": "LocalStack Deliverable",
+        "deliverableType": "Close Out Report",
+        "dueDate": "2026-07-10T12:00:00.000Z",
+        "status": "Upcoming"
+      }
+    }' >/dev/null
+
+echo "✅ Sent deliverable-created realtime email message to $QUEUE_NAME"
+echo "Tail logs with:"
+echo "aws --endpoint-url=$LOCALSTACK_ENDPOINT --region $AWS_REGION logs tail /aws/lambda/emailer --since 5m --follow"
```
dustbuster, line 72: Just a test script for local development

lambdas/emailer/emailTemplates/formatDate.ts
```diff
@@ -0,0 +1,9 @@
+export function formatDate(value: string): string {
+  const date = new Date(value);
+
+  if (Number.isNaN(date.getTime())) {
+    throw new Error(`Invalid email date value: ${value}`);
+  }
+
+  return date.toISOString().slice(0, 10);
```
dustbuster: Some of these can get combined into one file.

lambdas/emailer/emailTemplates/components/EmailLayout.tsx
```diff
@@ -0,0 +1,27 @@
...
+export function EmailLayout({ children }: { children: ReactNode }) {
+  return (
+    <Html lang="en">
+      <Head />
+      <Body style={bodyStyle}>
+        <Container style={containerStyle}>{children}</Container>
+      </Body>
+    </Html>
+  );
+}
```
dustbuster, line 27: Was able to really break this down to small chunks.

lambdas/emailer/emailTemplates/parts/createDeliverableEmailTemplate.tsx
```diff
@@ -0,0 +1,64 @@
+import { formatDate, getRequiredValue } from "../EmailHelper";
+import type { EmailRecipientGroups, EmailTemplateDefinition } from "../types";
+import { DeliverableEmail } from "./DeliverableEmail";
+import type { DeliverableEmailConfig, DeliverableEmailInput, DeliverableEmailProps } from "./types";
+
+// TO BE SET IN A BETTER WAY SOON
+const demosAppUrl = "http://localhost:3000";
```
dustbuster, line 7: I will come back to this.

server/src/model/email/emailSchema.ts
```diff
@@ -0,0 +1,23 @@
...
+export const emailSchema = gql`
+  input CreateEmailInput {
+    emailType: NonEmptyString!
+    entityType: NonEmptyString!
+    entityId: ID!
+    payload: JSONObject!
+  }
+
+  type Mutation {
+    createEmail(input: CreateEmailInput!): String!
```
connor-parke-icf: wondering if we should be exposing the email sending functionalility to the frontend. it might be better to handle this as a side-effect of other actions, rather than require the frontend to call it. Only usecase i imagine for this is for a test function in the admin, in which it may be good to call it createTestEmail, if its needed.

server/src/services/emailQueue.ts
```diff
@@ -0,0 +1,121 @@
...
+    },
+    triggeredAt: new Date().toISOString(),
+    idempotencyKey: `${input.emailType}:${input.entityType}:${input.entityId}`,
+    payload: input.payload,
+  };
+}
+
+export async function enqueueRealtimeEmail(message: RealtimeEmailEnvelope): Promise<string> {
+  const queueUrl = await resolveEmailQueueUrl();
+  log.info(
```
connor-parke-icf, line 109: i imagine we might want to create a db table to also hold the record of the email queue entries (when the database is not locked down, ofc)

server/src/services/emailQueue.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  cachedQueueUrl = response.QueueUrl;
+  return cachedQueueUrl;
+}
+
+export function buildRealtimeEmailEnvelope(input: {
+  emailType: string;
+  entityType: string;
+  entityId: string;
+  payload: RealtimeEmailEnvelope["payload"];
```
connor-parke-icf, line 73: we can probably use the type of the constants here so we arent using string, but rather filtering the expected values at the typescript level. If we are going to keep this as a resolver, we can add a Scalar to represent the allowed values, like how we've done it for Date Type for instance

server/src/services/emailQueue.ts
```diff
@@ -0,0 +1,121 @@
+import { GetQueueUrlCommand, SQSClient, SendMessageCommand } from "@aws-sdk/client-sqs";
+import { PRIMARY_AWS_REGION } from "../constants";
+import { log } from "../log";
+
+export const REALTIME_EMAIL_TYPES = [
```
connor-parke-icf: i dont think it matters here, but i suspect eventually we will want to define this as a table in the database. When we do that, we should move this to the consts.ts file.

connor-parke-icf (commented):
Only have reviewed the server changes so far, but theres a lot in here i think is great! i think theres a little tightening we could do with typescript, but i think the overall flow makes sense. One thing i think we should clarify is whether or not this should be accessible via the API or if it should be handled as side effects to existing operations. It being a resolver puts the responsibility of sending emails on the client/consumer, which may cause some issues (client forgets to call the send email function, client spams the API, etc). if we want to keep at least one endpoint for sending off a test email, we could probably rename the mutation to be "testSendEmail" or something, and make it only accessible to admin users.

server/src/model/email/emailSchema.ts
dustbuster: SO my thoughts for all these lambdas, is that you place a request to endpoint and it does the thing. That how it handled uipath, so i'm really trying to mimic that. 

So that all the 3rd party lambdas work the same way I really think this should be leveraged in the backend rather than the frontend. This way we are at least talk lam to lam and not client to lam. 

We also need to write to tables. Which is like next on the docket. I got another task i need to handle. but then i can come back to this.

server/src/services/emailQueue.ts
dustbuster: I have a plan.

dustbuster:
I generated a new branch from main. Closing this out. I wish there was a way to keep the comments!

PR #1561, opened 2026-07-14
Use vite built-in path resolution on `client`

tresdonicf:
Been getting this warning when running the client - now vite supports resolution of the `tsconfig.paths` variable itself!

Code_11_11_02


tresdonicf:
Starting over here

PR #1562, opened 2026-07-14
Upgrade to eslint v10 on `client`

tresdonicf:
This PR upgrades the client to `eslint v10`!

It accomplishes this by dropping the `eslint-plugin-react` package which keeps being annoying in regards to really slow update cycles (last publish was a year ago) and it's not providing much value given that we already use `typescript-eslint` which understands `.tsx` syntax.

We should look into including the react hooks linting rules at some point, these have some really good rules in regards to anti-patterns in hook behavior that can help to catch bugs: https://react.dev/reference/eslint-plugin-react-hooks

### DIFF
--- a/client/.snyk
+++ b/client/.snyk
@@ -2,11 +2,6 @@
 version: v1.25.0
 # ignores vulnerabilities until expiry date; change duration by modifying expiry date
 ignore:
-  SNYK-JS-BRACEEXPANSION-17706650:
-    - '*':
-        reason: No clean fix available in current ESLint React plugin stack
-        expires: 2026-07-14T00:00:00.000Z
-        created: 2026-06-30T14:21:33.964Z
   SNYK-JS-JSYAML-17900054:
     - '*':
         reason: No direct upgrade or patch available in current ESLint dependency chain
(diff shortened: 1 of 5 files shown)


client/src/router/DemosRouter.test.tsx
```diff
@@ -82,12 +82,9 @@ vi.mock("react-oidc-context", () => {
 });
 
 vi.mock("./DemosApolloProvider", async () => {
-  const React = (await import("react")).default;
```
tresdonicf, line 85: Unused code

client/.snyk
```diff
@@ -1,17 +1,7 @@
...
-  SNYK-JS-BRACEEXPANSION-17706650:
-    - '*':
-        reason: No clean fix available in current ESLint React plugin stack
-        expires: 2026-07-14T00:00:00.000Z
-        created: 2026-06-30T14:21:33.964Z
-  SNYK-JS-JSYAML-17900054:
-    - '*':
-        reason: No direct upgrade or patch available in current ESLint dependency chain
-        expires: 2026-07-23T00:00:00.000Z
-        created: 2026-07-09T00:00:00.000Z
```
tresdonicf, line 14: Can get rid of both ignores now!

connor-parke-icf (approved):

PR #1566, opened 2026-07-14
Migration document copy script

connor-parke-icf:
Implementation of migrating the documents from the existing s3 bucket holding the pmda documents to the proper s3 bucket for DEMOS. Also refactored some of the existing code to share helper functions like logging and establishing db connection through duckdb. 

Reviewed with Tom to go over python conventions and styles and some misc issues with the earlier passes of the code.


james-becwar-cms (approved):

data/demos_data_tools/duckdb_connection_manager.py
```diff
@@ -47,20 +45,14 @@ def load_db_configs_from_env() -> dict:
...
-        config (dict | None): A configuration for the two databases being connected.
-            When omitted, the configuration is loaded from environment variables.
-
     Returns:
         "DuckConn": A configured DuckDB connection.
     """
     logger.info("Creating DuckDB database")
-
-    if config is None:
-        config = load_db_configs_from_env()
```
connor-parke-icf, line 63: nice; yeah not really a point in passing in a config. no need for it being optional

data/demos_data_tools/duckdb_connection_manager.py
```diff
@@ -84,7 +76,7 @@ def create_duckdb_conn(config: dict | None = None) -> "DuckConn":
 
     conn.execute(f"ATTACH 'sslmode={ddb_demos_config['sslmode']}' AS {DEMOS_DDB_ATTACH_NAME} (TYPE postgres);")
     conn.execute("SET pg_null_byte_replacement=''")  # This is necessary to handle nulls from MySQL
-    logger.info("Attached DEMOS PostgreSQL database AS %s", DEMOS_DDB_ATTACH_NAME)
+    logger.info(f"Attached DEMOS PostgreSQL database AS {DEMOS_DDB_ATTACH_NAME}")
```
connor-parke-icf, line 79: nice, i prefer the template strings to the string substitution method. reads much better

data/demos_data_tools/history_trigger_generator.py
```diff
@@ -240,6 +240,7 @@ def main() -> None:
     with open("../../server/src/sql/history_triggers.sql", "w") as query_file:
         query_file.write(DROP_QUERY)
         query_file.writelines(queries)
+    return None
```
connor-parke-icf, line 243: i imagine this started flagging once we had the linter in place? explicit returns, even if None?

data/demos_data_tools/logger_utils.py
```diff
@@ -0,0 +1,51 @@
+"""Shared logging helpers for the data tools scripts."""
+
+import logging
+from datetime import datetime
+from typing import TYPE_CHECKING
+
+# Note these imports are done in this fashion to make the mocking easier in tests
+from logging import FileHandler, Formatter, StreamHandler, getLogger
```
connor-parke-icf, line 8: i take it, by importing just the specific elements of the library, we can mock just those elements instead of mocking the entire logging package?

data/demos_data_tools/logger_utils.py
```diff
@@ -0,0 +1,51 @@
...
+        name (str): The logger name to initialize.
+        verbose (bool): Whether to enable debug-level logging.
+
+    Returns:
+        Logger: The configured logger instance.
+    """
+    logger = getLogger(name)
+
+    # We use this sentinel to avoid reconfiguring a logger that's already set up
+    if getattr(logger, _CONFIGURED_FLAG, False):
```
connor-parke-icf, line 31: if i understand correctly, there is a core singleton logger.  we fetch it, apply some attributes to it, including an indicator that it has been configured to this file's standard. Then we can just grab it whenever we need it

data/demos_data_tools/migrate_files.py
```diff
@@ -40,46 +49,17 @@ class CopyRow(TypedDict):
...
+    Returns:
+        List[CopyRow]: A list of the rows to copy.
     """
     rows = connection.execute(SELECT_UNMIGRATED_FILES_QUERY).fetchall()
-    for row in rows:
-        yield {
-            "old_path": row[0],
-            "new_path": row[1],
-        }
+    return [{"old_path": row[0], "new_path": row[1]} for row in rows]
```
connor-parke-icf, line 62: i see; yeah theres no need to treat this as an iterator since the list fetching is unfront anyway, we can just return the list.

data/demos_data_tools/migrate_files.py
```diff
@@ -91,12 +71,13 @@ def mark_row_copied(connection: "DuckConn", row: CopyRow) -> None:
...
         MARK_FILE_MIGRATED_QUERY,
-        [row["old_path"], row["new_path"]],
+        {"old_path": row["old_path"], "new_path": row["new_path"]},
     )
+    return None
 
 
 def copy_s3_object(
-    s3_client: Any,
+    s3_client: "S3Client",
```
connor-parke-icf, line 80: i still dont think i quite understand the quoting distinction here; why is it typed as a "S3Client" instead of just S3Client?

data/demos_data_tools/migrate_files.py
```diff
@@ -142,29 +124,25 @@ def migrate_file(
...
+    db_connection = create_duckdb_conn()
+    s3_client = boto3.Session().client("s3")
+    copied_count = 0
+    unmigrated_files = get_unmigrated_files(db_connection)
 
-    Returns:
-        None: This function does not return a value.
-    """
-    load_dotenv(ENV_FILE)
+    while copied_count < len(unmigrated_files):
```
connor-parke-icf, line 141: makes sense to use a while loop, since the index isnt important to us, just the overall count

data/demos_data_tools/test_duckdb_connection_manager.py
```diff
@@ -76,38 +83,39 @@ def test_load_db_configs_from_env(self, mocker):
...
 
-    def test_create_duckdb_conn_01(self, mock_duckdb_connect):
+    def test_create_duckdb_conn_01(self, mock_duckdb_connect, mock_load_credentials):
         """Test duckdb_connection_manager.py functions.
 
         ::create_duckdb_conn
 
         ::It should invoke an in-memory database.
         """
-        duckdb_connection_manager.create_duckdb_conn(self.db_conn_config)
```
connor-parke-icf, line 86: i ahh believe i see; instead of loading the connection config at specific places throughout the test, we can put that behavior into a fixture to be called at the start of every test. I imagine the same set of creds is used at each test, or at least can be overridden at a specific test

data/demos_data_tools/test_logger_utils.py
```diff
@@ -0,0 +1,79 @@
...
+import logger_utils
+
+
+class TestLogger:
+    """A class for the tests for the logger.py file."""
+
+    @pytest.fixture
+    def logger_mocks(self, mocker):
+        """Isolate get_logger from the logging module and the filesystem."""
+        mock_logger = mocker.MagicMock()
```
connor-parke-icf, line 16: MagicMock seems odd, but i think thats just because of the name. from what i understand, you use it when you have a more complex object that needs to be mocked? like, the core singleton logger?

PR #1579, opened 2026-07-15
DEMOS-2453: updated regex to escape the input in keyword search

dustbuster:
Basically we were not escaping the input search values. Simple fix. 

https://github.com/user-attachments/assets/d9874090-5ad3-45b3-9059-686c5e8f023d

Screenshot 2026-07-15 at 12 47 56


client/src/components/table/KeywordSearch.tsx
```diff
@@ -51,12 +53,13 @@ export function highlightCell<TData>({
...
 
-  const pattern = validKeywords.map((k) => `(${k})`).join("|");
+  const pattern = `(${validKeywords.map(escapeRegExp).join("|")})`;
   const regex = new RegExp(pattern, "gi");
+
   const parts = text.split(regex);
 
   return parts.map((part, index) =>
-    regex.test(part) ? (
+    index % 2 === 1 ? (
```
dustbuster, line 60: The previous regex.test(part) was unreliable because the regex has the global g flag. Global regexes mutate their lastIndex each time .test() runs, potentially causing alternating or missed highlights. Checking the split position avoids that stateful behavior.

client/src/components/table/KeywordSearch.tsx
```diff
@@ -51,12 +53,13 @@ export function highlightCell<TData>({
 
   if (!validKeywords.length) return text;
 
-  const pattern = validKeywords.map((k) => `(${k})`).join("|");
+  const pattern = `(${validKeywords.map(escapeRegExp).join("|")})`;
```
dustbuster: Further escaping!

client/src/components/table/KeywordSearch.tsx
```diff
@@ -19,6 +19,8 @@ export interface KeywordSearchProps<T> {
   placeholder?: string;
 }
 
+const escapeRegExp = (value: string) => value.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
```
james-becwar-cms: Mind if swap out this with https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/escape ?

james-becwar-cms (changes requested):

james-becwar-cms (approved):

client/types/shims.d.ts
```diff
@@ -1,7 +1,15 @@
 declare const __GIT_COMMIT__: string;
 
+interface RegExpConstructor {
+  escape(str: string): string;
+}
+
```
dustbuster, line 6: This makes TS happy, AND let's use use `RegEx.escape`

PR #1589, opened 2026-07-17
DEMOS-2446: person_state: added person_state migration

connor-parke-icf:
migrates state assignment from user_authrzd_state_acs to person_state. Rejects states that are not valid: 'XX' and empty. Only state users are assigned, as other users are assigned to all states.


data/migration/stage_pmda_for_migration/models/cleaned/cleaned_demos_app_person_state.sql
```diff
@@ -0,0 +1,13 @@
...
+    au.geo_ansi_state_cd AS state_id
+FROM
+    {{ source('legacy_pmda_raw', 'user_authrzd_state_acs') }} AS au
+INNER JOIN
+    {{ ref('users_cleaned_demos_app_person_state_user') }} AS cp
+    ON
+        au.user_id = cp._legacy_id
+WHERE
+    au.geo_ansi_state_cd NOT IN (
+        SELECT e1.geo_ansi_state_cd FROM {{ ref('errors_pmda_person_states_with_invalid_states') }} AS e1
```
connor-parke-icf, line 12: followed convention of the cleaned_demos_app_person to filter the invalid states based on what appears in the errors.  Not sure if theres a better way to do this (maybe reference both person_id and state_cd, since its a join table?)

TARehman: I think this is right way for now, and we can revisit if we want.

TARehman (approved):
LGTM. Only thing I'd say is maybe add a note about the _column convention we discussed that you're using for _legacy_id in the README.

connor-parke-icf:
added note in readme to followup PR

PR #1590, opened 2026-07-17
Demos 2446 default system role assignment

connor-parke-icf:
add default role assignments for migrated users according to the convention used in the gql context building. implemented as a cleaned table with a csv mapping, similar to the crosswalk.


TARehman (approved):
LGTM

PR #1593, opened 2026-07-20
DEMOS 2427 Prevent Deletion of Linked Demo Types

JeffreyM-GA:


client/src/components/table/tables/TypesTable.test.tsx
```diff
@@ -169,6 +179,18 @@ describe("TypesTable", () => {
       const deleteButton = screen.getByTestId("delete-type");
       expect(deleteButton).not.toBeDisabled();
     });
+
+    it("disables delete when a selected demonstration type is linked to a deliverable", async () => {
```
connor-parke-icf, line 183: might can include a check for the tooltip text in this test, but not strictly needed. If you do, i would put the text into a const and import it so we dont have to update the tests if we want to change the text later

connor-parke-icf (approved):
LGTM!

PR #1595, opened 2026-07-20
DEMOS-2464: added demo type string match on reference table

dustbuster:
This allows search and highlight text for demo type for reference tables.

Added spinners to the dialog button and the table button!

It's really fast since it's local and inherently quick. 
https://github.com/user-attachments/assets/baa7c562-cebf-442e-820c-b834767c0017

Screenshot 2026-07-20 at 11 42 06


james-becwar-cms (approved):

client/src/components/table/columns/ReferencesColumns.tsx
```diff
@@ -50,6 +52,8 @@ export function ReferencesColumns() {
       id: "actions",
       header: () => <span className="sr-only">Actions</span>,
       cell: ({ row }) => {
+        const isDownloading = downloadingReferences.has(row.original.id);
```
connor-parke-icf: we might can simplify this by keeping the onDownload behavior within the column. that way we dont need to keep a Set of the currently in-progress downloads in the parent and pass down the onDownload, and instead each button keeps track of whether itself is currently in progress. In the same way as you have the ReferenceAgreementDialog, the dialog keeps track of its own downloading state, not its parent

connor-parke-icf (approved):
one suggestion for possible simplification, but not blocking, lgtm

client/src/components/table/columns/ReferencesColumns.tsx
dustbuster: I will take a look at that.

dustbuster: done.

PR #1596, opened 2026-07-20
(1/2) DEMOS-1936: Add Support for Eager Tooltips to Buttons

tresdonicf:
Per DEMOS-1936, this PR adds the ability to add an eager tooltip to any kind of button. It does this using relatively new technologies of the anchor positioning + popover API in order to keep the implementation clean and minimal.

https://github.com/user-attachments/assets/c5875ad3-71b0-4472-abdd-157e5ef178bc

### DIFF
--- a/client/src/components/button/BaseButton.tsx
+++ b/client/src/components/button/BaseButton.tsx
@@ -22,6 +22,8 @@ disabled:text-text-placeholder
 disabled:cursor-not-allowed
 `;
 
+const NON_ALPHANUMERIC = /[^a-zA-Z0-9]/g;
+
 const getSizeClasses = (isCircle: boolean, buttonSize: ButtonSize) => {
   if (isCircle) {
     return {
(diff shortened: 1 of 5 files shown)


client/src/components/button/BaseButton.tsx
```diff
@@ -66,23 +70,38 @@ export const BaseButton: React.FC<ButtonProps> = ({
...
+        data-testid={name}
+        aria-label={ariaLabel || name}
+        aria-describedby={eagerTooltip ? tooltipId : undefined}
+        type={type}
+        onClick={onClick}
+        {...(form ? { form } : {})}
+        className={`${BASE_BUTTON_STYLES} ${sizeClasses} ${circleClasses} ${className}${eagerTooltip ? " [anchor-name:var(--anchor)]" : ""}`}
+        disabled={disabled}
+        {...(tooltip ? { title: tooltip } : {})}
+        {...(eagerTooltip ? { style: { "--anchor": anchorName } as React.CSSProperties } : {})}
```
tresdonicf, line 96: This might be the only thing about the implementation I'm not a massive fan of, right now tailwind doesn't have an API for supporting the anchor positioning API and thus we have to go back to `style` for this particular case.

client/src/components/tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,59 @@
...
+  }, [anchorRef]);
+
+  return (
+    <div
+      ref={popoverRef}
+      id={id}
+      role="tooltip"
+      popover="manual"
+      className={TOOLTIP_STYLES}
+      style={{ "--anchor": anchorName } as React.CSSProperties}
```
tresdonicf, line 54: And then same thing needed here to match

client/src/components/button/BaseButton.tsx
tresdonicf: Overall it's better than what would be needed to do it in JS though with `getBoundingClientRect()` and all of the math that needs to go on after that 🤮

james-becwar-cms (approved):

PR #1597, opened 2026-07-20
DEMOS-2445-basic-demo-adding

TARehman:
This is partial progress toward migration demonstrations. It is not complete, but provides a foundation that will help @connor-parke-icf work on some of his parts. Summary of changes below.

- Added two active filters, one for pending demonstration applications and one for finalized ones.
- Added validation logic that attempts to extract valid ID numbers from the in-progress / pending and the finalized demonstrations.
- Stack the two together with the fields that are reasonably derivable / crosswalked so that we have a starting point for the final cleaned table. Note that I left a placeholder for cleaned, because I'd already put the columns into the YAML file. We'll need to flip that one to be enforced when the cleaned table is ready.

Some other miscellaneous changes.
- Added `partial_parse` as `false` so when parsing, we'll get warnings if we have YAML models that don't have SQL.
- The linter ran on the code in `functions.sql` and found a small capitalization issue, so it just got fixed.
- I changed the sort order of what was in `models.yml` which explains some of the changes.
- Added a `gitignore` for `__pycache__` files to `data-migration`.


data/migration/stage_pmda_for_migration/models/errors/errors_invalid_demo_nums_in_finalized_pmda_demos.sql
```diff
@@ -0,0 +1,8 @@
+SELECT *
+FROM
+    {{ ref('apps_active_finalized_pmda_demos_mdcd_num_validations') }}
+WHERE
+    cleaned_mdcd_demo_num_format_valid = 'Invalid'
+    OR cleaned_mdcd_demo_num_location_valid = 'Invalid'
+    OR cleaned_mdcd_scndry_demo_num_format_valid = 'Invalid'
+    OR cleaned_mdcd_scndry_demo_num_location_valid = 'Invalid'
```
connor-parke-icf, line 8: i kinda like this strategy for validation, where we add a flag as metadata to the row, essentially. Keeps the validation outside of a global space and tied to the thing thats getting validated.  I imagine then, we still use the same where filter in the cleaned tabe to filter out the errors?

data/migration/stage_pmda_for_migration/models/models.yml
```diff
@@ -181,3 +164,109 @@ models:
         data_type: timestamptz
       - name: updated_at
         data_type: timestamptz
+
+  - name: cleaned_demos_app_demonstration
```
connor-parke-icf, line 168: i imagine we will need a similar table for application while we're at it, even though we dont have amendments and extensions yet

james-becwar-cms (approved):

connor-parke-icf (approved):

PR #1598, opened 2026-07-20
DEMOS-1876: Display User Management functionality on Admin

nunesmlucas:
## User Management — a couple of notes

Two spots in `UserManagement.tsx` look slightly odd on purpose. Leaving a note so nobody
"fixes" them later and quietly breaks a story requirement.

## Default sort is applied to the data, not the table's sort state

The AC wants unassigned records ("-") at the top, then Name alphabetically.

The obvious way to do that is `initialState.sorting`, and that was the first attempt — but
"Assigned to State(s)" isn't a sortable column, and TanStack skips columns marked
`enableSorting: false` when it builds the sorted row model. So the "-" rows never moved.
It also meant the first click on the Name header sorted *descending*, since that column was
already sitting at ascending.

So the default order is applied to the array itself (`sortUnassignedFirstThenByName`) and the
table starts with empty sort state. Clicking Name / IDM User Role / Last Login then behaves
like every other table in the app.

## Last Login sorts on a custom function

The cell accessor returns the formatted `MM/DD/YYYY` string, so keyword search matches what's
actually on screen. That string sorts wrong alphabetically though (`04/03/2026` lands before
`11/02/2025`), so `sortByLastLoginDate` compares the underlying timestamps instead.

Same trade-off applies if anyone adds another date column here.

https://github.com/user-attachments/assets/46deb75f-274c-4fa0-a64f-8639e3102aa3


client/src/pages/admin/UserManagement.tsx
```diff
@@ -1,7 +1,137 @@
...
 
+const UNASSIGNED = "-";
+const ALL_STATES = "All States";
+const SEARCH_STORAGE_KEY = "user-management-keyword-search";
+
+const IDM_USER_ROLE_LABELS: Record<PersonType, string> = {
+  "demos-admin": "Admin User",
+  "demos-cms-user": "CMS User",
+  "demos-state-user": "State User",
+  "non-user-contact": UNASSIGNED,
```
connor-parke-icf: we probably dont need to include this, and define it as `Record<UserType, string>`, since no users should have the non-user-contact person-type

client/src/pages/admin/UserManagement.tsx
```diff
@@ -1,7 +1,137 @@
...
+import { createColumnHelper, SortingFn } from "@tanstack/react-table";
+import { Person, PersonType, State, User } from "demos-server";
+import { Table } from "components/table/Table";
+import { highlightCell, KeywordSearch } from "components/table/KeywordSearch";
+import { PaginationControls } from "components/table/PaginationControls";
+import { formatDateForDisplay } from "util/formatDate";
 
 export const USER_MANAGEMENT_TEST_ID = "user-management";
 
+const UNASSIGNED = "-";
```
connor-parke-icf: maaaybe dont need this as a const but isnt hurting i guess. I would suggest, if we want this a const we should also export it and replace the the tests that check for "-" with it. but, since its one character... it may just be better to inline "-"

client/src/pages/admin/UserManagement.tsx
```diff
@@ -1,7 +1,137 @@
...
+    return ALL_STATES;
+  }
+  const stateNames = user.person.states
+    .map((state) => state.name)
+    .sort((a, b) => a.localeCompare(b));
+  return stateNames.length > 0 ? stateNames.join(", ") : UNASSIGNED;
+};
+
+const getLastLogin = (user: ManagedUser): string =>
+  user.lastLogin ? formatDateForDisplay(user.lastLogin) : UNASSIGNED;
```
connor-parke-icf: another argument for inlining "-": "unassigned" might be semantically a little misleading here, since last login isnt something to be assigned to a user

connor-parke-icf (approved):
LGTM! couple of small suggestions. Overall, i think we could afford to move the table and column definitions into the components/table and components/table/columns folder to isolate the concerns of each component. I also think that we might can simplify some of the sorting logic somehow, but not exactly sure how. Checking against the string "-" to determine whether a row is all-states or not feels a little flimsy to me, we might can just check the person-type in the column. Either way, imo none of this is blocking!

connor-parke-icf (approved):
nice looks great! now the page only cares about rendering the table, the "unassigned" logic seems more closely tied to the count of the assigned states instead of the string result. nice work

PR #1601, opened 2026-07-21
DEMOS-2445-add-trigger-control-to-last-migration-step

TARehman:
Point the pre-commit doc linter to the right pyproject.toml file. Update the loader. Add tests to the loader.

### DIFF
--- a/data/demos_data_tools/load_staged_data_to_demos_app.py
+++ b/data/demos_data_tools/load_staged_data_to_demos_app.py
@@ -2,8 +2,9 @@
 
 import argparse
 import os
+from dataclasses import dataclass
 from logging import getLogger
-from typing import TYPE_CHECKING, List, Tuple, TypedDict
+from typing import TYPE_CHECKING, List, Literal, Set, Tuple, assert_never
 
 from dotenv import load_dotenv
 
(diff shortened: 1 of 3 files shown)


james-becwar-cms (approved):

data/demos_data_tools/load_staged_data_to_demos_app.py
```diff
@@ -20,31 +21,69 @@
...
-    },
-    {
-        "staged_table": "cleaned_demos_app_users",
-        "target_table": "users",
-        "column_list": [
+type MigrationConfiguration = Tuple[TableInsertActionConfiguration | TriggerActionConfiguration, ...]
+type GeneratedSql = GeneratedInsertActionSql | GeneratedTriggerActionSql
+type MigrationSql = List[GeneratedSql]
+
+MIGRATION_CONFIGURATION: MigrationConfiguration = (
```
connor-parke-icf, line 73: i take it, whenever we update this, we will need to disable all the triggers on it that might be disruptive before we register the insert action with it, reenabling after? is this all triggers in general or is there a subset of triggers that we will need to keep

connor-parke-icf (approved):
all appears to check out to me. If i understand this, the actual usage of it is fairly simple, still revolves around the main configuration in which we need to disable/enable triggers along with the main insert

data/demos_data_tools/load_staged_data_to_demos_app.py
TARehman: I just put those ones in there to have examples and because we don't need the redundant checks while doing inserts for this so they're a good example. We can use this to disable / enable whatever triggers we like.

PR #1603, opened 2026-07-21
DEMOS-1936: Add Usages of EagerTooltip

tresdonicf:
Per DEMOS-1936: Adds usages of the `eagerTooltips` to the phase buttons

https://github.com/user-attachments/assets/7a62c56c-57e6-4c0d-9f6e-f186cd70c8b5


connor-parke-icf (approved):
This all looks great to me, no real comments, nice work. This gives a good example of a validation message that may not fit well into the pattern i was proposing in one of my earlier PRs, so would be a good example to work with for any future edits to it

PR #1606, opened 2026-07-21
Fix Client Tests on `main`

tresdonicf:
Two issues coming up with the client tests
1. The `RegExp.escape` usage that was just added is a bit new and not available in all envs (JSDOM) being one of them
2. A date that was set in the "future" has now become the past

### DIFF
--- a/client/src/components/dialog/deliverable/EditDeliverableDialog.test.tsx
+++ b/client/src/components/dialog/deliverable/EditDeliverableDialog.test.tsx
@@ -53,6 +53,8 @@ const TEST_DELIVERABLE: EditDeliverableDialogDeliverable = {
   demonstrationTypes: [],
 };
 
+const FUTURE_DUE_DATE = "2222-06-30";
+
 const MOCK_TAGS: Tag[] = [
   { tagName: "Aggregate Cap", approvalStatus: "Approved" },
   { tagName: "Annual Limits", approvalStatus: "Approved" },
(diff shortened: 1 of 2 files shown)


client/src/components/dialog/deliverable/EditDeliverableDialog.test.tsx
```diff
@@ -113,7 +115,7 @@ describe("EditDeliverableDialog", () => {
   it("shows the Reason for Change field when due date is modified", () => {
     setup();
     fireEvent.change(screen.getByTestId(SINGLE_DELIVERABLE_DUE_DATE_NAME), {
-      target: { value: "2026-07-20" },
```
tresdonicf, line 116: This is no longer in the future...

client/vitest.setup.ts
```diff
@@ -1,5 +1,11 @@
 // Import DOM testing library for Jest, to be used by all test files
 import "@testing-library/jest-dom";
+
+// Polyfill RegExp.escape (TC39 proposal, not yet available in jsdom)
+if (!("escape" in RegExp)) {
+  (RegExp as unknown as Record<string, unknown>).escape = (s: string) =>
+    s.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
+}
```
tresdonicf, line 8: If the RegExp package in our runtime doesn't have the `.escape` then we need to polyfill otherwise we get `RegExp.escape is not a function`

connor-parke-icf (approved):
looks good to me; i dont know much about the polyfill change, but i suspect its not very often this comes up.

tresdonicf:
> looks good to me; i dont know much about the polyfill change, but i suspect its not very often this comes up.

Yeah it's coming up in my env but it only applies for tests so eventually `jsdom` should have support for this but I know we were talking about maybe switching anyway to `happy-dom` or something

PR #1608, opened 2026-07-21
DEMOS-2468: added ability to arrow ^ & v w/keybrd

dustbuster:
i found that no dropdowns had arrow control. It would move the selector but DD would not scroll to non-visible items

https://github.com/user-attachments/assets/078e286d-a6ff-4849-9678-2c9edd774aad

This was borked everywhere apparently!
https://github.com/user-attachments/assets/3e2e0e43-7615-49a2-adfe-6f97da96eea8

### DIFF
--- a/client/vitest.setup.ts
+++ b/client/vitest.setup.ts
@@ -12,6 +12,12 @@ import React from "react";
 
 import { vi } from "vitest";
 
+// Element.scrollIntoView is not implemented by jsdom
+Object.defineProperty(HTMLElement.prototype, "scrollIntoView", {
+  configurable: true,
+  value: vi.fn(),
+});
+
 // Mock HTML dialog element for tests
 // HTMLDialogElement API is not fully supported in jsdom yet, so we need these mocks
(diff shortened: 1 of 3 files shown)


client/vitest.setup.ts
```diff
@@ -12,6 +12,12 @@ import React from "react";
 
 import { vi } from "vitest";
 
+// Element.scrollIntoView is not implemented by jsdom
+Object.defineProperty(HTMLElement.prototype, "scrollIntoView", {
+  configurable: true,
+  value: vi.fn(),
+});
+
```
dustbuster, line 20: Got 23 tests failing and putting this here fixed them all. 

```log
TypeError: activeOptionRef.current?.scrollIntoView is not a function
 ❯ src/components/input/select/AutoCompleteSelect.tsx:114:30
    112|
    113|   useEffect(() => {
    114|     activeOptionRef.current?.scrollIntoView({ block: "nearest" });
       |                              ^
    115|   }, [activeIndex]);
```

tresdonicf (approved):

PR #1610, opened 2026-07-21
DEMOS-2469 User needs to refresh the page to see uploaded BN workbook

nunesmlucas:
The upload dialog blocked on `waitForBNValidation` — a poll of up to **60s** — while the
document row is created much earlier, right after the virus scan. So the file showed up in
the table while the modal kept spinning. QA refreshed at ~20s and reported it as stuck.

The wait did nothing anyway: an unfinished validation returned `Pending`, which isn't
`"Failed"`, so the dialog treated it as success — after burning the full 60 seconds.

## Fix

Removed the blocking BN validation wait. Flow is now mutation -> S3 PUT -> virus scan ->
refetch -> close. Validation still runs server-side; the dialog no longer waits on it.

Safe because the same ruleset already runs client-side before upload
(`useBNWorkbookPreValidation`) and disables the Upload button on failure.

Also wrapped `onSubmit` in `try`/`catch`. A throw previously left the dialog on
`uploading`, which disables both Cancel and Upload — unescapable without a refresh.

## Key changes

- `client/.../AddDocumentToDeliverableDialog.tsx` — removed the blocking BN wait
- `client/.../DocumentDialog.tsx` — `try`/`catch` so the dialog can't latch on `uploading`
- `server/.../LocalS3Adapter.ts` — local dev only; `LOCAL_SIMPLE_UPLOAD` never created the
  `budget_neutrality_workbook` row, so a constraint trigger rolled back every local BN
  upload. No production impact.

### DIFF
--- a/client/src/components/dialog/document/AddDocumentToDeliverableDialog.tsx
+++ b/client/src/components/dialog/document/AddDocumentToDeliverableDialog.tsx
@@ -116,11 +116,8 @@ export const AddDocumentToDeliverableDialog: React.FC<AddDocumentToDeliverableDi
       return "virus-scan-failed";
     }
 
-    // Budget Neutrality validation runs asynchronously server-side and can take far longer than a
-    // dialog should stay open (GuardDuty latency alone is unbounded). Blocking on it here left the
-    // modal spinning long after the document had already landed in the table. The same ruleset runs
-    // client-side via useBNWorkbookPreValidation before upload is permitted, so nothing is lost by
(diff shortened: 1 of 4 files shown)


dustbuster:
I pullled this down to test it. More for my own understanding. This does fix the issue of the refresh. Got `tsx Could not Fast Refresh` before, but now that's gone! nice work!

dustbuster (approved):
See comment. No concerns and this works as designed!

PR #1611, opened 2026-07-22
DEMOS 2087 Update references to State/Territory to name instead of ID

JeffreyM-GA:
Updates State/Territory to name instead of ID in:
- Demonstration Tables
- Deliverable Tables
- Demonstration Detail Header
- Reports

image


tresdonicf (approved):
`client/` side changes look good! I don't know much about the `onDemandReports` however

TARehman (approved):
Report changes all look fine to me, and if they were run in local dev or equivalent, that confirms that the Zod checks are working as well. Didn't look over the front end but all the Excel stuff is 👍🏻 from me.

PR #1615, opened 2026-07-22
DEMOS-2446: basic date/phase crosswalk

connor-parke-icf:
This is a fairly complicated migration, mainly due to the logic with dates that exists in the API layer instead of the database. Another layer of complication is that many of the dates that we have defined in DEMOS dont exist in PMDA, and if they do, more likely than not they are null. So assumptions must be made, mainly pertaining to the phase start and end dates. 

The flow is largely outlined as follows: 
PMDA application table is pulled in then crosswalked. In this crosswalk, we initialize any missing start and end phase dates. Then the dates are validated; this is done in two steps as some date validations are dependent on validations on other dates so if one is made invalid, this should cascade to the other dates. Mainly this is in the start and end dates.  Once date validation flags are in place, we can filter the failing dates. Finally, these dates are pivoted into the shape of the application date, where they can be exported to errors or passed along to the final dates. 

Phase statuses are also derived, since the current phase is derived from the existence of the completion date. This status is recombined back with the original demonstrations to set the current_phase_id. 

At the end, we have pending demonstrations, pending applications, application dates and application phases.


data/migration/stage_pmda_for_migration/models/apps/apps_active_in_progress_pmda_demos.sql
```diff
@@ -1,4 +1,6 @@
 SELECT
+    gen_random_uuid() AS id,
```
TARehman: It's odd to me to generate this here and not in `cleaned` or possibly even `final`.

data/migration/stage_pmda_for_migration/models/apps/apps_active_in_progress_pmda_demos.sql
```diff
@@ -8,15 +10,15 @@ SELECT
...
     aplctn.mdcd_demo_aplctn_stus_dt,
     pendg.geo_ansi_state_cd,
-    pendg.mdcd_chip_div_cd
+    pendg.mdcd_chip_div_cd,
+    pendg.proj_ofcr_user_id
 FROM
     {{ source('legacy_pmda_raw', 'mdcd_pendg_demo') }} AS pendg
-INNER JOIN
-    {{ source('legacy_pmda_raw', 'mdcd_demo_aplctn') }} AS aplctn
+INNER JOIN {{ source('legacy_pmda_raw', 'mdcd_demo_aplctn') }} AS aplctn
```
TARehman: Are we actually safe to do an inner here, or should we do a LEFT with some kind of filter for errors later?

data/migration/stage_pmda_for_migration/models/apps/apps_in_prog_dates_wide_normalized.sql
```diff
@@ -0,0 +1,60 @@
+-- Apply timezone conversion and start-of-day/end-of-day timestamps based on DEMOS date type semantics
+-- (see DATE_TYPES_WITH_EXPECTED_TIMESTAMPS in constants.ts)
+
+SELECT
+    id,
+    mdcd_demo_aplctn_id,
+
+    -- Concept Phase - all Start of Day
+    (concept_start_date + TIME '00:00:00.000') AT TIME ZONE 'America/New_York' AS concept_start_date,
```
TARehman: So strictly speaking, I'm realizing that adding the TIME here isn't really doing anything in terms of the data - what it's actually doing is casting concept_start_date to a TIMESTAMP implicitly. If we just did a cast it would work as well, I think. To be clear, this is because we're aiming at midnight here - if we just did concept_start_date::TIMESTAMP then it defaults to midnight. I think its fine to keep doing it this way, however, for explicitness. We might want to think about a macro for this since we do it a lot.

data/migration/stage_pmda_for_migration/models/apps/apps_in_prog_dates_crosswalked.sql
TARehman, line 1: So, this looks like basically a column rename with some basic logic in here. I don't like all the commented-out NULL code. If that needs documented, let's document it somewhere else.

data/migration/stage_pmda_for_migration/models/apps/apps_in_prog_dates_crosswalked.sql
```diff
@@ -0,0 +1,85 @@
...
+    -- Approval Summary Phase - can be assumed to be skipped for in progress demos
+    -- NULL AS application_details_marked_complete_date, -- No valid mapping from PMDA
+    -- NULL AS application_demonstration_types_marked_complete_date, -- No valid mapping from PMDA
+    -- NULL AS approval_summary_start_date, -- No valid mapping from PMDA
+    -- NULL AS approval_summary_completion_date, -- No valid mapping from PMDA
+
+    -- NULL AS application_approval_date -- In Progress demos can be assumed to not be approved
+
+FROM {{ ref('apps_active_in_progress_pmda_demos') }} AS apps
+INNER JOIN {{ source('legacy_pmda_raw', 'mdcd_demo_aplctn') }} AS mdcd_demo_aplctn
```
TARehman: This feels redundant with the item before it as you're joining to the same table you joined to previously. I'd rather have us pull the columns in the previous model where we're already doing the join once, and then put all the renaming logic here if the point was to make it separate.

data/migration/stage_pmda_for_migration/models/apps/apps_in_prog_dates_wide_normalized.sql
```diff
@@ -0,0 +1,60 @@
...
+    (concept_completion_date + TIME '00:00:00.000') AT TIME ZONE 'America/New_York' AS concept_completion_date,
+    (concept_skipped_date + TIME '00:00:00.000') AT TIME ZONE 'America/New_York' AS concept_skipped_date,
```
TARehman: Also note that your formats aren't consistent here, when I ran this crosswalked (the preceding model) has concepts as timestamps already. A consequence of that is that you get dates like `2021-08-27 16:15:18.000000` in the concept phase because you're adding 0 to a thing that isn't zero to start.

data/migration/stage_pmda_for_migration/models/apps/apps_in_prog_dates_validated.sql
```diff
@@ -0,0 +1,138 @@
+WITH all_dates AS (
```
TARehman: I don't think I understand what value we're getting from this CTE in particular.

data/migration/stage_pmda_for_migration/models/apps/apps_in_prog_dates_validated.sql
```diff
@@ -0,0 +1,138 @@
...
+    SELECT
+        id,
+        mdcd_demo_aplctn_id,
+        date_type,
+        date_value
+    FROM {{ ref('apps_in_prog_dates_unpivoted') }}
+),
+
+-- Self-join to get related dates for validation
+dates_with_context AS (
```
TARehman: This feels really heavy and redundant. You're essentially repivoting the data you just unpivoted so you can run the validation? Why not do the validations on the already pivoted data before you unpivot it?

data/migration/stage_pmda_for_migration/models/apps/apps_in_prog_dates_validated.sql
```diff
@@ -0,0 +1,138 @@
...
+)
+
+SELECT
+    id,
+    mdcd_demo_aplctn_id,
+    date_type,
+    date_value,
+
+    -- Validation failures (NULL if passed, string describing failures if failed)
+    CASE
```
TARehman: Because of how this CASE statement works, you only get the first failure, which is fine, but its a reason to consider not using the long table to do this, and instead, to have this in the already-wide table with a column for each validation so we can see exactly what tests fail.

data/migration/stage_pmda_for_migration/models/apps/apps_in_prog_phase_statuses.sql
TARehman, line 1: So we _unpivot_ the dates, then validate by repivoting them, then have to repivot them in the CTE in order to make the phase statuses, and then later you unpivot them again? I think we need to refactor the flow of this because it seems kinda redundant.

data/migration/stage_pmda_for_migration/models/apps/apps_unfiltered_staged_in_prog_pmda_demos.sql
```diff
@@ -1,5 +1,5 @@
 SELECT
-    NULL AS mdcd_demo_id,
```
TARehman, line 2: This feels dangerous, again, largely because we have introduced our DEMOS ID a bit too early in the process.

data/migration/stage_pmda_for_migration/models/cleaned/final/final_demos_app_application_phase.sql
```diff
@@ -1 +1,3 @@
 SELECT * FROM {{ ref('cleaned_demos_app_app_phases_finalized_demos') }}
+UNION
```
TARehman: Not that it's likely to matter here, but you should use UNION ALL; UNION scans and attempts to detect duplicates which is not needed here and impacts performance.

data/migration/stage_pmda_for_migration/models/cleaned/final/final_demos_app_demo_role_assignment.sql
```diff
@@ -1 +1,3 @@
 SELECT * FROM {{ ref('cleaned_demos_app_demo_role_prim_po_finalized_demos') }}
+UNION
```
TARehman: UNION ALL

data/migration/stage_pmda_for_migration/models/cleaned/final/final_demos_app_demonstration.sql
```diff
@@ -1 +1,3 @@
 SELECT * FROM {{ ref('cleaned_demos_app_demonstration_finalized_demos') }}
+UNION
```
TARehman: UNION ALL

data/migration/stage_pmda_for_migration/models/cleaned/final/final_demos_app_primary_demo_role_assignment.sql
```diff
@@ -1 +1,3 @@
 SELECT * FROM {{ ref('cleaned_demos_app_prim_demo_role_prim_po_finalized_demos') }}
+UNION
```
TARehman: UNION ALL

data/migration/stage_pmda_for_migration/models/cleaned/cleaned_demos_app_demonstration_finalized_demos.sql
```diff
@@ -16,7 +16,6 @@ SELECT
     chip_id,
     current_timestamp AS created_at,
     current_timestamp AS updated_at,
-    mdcd_demo_id AS _legacy_mdcd_demo_id,
```
TARehman, line 19: Why is this being removed? I'm using it on my branch.

data/migration/stage_pmda_for_migration/models/cleaned/cleaned_demos_app_app_phases_in_prog_demos.sql
```diff
@@ -0,0 +1,12 @@
+SELECT
+    d.id AS application_id,
+    psu.phase_id,
+    psu.phase_status_id,
+    current_timestamp AS created_at,
+    current_timestamp AS updated_at
+FROM
+    {{ ref('cleaned_demos_app_demonstration_in_prog_demos') }} AS d
+INNER JOIN
```
TARehman: Every place we do an INNER JOIN we have to be sure that there, or at some point before, we've done a test or a filter that ensures that we're not silently filtering out rows, because that is what happens in an INNER JOIN. How do we know that isn't going to happen here? We'd have to have derived `apps_in_prog_phase_statuses_unpivoted` fro the table in the original `FROM` statement, or some other path. There may be a test configured to cover this, however.

data/migration/stage_pmda_for_migration/models/cleaned/cleaned_demos_app_application_date_in_prog_demos.sql
```diff
@@ -0,0 +1,20 @@
...
+    current_timestamp AS created_at,
+    current_timestamp AS updated_at
+FROM
+    {{ ref('apps_in_prog_dates_unpivoted') }} AS dates
+INNER JOIN
+    {{ ref('cleaned_demos_app_demonstration_in_prog_demos') }} AS demos
+    ON
+        dates.id = demos.id
+WHERE
+    NOT EXISTS (
```
TARehman: You can just look for the ID to be not in the error IDs, I think.

TARehman (changes requested):
I made it through a partial review but I think this needs a refactor; we are adding a lot of redundant steps along the way that I think we can minimize, and there's concerns about the join strategy in a few places.

data/migration/stage_pmda_for_migration/models/apps/apps_active_in_progress_pmda_demos.sql
connor-parke-icf: was able to move this further down the stack in the latest iteration

data/migration/stage_pmda_for_migration/models/apps/apps_in_prog_dates_wide_normalized.sql
connor-parke-icf: was able to add a test that should ensure the map of start of day/end of day is upheld for DC time

data/migration/stage_pmda_for_migration/models/apps/apps_in_prog_dates_crosswalked.sql
connor-parke-icf: refactored this, now does the crosswalking and the time-mapping in one step. Removed the unneeded CTE and simplified the flow

data/migration/stage_pmda_for_migration/models/apps/apps_in_prog_dates_crosswalked.sql
connor-parke-icf: removed redundant joins, passing through the needed columns as i go

data/migration/stage_pmda_for_migration/models/apps/apps_active_in_progress_pmda_demos.sql
connor-parke-icf: made this a left join and hooked up an error table to it

TARehman (approved):
Live review. Work in progress so we know there's some additional items to do, @connor-parke-icf will be updating `MIGRATION_LOGIC.md` to document and will open additional PRs as we go forward.

PR #1616, opened 2026-07-22
DEMOS-2387 Fix underscores in downloaded file names

nunesmlucas:
Downloading a Word/Excel document produced underscores in the name:

```
Document Title:  DEMOS-892 (3) \ / : * ? " < > |
Downloaded as:   DEMOS-892 (3)_________
```

Chrome adds underscores when it's handed a file name with illegal
characters. So the underscores proved our sanitizer never ran.

The server computes the download name and exposes it, so the client never builds a
file name itself.

- New `downloadFileName` field on the `Document` GraphQL type. Computed: sanitized
  `name` plus the extension from the S3 Content-Type.
- Extracted `buildDownloadFileName` in `AwsS3Adapter`, now shared by both the
  `Content-Disposition` header and the new field so they can't drift.
- The client queries that field and uses it for the download.

The heading and Title column still show the raw `name`. Only the downloaded file gets
a cleaned name.

## Why we renamed `filename` to `downloadFileName`

Just a rename, nothing behaves differently. The prop used to get the raw title and now gets the cleaned-up download name. If we left it called filename, it would be easy for someone to plug the raw title back in and bring the underscores back. The new name says what it actually holds.

## Invalid characters now become a space

They used to be dropped, which fused tokens:

| Title | Before | Now |
| --- | --- | --- |
| `Budget FY25/26` | `Budget FY2526` | `Budget FY25 26` |
| `Budget*****FY26` | `BudgetFY26` | `Budget FY26` |

Runs of invalid characters collapse to one space, and leading/trailing whitespace is
trimmed, so `DEMOS-892 (3) \ / : * ? " < > |` downloads as `DEMOS-892 (3).docx`.

Database impact: none

https://github.com/user-attachments/assets/1bdac711-086d-46fc-acf4-465d181a4085

### DIFF
--- a/client/src/pages/DocumentDetails/DocumentDetail.test.tsx
+++ b/client/src/pages/DocumentDetails/DocumentDetail.test.tsx
@@ -12,16 +12,21 @@ import {
 
 // Mock DocumentPreview component
 vi.mock("./DocumentPreview", () => ({
-  DocumentPreview: ({ filename }: { filename: string; presignedDownloadUrl: string }) => (
-    <div data-testid="document-preview">Preview: {filename}</div>
-  ),
+  DocumentPreview: ({
+    downloadFileName,
+  }: {
+    downloadFileName: string;
+    presignedDownloadUrl: string;
+  }) => <div data-testid="document-preview">Preview: {downloadFileName}</div>,
 }));
 
+// The title keeps whatever the user typed; only the download name is sanitized.
 const mockDocument = {
   id: "doc-123",
-  name: "test-document.pdf",
+  name: 'test-document \\ / : * ? " < > |',
   createdAt: "2026-02-10T10:00:00Z",
   presignedDownloadUrl: "https://example.com/presigned-url",
+  downloadFileName: "test-document.pdf",
   application: {
     __typename: "Demonstration" as const,
     id: "demo-123",
@@ -160,8 +165,9 @@ describe("DocumentDetail", () => {
   it("renders document details successfully", async () => {
     renderDocumentDetail("doc-123", [DocumentDetailMock]);
 
+    // The heading shows the title exactly as the user saved it, invalid characters included.
     await waitFor(() => {
-      expect(screen.getByText("test-document.pdf")).toBeInTheDocument();
+      expect(screen.getByText(mockDocument.name)).toBeInTheDocument();
     });
 
     expect(screen.getByText("Test Demonstration")).toBeInTheDocument();
@@ -227,7 +233,7 @@ describe("DocumentDetailPage", () => {
     renderWithRouter();
 
     await waitFor(() => {
-      expect(screen.getByText("test-document.pdf")).toBeInTheDocument();
+      expect(screen.getByText(mockDocument.name)).toBeInTheDocument();
     });
   });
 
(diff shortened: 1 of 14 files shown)


client/src/pages/DocumentDetails/DocumentDetail.test.tsx
```diff
@@ -12,16 +12,21 @@ import {
...
+  DocumentPreview: ({
+    downloadFileName,
+  }: {
+    downloadFileName: string;
+    presignedDownloadUrl: string;
+  }) => <div data-testid="document-preview">Preview: {downloadFileName}</div>,
 }));
 
+// The title keeps whatever the user typed; only the download name is sanitized.
 const mockDocument = {
```
dustbuster, line 24: Missed opportunity to call it a Mocument. :-)

client/src/pages/DocumentDetails/DocumentDetail.tsx
```diff
@@ -128,7 +132,7 @@ export const DocumentDetail: React.FC<{ documentId: string }> = ({ documentId })
       <div className="mt-2 flex-1 min-h-0">
         <DocumentPreview
           presignedDownloadUrl={document.presignedDownloadUrl}
-          filename={document.name}
```
tresdonicf, line 131: My instinct is suggesting that we might be able to wrap this `document.name` in a function to clean up the underscores, would this not be possible? Or is it that we would have to strip out spaces so the browser doesn't turn them into underscores?

nunesmlucas: Correct, we have to strip them because browser is handling it like a pain

tresdonicf (commented):
Client-side changes look okay though I do still think we might be able to do this with a smaller surface area, perhaps by using the `filename` parameter of the `downloads` api if there's nothing preventing us from moving to this: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/downloads/download#filename

If we want to keep this approach we might just need eyes from the B/E since there's a good amount changing in the `server`

server/src/model/document/documentSchema.ts
```diff
@@ -32,6 +32,8 @@ export const documentSchema = gql`
     application: Application! @auth(requires: ["Access CMS Field"])
     phaseName: PhaseName @auth(requires: ["Access CMS Field"])
     presignedDownloadUrl: String! @auth(requires: ["Access CMS Field"])
+    "The sanitized file name, with extension, that this document should download as."
+    downloadFileName: String! @auth(requires: ["Access CMS Field"])
```
connor-parke-icf, line 36: i think we want this to be accessible to any user, since state users should be able to download documents on Deliverables.  But i also see that the presignedDownloadUrl is restricted to CMS Users. this seems odd to me, are state users currently unable to download documents theyve uploaded to deliverables?

connor-parke-icf (approved):
Approving, but we should follow up on whether or not state users can download documents from deliverables. Looking at the code, this seems to imply that State users cannot, which feels wrong to me. 

I think we may have been able to do this entirely on the frontend, since this document name isnt part of the url and looks like we're just using it to pass into DocumentPreview. That being said, we're likely going to restructure this page in the near future with a different previewer so these issues might just go away.

PR #1619, opened 2026-07-24
Demos graphql nplus1 dataloaders

connor-parke-icf:
## Changes

-

## Automated review

- [x] Run AI Code Review

PR #1620, opened 2026-07-24
Hotfix/demonstration deliverable table

connor-parke-icf:
Frontend bugfix: we were fetching all deliverables instead of just the subset that belong to a demonstration on the demonstration deliverables table. Fix is just to wrap the query as a demonstration query, subresolving its deliverables. For this, i changed as little as possible but future improvements could be to reduce the amount being queried here. This is in tandem with the [dataLoaders implementation](https://github.com/Enterprise-CMCS/demos/pull/1619) to reduce the calls to the database and speed up load times. Though that fix addresses the bulk of the issue, this query needed to be changed as well. 

## Automated review

- [x] Run AI Code Review

### DIFF
--- a/client/src/pages/DemonstrationDetail/deliverables/DeliverablesTab.tsx
+++ b/client/src/pages/DemonstrationDetail/deliverables/DeliverablesTab.tsx
@@ -6,15 +6,70 @@ import { useDialog } from "components/dialog/DialogContext";
 import { AddDeliverableSlotDemonstration } from "components/dialog/deliverable/AddDeliverableSlotDialog";
 import { DemonstrationDeliverableTable } from "components/table/tables/DemonstrationDeliverableTable";
 import { getCurrentUser } from "components/user/UserContext";
-import type { UserType } from "demos-server";
-import {
-  DELIVERABLES_PAGE_QUERY,
-  type DeliverablesQueryResult,
-} from "components/table/tables/DeliverableTable";
-import { useQuery } from "@apollo/client";
+import type { Tag, UserType } from "demos-server";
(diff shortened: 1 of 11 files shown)


client/src/pages/DemonstrationDetail/deliverables/DeliverablesTab.tsx
```diff
@@ -6,16 +6,71 @@ import { useDialog } from "components/dialog/DialogContext";
...
+import { type DeliverablesQueryResult } from "components/table/tables/DeliverableTable";
+import { gql, useQuery } from "@apollo/client";
 import { useNavigate } from "react-router-dom";
 
 export const ADD_DELIVERABLE_SLOT_BUTTON_NAME = "button-add-deliverable-slot";
 
+export const DEMONSTRATION_DELIVERABLE_TAB = gql`
+  query GetDeliverablesPage($id: ID!) {
+    demonstration(id: $id) {
+      deliverables {
```
connor-parke-icf, line 19: Previously we were using the same query on the demonstration deliverables table as the main all deliverables table. This meant that we were fetching thousands of deliverables (subresolving each), which led to slowdowns. Fix was to just query the deliverables on the current demonstration instead of all deliverables. So copy/pasted the query here and wrapped it as a demonstration query.

dustbuster: Yea, i supposed thinking we could load everything and it be okay was a bit naive. I dont think we considered deliverables. And that there are 7000 of them!

dustbuster (approved):
Good catch m8!

PR #1623, opened 2026-07-27
DEMOS-2362: split the demo detail query - this splits in a manner that may not be worth it. backburnering it. 

dustbuster:
## Summary

This is a medium, focused refactor. Keep the main Demonstration tab unchanged, but stop loading every amendment, extension, and modification document on initial page load.

Initial load becomes:

1. Demonstration shell data and modification counts.
2. Amendment or extension list when its outer tab opens.
3. Full workflow/details/documents for only the selected modification.

## Changes

- Replace `amendments` and `extensions` in the initial query with `amendmentCount` and `extensionCount`.
- Add those two GraphQL fields to `Demonstration`, backed by one request-scoped loader that obtains both relation counts.
- Apply the same CMS-field authorization as the existing collections.
- Add lightweight client queries:
  - `GetDemonstrationAmendments`: `id`, `name`, `createdAt`
  - `GetDemonstrationExtensions`: `id`, `name`, `createdAt`
- Run each list query inside its corresponding tab. The existing `Tabs` component already mounts only the selected tab, so no tab-system rewrite is needed.
- Preserve newest-first selection and URL selection such as `?amendments=<id>` and `?extensions=<id>`.

## Selected Modification

- Change `ModificationTabs` to accept lightweight list items.
- Once an item is selected, run the existing amendment or extension workflow query for that ID.
- Lift that query to a selected-modification container and pass the loaded object to:
  - Application workflow
  - Details summary
  - Documents table
- Keep documents in the workflow query because the workflow phases use them. This still limits document loading to one selected modification.
- Apollo’s normalized cache should retain previously selected modifications when switching tabs.

## Cache and Mutation Updates

- Rename the monolithic query to reflect its shell purpose and update its consumers.
- After creating a modification, refetch only the shell/count query and the matching amendment or extension list.
- Let edit mutations update normalized modification fields from their response; remove the full-detail refetch.
- After modification document or workflow changes, refetch only the selected amendment/extension workflow query.
- Replace remaining mutation references to the old monolithic query with the smallest affected shell, list, or selected-item query.

## Testing details:

Examples:
- Tested locally using...
- Verified in Chrome

Actual Unit testing:
- Verify initial page load does not request amendment lists, extension lists, or modification documents.
- Verify opening each outer tab triggers only its lightweight list query.
- Verify the newest item and URL-selected item trigger only that item’s workflow query.
- Verify empty-list states and tab counts remain correct.
- Verify switching items updates workflow, summary, and documents together.
- Verify creation updates counts and list contents; edits and uploads refresh only targeted data.
- Update GraphQL resolver tests, Apollo mocks, tab tests, and mutation-refetch expectations.
- 

https://github.com/user-attachments/assets/06953cc9-d881-45ac-ace7-dc2b9fd52314

## Risk and rollback
This is quite a change. May cause some PR conflicts. 
But easy to verify, harder to roll back as time goes on. 

## Automated review

- [ ] Run AI Code Review

### DIFF
--- a/client/src/components/application/amendment/AmendmentWorkflow.test.tsx
+++ b/client/src/components/application/amendment/AmendmentWorkflow.test.tsx
@@ -1,14 +1,26 @@
 import React from "react";
-import { describe, expect, it } from "vitest";
+import { describe, expect, it, vi } from "vitest";
 import { render, screen } from "@testing-library/react";
 import { AmendmentWorkflow } from "./AmendmentWorkflow";
 import { TestProvider } from "test-utils/TestProvider";
+import type { ApplicationWorkflowAmendment } from "./AmendmentWorkflow";
+
+vi.mock("components/application/phase-selector/PhaseSelector", () => ({
+  PhaseSelector: () => <div data-testid="phase-selector" />,
+}));
 
 describe("AmendmentWorkflow", () => {
   it("renders APPLICATION heading", async () => {
     render(
       <TestProvider>
-        <AmendmentWorkflow amendmentId="1" />
+        <AmendmentWorkflow
+          amendment={
+            {
+              id: "1",
+              status: "Pre-Submission",
+            } as ApplicationWorkflowAmendment
+          }
+        />
       </TestProvider>
     );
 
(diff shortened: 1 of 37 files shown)


dustbuster:
Closing for meow.

PR #1624, opened 2026-07-27
(2/3) DEMOS-2386: Persist dates in first half of Application Workflow

tresdonicf:
## Summary

Utilize the new `useSessionStorage` to ensure dates are persisted when a user navigates away from a step. 

## Changes

- Plain `useState` calls have been replaced with `useSessionStorage` as a drop-in replacement
- Removed an unused parameter from `getConceptPhaseComponentFromApplication`

## Validation

- Tested manually (screenshots attached) and existing unit tests pass

## Screen Recordings

https://github.com/user-attachments/assets/13942017-a541-4b90-b0c5-248662a878da

## Next Steps

The last PR for this work will be to implement this for the second half. Those phases use a different strategy of having the dates together in a form field so I'm strategizing about how to make the change there as simple as possible

### DIFF
--- a/client/src/components/application/phase-selector/PhaseSelector.tsx
+++ b/client/src/components/application/phase-selector/PhaseSelector.tsx
@@ -115,11 +115,7 @@ export const PhaseSelector = ({
   const renderPhase = (phaseName: PhaseName) => {
     switch (phaseName) {
       case "Concept":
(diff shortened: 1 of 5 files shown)


dustbuster (approved):
LGTM! 🚀

PR #1626, opened 2026-07-27
DEMOS-2503 BN workbook `actuals` column is null after upload

nunesmlucas:
# DEMOS-2503: BN workbook `actuals` column is null after upload

## What was broken

Every BN upload since 7/13 left its row stuck at `Pending` with `actuals` null, on TEST and
IMPL. The Lambda crashed the moment it loaded and never opened the spreadsheet. From IMPL
CloudWatch:

```
"errorMessage":"Dynamic require of \"node:os\" is not supported"
"durationMs":0.0
```

## Why pino has to come back

The Lambda has always used pino — `log.ts` builds the logger from `shared_library`, and that
didn't change. What changed on 7/10 (`ba6e52ed`) is *how it gets packaged*: pino stopped being
installed next to the bundle and started being inlined into it.

pino is CommonJS. This Lambda builds as an ES module. Inlining a CommonJS package into an ES
module breaks its internal `require()` — and since the logger is created at import time, the
whole module fails to load on every invocation.

So this isn't a new dependency. It's restoring how pino was packaged before 7/10, which is
exactly what UIPath (also ESM) still does:

- `deployment/lib/budgetNeutralityProcessor.ts` — `nodeModules: ["pg", "pino"]`
- `lambdas/budgetNeutrality/package.json` + lockfile — declare `pino`, or CDK fails at synth
  with `Cannot extract version for module 'pino'`

## Second bug, also fixed

An extractor throwing on a bad cell used to re-throw without writing to the DB — the same stuck
`Pending` row. Now each rule and extractor runs in its own try/catch, readable values are kept,
and the Lambda always writes a status.

- `shared_library/src/BN/validation.ts`, `lambdas/budgetNeutrality/index.ts`

## Notes

- The devcontainer builds this Lambda as CommonJS, so this bug can't reproduce locally — only
  the ESM build shows it.
- Lockfile was regenerated on Node 24; doing it on older Node strips `libc` fields and trips
  Snyk's sync check.
- Snyk failures in CI are pre-existing on `main`, not from this PR — raised in the dev channel.
- Follow-up: `B8`/`AS436` are hardcoded cells, so a different Summary layout still fails — just
  visibly now instead of silently.

### DIFF
--- a/lambdas/budgetNeutrality/index.test.ts
+++ b/lambdas/budgetNeutrality/index.test.ts
@@ -55,13 +55,6 @@ vi.mock("demos-shared-library/BN/validation", () => ({
 
 import { handler } from "./index";
 import { documentExists } from "./budgetNeutralityValidation";
-import { parseBNFileFromPath } from "demos-shared-library/BN";
-import { validateBNWorkbook } from "demos-shared-library/BN/validation";
-
-const BN_WORKBOOK_EVENT = {
-  Records: [{ body: JSON.stringify({ documentId: "doc-1", documentTypeId: "BN Workbook" }) }],
-} as unknown as SQSEvent;
-const CONTEXT = { awsRequestId: "test-request-id" } as Context;
 
 describe("budgetNeutrality index", () => {
   beforeEach(() => {
@@ -112,8 +105,8 @@ describe("budgetNeutrality index", () => {
       "BN Workbook",
       "Succeeded",
       "[]",
-      null,
-      null,
+      undefined,
+      undefined,
     ]);
     expect(mocks.logInfoMock).toHaveBeenCalledWith(
       {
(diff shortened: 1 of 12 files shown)


lambdas/budgetNeutrality/index.ts
```diff
@@ -45,11 +46,27 @@ export async function updateBudgetNeutralityWorkbook(
...
+export async function parseAndValidateWorkbook(documentPath: string): Promise<ValidationResult> {
+  try {
+    const parsedData = await parseBNFileFromPath(documentPath);
+    return await validateBNWorkbook(parsedData, validations, extractorFunctions);
+  } catch (error) {
+    const parseError: ValidationError = {
+      code: PARSE_ERROR_CODE,
+      message: `Unable to read the workbook: ${(error as Error).message}`,
+    };
+    log.error({ error: parseError.message }, "Unable to parse BN workbook.");
```
james-becwar-cms: Lets add the documentPath so we have something to go on to debug the error.

lambdas/budgetNeutrality/index.ts
```diff
@@ -45,11 +46,27 @@ export async function updateBudgetNeutralityWorkbook(
...
+// Never throws: an unreadable workbook is a problem with the file, not this Lambda. Throwing
+// here would leave the budget_neutrality_workbook row stuck at "Pending" with no actuals.
+export async function parseAndValidateWorkbook(documentPath: string): Promise<ValidationResult> {
+  try {
+    const parsedData = await parseBNFileFromPath(documentPath);
+    return await validateBNWorkbook(parsedData, validations, extractorFunctions);
+  } catch (error) {
+    const parseError: ValidationError = {
+      code: PARSE_ERROR_CODE,
+      message: `Unable to read the workbook: ${(error as Error).message}`,
```
james-becwar-cms:
```suggestion
      message: `Unable to read the workbook: ${error instanceof Error ? error.message : String(error)}`,
```

lambdas/budgetNeutrality/index.test.ts
```diff
@@ -122,6 +129,49 @@ describe("budgetNeutrality index", () => {
...
+    expect(response).toEqual({ statusCode: 200, body: "Processed 1 records." });
+    expect(mocks.queryMock.mock.calls[0]?.[1]).toEqual([
+      "doc-1",
+      "BN Workbook",
+      "Failed",
+      JSON.stringify([{ code: "2", message: "Reporting Year value is missing." }]),
+      "Actuals Only",
+      null,
+    ]);
+  });
```
james-becwar-cms: If the validation fails the extracted values are not considered valid, and we should not store them.

lambdas/budgetNeutrality/index.ts
```diff
@@ -45,11 +46,27 @@ export async function updateBudgetNeutralityWorkbook(
...
+    };
+    log.error({ error: parseError.message }, "Unable to parse BN workbook.");
+    return { isValid: false, errors: [parseError], extractedValues: new Map() };
```
james-becwar-cms: I think the errors: here won't map to extraction errors, this is changing the signature of the function.

lambdas/budgetNeutrality/index.ts
```diff
@@ -80,14 +97,13 @@ export const handler = async (event: SQSEvent, context: Context) =>
...
 
-      log.info("Parsing completed. Starting validation against ruleset.");
-      const validationResults = await validateBNWorkbook(parsedData, validations, extractorFunctions);
-      
-
-      log.info("Validation completed. Inserting BN results into database.");
+      log.info(
+        { isValid: validationResults.isValid, errorCount: validationResults.errors.length },
+        "Validation completed. Inserting BN results into database."
+      );
```
james-becwar-cms: I wouldn't combine parse and validate above, I would add my error handling here.

shared_library/src/BN/validation.test.ts
```diff
@@ -55,11 +60,66 @@ describe("validateBNWorkbook", () => {
     expect(result).toEqual({
       isValid: false,
       errors: [expectedError],
-      extractedValues: new Map(),
+      extractedValues: new Map([["actuals", "FY25"]]),
```
james-becwar-cms: If the validation is not valid we should not extract values.

shared_library/src/BN/validation.ts
```diff
@@ -7,39 +10,78 @@ export type ValidationError = {
...
+function toMessage(cause: unknown): string {
+  return cause instanceof Error ? cause.message : String(cause);
+}
 
-export async function validateBNWorkbook(data:ExcelData, validations: ValidationFunction[], extractionFunctions: ExtractionFunction[]): Promise<ValidationResult>{
-  
-  const errors: ValidationError[] = validations.map((validation) => validation(data)).filter((error) => error !== null) as ValidationError[] ;
-  const isValid = errors.length === 0;
+// A rule that throws is reported as an error rather than aborting the run, so callers
+// always get a result they can persist.
```
james-becwar-cms: This prevents the DLQ from working.

shared_library/src/BN/validation.ts
```diff
@@ -7,39 +10,78 @@ export type ValidationError = {
...
+        message: `Workbook is missing expected structure: ${toMessage(ruleError)}`,
+      });
+    }
+  }
+
+  return errors;
+}
+
+// Extractors run independently so one failure still yields the other values. A failed
+// extraction is an error, which keeps a partially extracted workbook from reading as valid.
```
james-becwar-cms: We do not want it to work this way.

james-becwar-cms (changes requested):
In general we don't want to extract information if validation is not successful.

james-becwar-cms (approved):

PR #1636, opened 2026-07-29
Demos 2446 additional refactors and cleanup

connor-parke-icf:
This PR addresses concerns in the previous PR for in progress applications/application dates. 
- Refactor phase completion logic to break up into multiple CTEs. Because the phase status should cascade depending on the status of the previous phase, the clearest way to represent this using DBT is a series of 7 CTEs, each representing its own phase. (priority 1).
This ended up not being quite needed, mainly because theres a mix of true PMDA dates and derived dates in the application date crosswalk. To address this, i implemented a more comprehensive test and appropriate-enough fallbacks.
- in apps_in_prog_phase_completion we are checking against the application_intake_start_date to determine if the concept start date has been skipped. Instead, we should check against the status of the application intake start date. 
- when initializing the federal comment period status, we should load it in as "Not Started" then rely on the existing stored procedure to update the phase status according to the its own rules. 
- we should add a test to ensure that the concept phase's skipped or completed dates are mutually exclusive to eachother
- we should rewire the inclusion of the current_phase_id to a new model between the Cleaned demonstration and the final demonstration tables, so clean up the existing final demonstration table which currently includes that join. We should strive to keep that table as clean as possible, relying primerily on selects and unions. 
- in cleaned_demos_app_app_phases_in_prog_demos, we have an inner join that may silently filter. We should replace it with an left join and add a corresponding test for nullness
- we can replace the apps_in_prog_missing_aplctn test with an inline test in the models.yaml since it is simple enough and concentrates the check back to the main configuration. 

- [x] Run AI Code Review


data/migration/stage_pmda_for_migration/models/apps/apps_in_prog_dates_crosswalked.sql
```diff
@@ -16,6 +16,10 @@ SELECT
     (CASE
         WHEN apps.phase_2_rcvd_dt IS NOT NULL
             THEN (apps.phase_2_rcvd_dt - INTERVAL '1 day')
+        WHEN
+            apps.phase_2_rcvd_dt IS NOT NULL
+            OR apps.phase_2_cmpltns_rvw_dt IS NOT NULL
+            THEN (apps.creatd_dt AT TIME ZONE 'America/New_York')::DATE
```
connor-parke-icf, line 22: when validating that each phase with a date had a start date, i found a couple that didnt have one. So ive provided a fallback to the application creation date since these are mostly new demos concepts anyway. might be a more appropriate fallback, but this one is pretty guaranteed to work

data/migration/stage_pmda_for_migration/models/apps/apps_in_prog_dates_crosswalked.sql
```diff
@@ -72,7 +96,4 @@ SELECT
     (apps.phase_4_strt_dt + TIME '00:00:00.000') AT TIME ZONE 'America/New_York' AS approval_package_start_date,
     (apps.phase_4_end_dt + TIME '00:00:00.000') AT TIME ZONE 'America/New_York' AS approval_package_completion_date
 
-FROM {{ ref('apps_active_in_prog_pmda_demos') }} AS apps
-WHERE
-    apps.mdcd_pendg_demo_id NOT IN
-    (SELECT e1.mdcd_pendg_demo_id FROM {{ ref('errors_apps_in_prog_missing_aplctn') }} AS e1)
```
connor-parke-icf, line 78: this validation was moved to an inline check in the yaml

data/migration/stage_pmda_for_migration/models/apps/apps_in_prog_phase_completion.sql
```diff
@@ -54,13 +48,13 @@ SELECT
...
         WHEN
-            concept_phase_status <> 'Completed'
-            AND concept_phase_status <> 'Skipped'
-            AND application_intake_start_date IS NULL
+            concept_phase_status = 'Started'
+            AND application_intake_phase_status = 'Not Started'
             THEN 'Concept'
-        WHEN application_intake_phase_status <> 'Completed' THEN 'Application Intake'
-        WHEN completeness_phase_status <> 'Completed' THEN 'Completeness'
-        WHEN federal_comment_period_phase_status <> 'Completed' THEN 'Federal Comment'
```
connor-parke-icf, line 63: modified this to be a positive check on the phase, rather than a negative one. Also removed federal comment period since that will be updated by the stored proc, and non of the demos currently are within their federal comment periods.

data/migration/stage_pmda_for_migration/models/apps/apps_unfiltered_staged_in_prog_pmda_demos.sql
```diff
@@ -19,7 +19,26 @@ SELECT
     ipdnum.cleaned_mdcd_demo_num AS medicaid_id,
     NULL AS chip_id,
     ipd.geo_ansi_state_cd AS state_id,
-    ipd.proj_ofcr_user_id
+    ipd.proj_ofcr_user_id,
```
connor-parke-icf, line 22: found a bug where i actually was silently filtering because i was pulling from a less-clean table than i joined to later. Made the more-clean version contain the columns i needed and pulled from that instead. as a result, the relevant columns needed to be piped through.

data/migration/stage_pmda_for_migration/tests/assert_phase_statuses_match_dates.sql
```diff
@@ -0,0 +1,333 @@
+-- This test validates that phase statuses are consistent with their dates.
+-- If any rows are returned, the test fails.
+-- Rules:
+-- - If phase status is 'Started', phase must have a start date
+-- - If phase status is 'Completed', phase must have both start and completion dates
+-- - If previous phase is 'Completed', next phase must be 'Started' (except Completeness → Federal Comment Period)
+-- - If dates exist for a phase, the phase must have a start date and be 'Started' (inverse check)
+
+WITH phase_data AS (
```
connor-parke-icf, line 9: went ahead and created a comprehensive test for dates and phases. many different reasons for failing, but in general validates the dates and their relationship to the phases.

data/migration/stage_pmda_for_migration/models/models.yml
```diff
@@ -474,6 +487,10 @@ models:
     columns:
       - name: application_id
         data_type: uuid
+        data_tests:
+          - not_null:
+              name: assert_id_not_null_in_cleaned_demos_app_app_phases_in_prog_demos
+
```
TARehman, line 493:
```suggestion
```

Extra line here, not sure if it was intended.

data/migration/stage_pmda_for_migration/models/models.yml
```diff
@@ -515,6 +532,7 @@ models:
         data_type: timestamptz
 
   # Apps Models
+
```
TARehman, line 535:
```suggestion
```

Same, extra line here, not sure if intended.

TARehman (approved):
LGTM! 👍🏻

PR #1638, opened 2026-07-30
DEMOS-2557 - State user cannot download documents

nunesmlucas:
# DEMOS-2557: State user cannot download documents

State users got a 500 opening any document because the DocumentDetail query requests
fields gated to `Access CMS Field`. Removed that gate from `presignedDownloadUrl`,
`downloadFileName`, and `application` so any user can read them. Which documents a user
can reach is still enforced upstream by row-level access, so this is safe.


server/src/model/document/documentSchema.ts
```diff
@@ -29,11 +29,11 @@ export const documentSchema = gql`
     s3Path: NonEmptyString! @auth(requires: ["Access CMS Field"])
     owner: User!
     documentType: DocumentType!
-    application: Application! @auth(requires: ["Access CMS Field"])
+    application: Application!
```
connor-parke-icf, line 32: nice catch here with the Application; i imagine this come from displaying the name/link to the Application on the preview page. A document belongs only to one application, and since to be able to see the document you would already have access to its parent Application so this seems fine to me.

connor-parke-icf (approved):

PR #1641, opened 2026-07-31
Demos 2446 program monitoring documents

connor-parke-icf:
This imports migrates monitoring documents to DEMOS. These doucments have been imported as regular, phase-less General Files with their description indicating their category, as we do not have a document type for it in DEMOS. In total there were 1380 documents here. 

## Automated review

- [x] Run AI Code Review


data/migration/stage_pmda_for_migration/models/docs/docs_pmda_prgm_docs_rpstry_crosswalked_desc.sql
```diff
@@ -0,0 +1,14 @@
+SELECT
+    trim(doc.orgnl_fil_name) AS name, -- noqa: RF04
+    doc_type.description || CASE WHEN doc.arcv_ind = 1 THEN ' (Archived)' ELSE '' END AS description,
```
connor-parke-icf, line 3: we dont have a document type that corresponds directly to Monitoring Call Agenda or Monitoring Call Minutes, so making these a "General File" and putting the old file category into the description, since these files dont have a description on PMDA. In addition, theres a notion of archiving a file that is used the the SELECT used for searching for these documents in the PMDA app. We dont have a concept of "archived" in DEMOS, but it seemd somewhat important, so i just add "(Archived)" to the doc's description if that flag exists.

zoeelkins (approved):
Looks good to me. Thank you again!

PR #1661, opened 2026-08-05
DEMOS-2570: Open DEMOS Orientation Videos in New Tab

tresdonicf:
Opens the DEMOS Orientation Video link in a new tab, tested manually


dustbuster (approved):
SEND IT!

PR #1662, opened 2026-08-05
DEMOS-1880 - ADMIN: Edit State access for State users on User Management

nunesmlucas:
# DEMOS-1880: Edit State access for State users on User Management

Admins can now edit which States a State user is assigned to from the User Management tab.

Adds a row-selection column and an Edit icon, disabled ("Select a State User to Edit") until
exactly one State user is selected. It opens a new `AssignStatesDialog` with the user's name and an
alphabetical multi-select of States and Territories — no "All States" option, since CMS and Admin
users implicitly get every State and aren't editable. Assign is disabled until a State is selected;
submitting calls the existing `setPersonStates` mutation, refetches, closes, and toasts "Updates
have been made successfully". Cancel reuses `BaseDialog`'s confirm-on-change. Client-only — the
mutation shipped with DEMOS-2021.

Two things for review. The mockup shows "All States" in the dropdown, which contradicts the ticket;
I built it per the ticket. Relatedly, Assign enables on one State selected (per the AC) rather than
on has-changes like our other dialogs — happy to flip it. Separately, the server rejects removing a
State while the user is assigned to a demonstration in it, and that surfaces as a generic error
toast; worth a follow-up to show why.

https://github.com/user-attachments/assets/755a86bd-90e5-410b-849e-e96d92fa89fa


client/src/components/table/tables/UserManagementTable.tsx
```diff
@@ -55,6 +67,24 @@ export const UserManagementTable: React.FC = () => {
...
+              aria-label="Edit State Assignments"
+              tooltip={canEdit ? EDIT_STATES_ENABLED_TOOLTIP : EDIT_STATES_DISABLED_TOOLTIP}
+              disabled={!canEdit}
+              onClick={() => canEdit && showAssignStatesDialog(selected[0].person)}
+            >
+              <EditIcon />
+            </CircleButton>
+          </div>
+        );
+      }}
```
tresdonicf, line 87: Would be kinda nice to break this out into its own component instead of inling it here:

```
const actionButtons(table) = ...
...
actionButtons={actionButtons(table)}
```

client/src/components/dialog/user/AssignStatesDialog.tsx
```diff
@@ -0,0 +1,118 @@
...
+}) => {
+  const { showSuccess, showError } = useToast();
+  const [setPersonStates, { loading: saving }] = useMutation(SET_PERSON_STATES_MUTATION, {
+    refetchQueries: [USER_MANAGEMENT_QUERY_NAME],
+  });
+
+  const initialStateIds = React.useMemo(
+    () => sortByStateName(person.states.map((state) => state.id)),
+    [person.states]
+  );
```
tresdonicf, line 70: I don't think we need the `useMemo` here

client/src/components/dialog/user/AssignStatesDialog.tsx
```diff
@@ -0,0 +1,118 @@
...
+        id
+        name
+      }
+    }
+  }
+`;
+
+const sameStates = (initialStateIds: string[], selectedStateIds: string[]): boolean =>
+  initialStateIds.length === selectedStateIds.length &&
+  initialStateIds.every((stateId) => selectedStateIds.includes(stateId));
```
tresdonicf, line 53: awesome  nice breaking these out!

tresdonicf (approved):
Couple of notes but looks good all in all!

tresdonicf (approved):

PR #1665, opened 2026-08-06
Demos 2542 amendment dates phases

connor-parke-icf:
## Changes

-

## Automated review

- [ ] Run AI Code Review


data/migration/stage_pmda_for_migration/models/apps/apps_in_prog_dates_crosswalked.sql
connor-parke-icf, line 1: refactored this file to break up the regular crosswalked dates from the derived start/completion dates. Now crosswalks the dates, then uses those to determine completion dates if applicable, then based on the completion dates, sets the start dates. 
There are still some issues with validation here that i dont think we will have the opportunity to work out, however those come mainly from start dates existing on phases that have had some of their dates invalidated. This does not include any logic for documents.

data/migration/stage_pmda_for_migration/models/apps/apps_in_prog_dates_long.sql
```diff
@@ -8,63 +8,139 @@ FROM
 CROSS JOIN
     LATERAL (
         VALUES
-        ('Concept Start Date', a.concept_start_date, a.cleaned_concept_start_date),
-        ('Concept Completion Date', a.concept_completion_date, a.cleaned_concept_completion_date),
-        ('Concept Skipped Date', a.concept_skipped_date, a.cleaned_concept_skipped_date),
-        ('Application Intake Start Date', a.application_intake_start_date, a.cleaned_application_intake_start_date),
+        (
+            'Concept Start Date',
+            (a.concept_start_date + TIME '00:00:00.000') AT TIME ZONE 'America/New_York',
```
connor-parke-icf, line 13: moved the setting of the start and end dates here as made validation easier when we dont need to consider those.

data/migration/stage_pmda_for_migration/models/apps/apps_in_prog_dates_with_checks_cleaned.sql
```diff
@@ -57,13 +59,19 @@ SELECT
...
             OR federal_comment_period_start_date_missing_federal_comment_period_end_date
+            OR state_application_deemed_complete_has_invalid_submitted_date
             OR federal_comment_period_start_date_not_30_days_before_federal_comment_period_end
             THEN NULL
         ELSE federal_comment_period_start_date
     END AS cleaned_federal_comment_period_start_date,
     CASE
         WHEN
-            federal_comment_period_start_date_missing_federal_comment_period_end_date
+            state_application_deemed_complete_missing_state_application_submitted_date
```
connor-parke-icf, line 69: all of the application intake dates are validated in a set, since the validation on one impacts the validation of another. If any of those validations fail, then the whole set must be throw away

data/migration/stage_pmda_for_migration/models/apps/apps_pending_amendments.sql
```diff
@@ -3,13 +3,32 @@ SELECT
     amendment_application.demonstration_status_id,
     pending_demo.mdcd_demo_name AS name, -- noqa: RF04
     pending_demo.mdcd_demo_desc AS description,
-    pending_demo.state_prfmnc_yr_strt_dt AS effective_date,
+    (pending_demo.state_prfmnc_yr_strt_dt + TIME '00:00:00.000') AT TIME ZONE 'America/New_York' AS effective_date,
```
connor-parke-icf, line 6: missed this when adding validation to effective date start-of-day validation. adding here.

data/migration/stage_pmda_for_migration/models/cleaned/cleaned_demos_app_app_phases_finalized_demos.sql
```diff
@@ -1,22 +1,22 @@
 WITH finalized_applications AS (
-    SELECT d.id
+    SELECT demo.id
```
connor-parke-icf, line 2: updated alias per previous pr's comments.

data/migration/stage_pmda_for_migration/models/cleaned/cleaned_demos_app_in_prog_amendment_current_phase.sql
```diff
@@ -1,19 +1,25 @@
...
+    amendment.signature_level_id,
+    amendment.created_at,
+    amendment.updated_at,
+    amendment._legacy_mdcd_demo_amndmt_id,
+    amendment._legacy_mdcd_pendg_demo_id,
+    amendment._legacy_mdcd_demo_aplctn_id
 FROM
-    {{ ref('cleaned_demos_app_in_prog_amendment') }}
+    {{ ref('cleaned_demos_app_in_prog_amendment') }} AS amendment
+LEFT JOIN {{ ref('apps_in_prog_phase_completion') }} AS phase_completion
```
TARehman, line 21: Isn't this plus the WHERE statement just an INNER JOIN with extra steps? Relevant here because it will silently filter things out at this point, so if the intent was to shunt things into an error table that isn't happening.

TARehman (approved):
Looks okay to me on a quick review but there is one comment that should be addressed.

data/migration/stage_pmda_for_migration/models/cleaned/cleaned_demos_app_in_prog_amendment_current_phase.sql
connor-parke-icf: good catch. Issue also exists in the similar table for demonstrations. I updated this to remove the where clause and created a check for non-null current_phase_id on this table, then removed the where clause. Still passed with 0 errors, so it wasnt actually filtering anything, but good to have it explicitly caught.

PR #1668, opened 2026-08-06
Demos 2542 amendment documents

connor-parke-icf:
## Changes

-

## Automated review

- [ ] Run AI Code Review


data/migration/stage_pmda_for_migration/models/docs/docs_pmda_app_docs_with_application.sql
```diff
@@ -1,30 +1,15 @@
-WITH demos AS (
-    (
-        SELECT
-            id,
-            mdcd_pendg_demo_id AS _legacy_mdcd_pendg_demo_id
-        FROM {{ ref('cleaned_demos_app_demonstration_in_prog_demos') }}
```
connor-parke-icf, line 6: because we're not pulling these into phase status, we can vastly simplify the flow by just pulling from the final application table. The documents that do not have an application already are captured in an error errors_app_docs_missing_application

data/migration/stage_pmda_for_migration/tests/assert_application_docs_of_type_bn_workbook.sql
```diff
@@ -0,0 +1,4 @@
+{{ config(severity='warn') }}
```
TARehman, line 1: NB: Should add a comment here just describing what we're warning about (name essentially tells us but still).

TARehman (approved):
LGTM

PR #1670, opened 2026-08-07
Demos 2532 demo type

connor-parke-icf:
migrates demonstration types assignment and associated tags into demos. Also applies a migrated from pmda application tag. Works similarly to how PMDA has a massive view unioning a bunch of demo type tables together. 

Note, one DEMOS migration was needed, as we were populating tags in a previous migration. We should probably modify the seeder to create those as a followup 

## Automated review

- [x] Run AI Code Review


james-becwar-cms (approved):

data/migration/stage_pmda_for_migration/models/apps/app_pmda_demo_types.sql
```diff
@@ -0,0 +1,697 @@
+WITH demo_types_union AS (
```
TARehman, line 4: NB: It might be worth adding a comment to explain exactly why this is as cursed as it is.

data/migration/stage_pmda_for_migration/models/cleaned/cleaned_demos_app_tag.sql
```diff
@@ -1,19 +1,23 @@
...
     'Application' AS tag_type_id,
     'System' AS source_id,
     'Approved' AS status_id,
     current_timestamp AS created_at,
     current_timestamp AS updated_at
 FROM
-    {{ ref('cleaned_demos_app_tag_name_migrated_from_pmda') }}
+    {{ ref('cleaned_demos_app_demo_type_tag_assign') }}
+GROUP BY
+    tag_name_id
```
TARehman: If we're not actually GROUPING, we should probably use a DISTINCT here to be more clear about what we intend to do.

data/migration/stage_pmda_for_migration/models/cleaned/cleaned_demos_app_tag.sql
```diff
@@ -1,19 +1,23 @@
 SELECT
-    id AS tag_name_id,
+    tag_name_id,
     'Application' AS tag_type_id,
```
TARehman: NB: For future reference, I would do this slightly differently.

```sql
WITH tag_type_ids AS (
    SELECT 'Application' AS tag_type_id UNION ALL
    SELECT 'Demonstration Type' AS tag_type_id
)
SELECT
    assign.tag_name,
    tag_type_ids.tag_type_id,
    ...
FROM
    {{ ref('cleaned_demos_app_demo_type_tag_assign') }}
INNER JOIN
    tag_type_ids ON TRUE
```

By joining like this, you get one row for each row in `tag_type_ids`.

data/migration/stage_pmda_for_migration/models/cleaned/cleaned_demos_app_tag_name.sql
```diff
@@ -1,4 +1,6 @@
 SELECT
-    'Migrated From PMDA' AS id,
+    tag_name_id AS id,
     current_timestamp AS created_at,
     current_timestamp AS updated_at
+FROM {{ ref('cleaned_demos_app_tag') }}
+GROUP BY id
```
TARehman: This GROUP BY definitely feels out of place.

data/migration/stage_pmda_for_migration/models/errors/error_new_demo_type_in_prog_not_from_other.sql
```diff
@@ -0,0 +1,8 @@
+SELECT pmda_tag_name.*
```
TARehman: I think rewriting these to use NOT IN or NOT EXISTS would be a lot more clear as to what is being checked (specifically, that all the non-other tags appear both in DEMOS and PMDA identically).

Unrelated, but its probably important to check that these are unique keys on both sides if you're doing a JOIN. We know it's good in DEMOS, but we could add a models.yml test here for the intermediate artifact.

data/migration/stage_pmda_for_migration/tests/assert_no_new_demo_type_not_from_other.sql
```diff
@@ -0,0 +1,3 @@
+SELECT * FROM {{ ref('error_new_demo_type_in_prog_not_from_other') }}
```
TARehman: A comment here explaining why we're doing these tests would be useful I think. It's not very clear from the file name.

data/migration/stage_pmda_for_migration/models/errors/errors_demo_type_effective_expiration_dates_pending.sql
TARehman, line 1: These models should be named `errors_` not `error_` to be consistent with the rest of the codebase.

TARehman, line 1: I don't see anywhere where you assert on there being records in these tables; that is part of the way we use the error tables. Filter, then have a test that asserts whatever with a warning, so that the filtering is surfaced in the logs.

data/migration/stage_pmda_for_migration/models/models.yml
```diff
@@ -294,22 +294,89 @@ models:
...
+          - relationships:
+              name: assert_application_id_exists_in_final_demos_app_application
+              arguments:
+                to: ref('final_demos_app_application')
+                field: id
+      - name: tag_name_id
+        data_type: text
+        data_tests:
+          - not_null:
+              name: assert_no_missing_tag_name_id_in_final_demos_app_application_tag_assignment
```
TARehman, line 313: Does this actually work correctly? I thought names had to be less than 64 characters.

data/migration/stage_pmda_for_migration/models/errors/errors_pending_demonstration_types_with_no_in_prog_demo.sql
TARehman, line 1: The file names here seem a little inconsistent; sometimes we're using demo_type and other times demonstration_types. It might help to organize them all with the same naming so they are more clear in the file explorer.

TARehman (commented):
I think we should probably clean up a few things here before we merge, noted throughout.

data/migration/stage_pmda_for_migration/models/models.yml
connor-parke-icf: validated that it does work by selecting a failing scenario
```
SELECT
    id AS application_id,
    'Some othing thing' AS tag_name_id,
    'Application' AS tag_type_id
FROM {{ ref('final_demos_app_application') }}
```
```
 Failure in test assert_application_tag_assignment_tag_name_id_exists_in_cleaned_demos_app_tag_name (models/models.yml)
13:26:53    Got 230 results, configured to fail if != 0
13:26:53  
13:26:53    compiled code at target/compiled/stage_pmda_for_migration/models/models.yml/assert_application_tag_assignment_tag_name_id_exists_in_cleaned_demos_app_tag_name.sql
```

data/migration/stage_pmda_for_migration/models/apps/app_pmda_demo_types.sql
connor-parke-icf: added an explanation

data/migration/stage_pmda_for_migration/models/cleaned/cleaned_demos_app_tag.sql
connor-parke-icf: changed to be selecting distinct names in a CTE then joining in the main select.

data/migration/stage_pmda_for_migration/models/cleaned/cleaned_demos_app_tag.sql
connor-parke-icf: followed this pattern since this was changing from the grouping to a distinct pattern anyway

data/migration/stage_pmda_for_migration/models/cleaned/cleaned_demos_app_tag_name.sql
connor-parke-icf: replaced with a CTE with a DISTINCT clause

data/migration/stage_pmda_for_migration/models/errors/error_new_demo_type_in_prog_not_from_other.sql
connor-parke-icf: reworked this to utilized `WHERE name NOT IN (SELECT name FROM demos)` instead of a left join + null filter. Reads much better now

data/migration/stage_pmda_for_migration/tests/assert_no_new_demo_type_not_from_other.sql
connor-parke-icf: added a corresponding comment

TARehman (approved):
LGTM thanks for addressing those other comments 👍🏻

PR #1684, opened 2026-08-10
DEMOS-2594: updated UIPath for multiple values

dustbuster:
This breaks down the multiple values smushed together into sepearate values. 

Screenshot 2026-08-10 at 14 03 04

Screenshot 2026-08-10 at 16 44 44

## Summary

Instead of JOINiNG the multiple values. We do not. I was prepared for multiple values, but they never happened until this latest model. 

## Changes
Improved existing mechanism for creating multiple rows. This was like 1/2 in place, this is hte completion of that work. 

## Risk
Very little. This was giving the multiple results as 1 string and now we are fixing it. I would say broken is probably better than working incorrectly. And working properly is better than it not working. 

## Automated review

- [x] Run AI Code Review

### DIFF
--- a/lambdas/UIPath/db/uiPathResults.ts
+++ b/lambdas/UIPath/db/uiPathResults.ts
@@ -3,6 +3,7 @@ import type { PoolClient } from "pg";
 import { getDbPool } from "../db";
 import { ExtractionStatus } from "../fetchExtractResult";
 import {
+  DEMO_TYPE_FIELD_ID,
   getConfidence,
   getExtractedFields,
   getTextStartIndex,
@@ -33,6 +34,7 @@ const INSERT_VALUE_SQL = `insert into ${DEMOS_SCHEMA}.uipath_value
   (id, uipath_result_id, document_id, application_id, field_id, value, text_length, text_start_index, confidence, token_list, updated_at)
  values
   ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10::jsonb, now())`;
+const SELECT_TAG_NAMES_SQL = `select id from ${DEMOS_SCHEMA}.tag_name`;
 
 export type PersistedUiPathExtraction = {
   extractedFieldCount: number;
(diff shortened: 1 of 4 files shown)


lambdas/UIPath/db/uiPathResults.ts
```diff
@@ -33,6 +34,7 @@ const INSERT_VALUE_SQL = `insert into ${DEMOS_SCHEMA}.uipath_value
   (id, uipath_result_id, document_id, application_id, field_id, value, text_length, text_start_index, confidence, token_list, updated_at)
  values
   ($1, $2, $3, $4, $5, $6, $7, $8, $9, $10::jsonb, now())`;
+const SELECT_TAG_NAMES_SQL = `select id from ${DEMOS_SCHEMA}.tag_name`;
```
connor-parke-icf: if we are looking for the names of available Application Tags, we should select from `tag where tag_type_id = "Application"` since selecting from tag_name will give us tags that may not be applicable as application tags, like Reference tags or Demonstration Type tags

lambdas/UIPath/uipathFields.ts
```diff
@@ -74,75 +74,147 @@ export function getTokenList(value: UiPathFieldValue): unknown[] {
...
+  const addAlias = (key: string, tagName: string) => {
+    const candidates = aliasesByKey.get(key) ?? new Set<string>();
+    candidates.add(tagName);
+    aliasesByKey.set(key, candidates);
+  };
+
+  for (const tagName of tagNames) {
+    exactNames.set(normalizeTagLookupKey(tagName), tagName);
+
+    for (const match of tagName.matchAll(/\(([^()]+)\)/g)) {
```
connor-parke-icf, line 117: im a little out of context for the difference between the tag name returned by uipath and the aliases/regex pattern to them.  Does the uipath return things that arent included in a set of existing application tag names?

dustbuster: That regex just extracts things from parens.

So i'm kinda at the mercy of the model. So whatever the model returns. I gotta make sure it's sane. And then match it. part of that is nuking the parens and seeing if HSRN is a stand alone tag. so we are checking the full value returned e.g. "Health Related Social Need (HSRN)". And then comparing the stuff in the parens "(HSRN)"
And seeing if we get a hit.

lambdas/UIPath/db/uiPathResults.ts
dustbuster: This is a good catch. Thank you sir!

dustbuster: I am going to test this change locally and make sure it's still giving the right results.

james-becwar-cms (approved):

lambdas/UIPath/db/uiPathResults.ts
dustbuster: Addressed!

lambdas/UIPath/uipathFields.ts
dustbuster: UiPath returned:
(LTSS)

But the database contains:
Long-Term Services and Supports (LTSS)

This was able to find and match to DB

PR #1698, opened 2026-08-12
Fix: client snyk findings

tresdonicf:
## Summary

- Addresses the snyk findings in the client codebase
- Ran `npm audit fix`
- Removed stale deps with `npm remove @fakerjs/faker`
- Removed snyk ignores

PR #1699, opened 2026-08-12
Demos 2605 auto refresh types tab

connor-parke-icf:
This resolves two related issues under https://jiraent.cms.gov/browse/DEMOS-2605. 

1) when making changes to deliverables the user needs to refresh to see the updates: this was occurring because we had the wrong query getting refetched after creating a new deliverable or deleting a deliverable. For creating a new one, since this functionality only exists on the demonstration deliverable page, it was simple enough to just swap the referenced query with the correct one. For removing a deliverable however, since this is called from both the main deliverables page and the demonstration deliverables tab, we need to pass in the correct refetch. We _couldve_ potentially added both to the refetch, but because the larger deliverable query fetched _all_ the deliverables and would take a long time to refetch. So added the refetch as a prop

2) To resolve the demonstration types issues was a little more complicated, but it turned out to be a caching conflict between the demonstration fetched by the deliverables page and the demonstration fetched at the top of the demo details. basically, when we were fetching the demonstration on the details page, it would be cached with its typical id:typename. But on the demonstration deliverable tab, we werent fetching the ID of the demonstration, so it would be cached as the name of the query + its params. This would conflict with the other cache key, and the result would be an empty object with just the typename field. All this was resolved simply by including the demonstration ID in the fetch results on the demonstration deliverable listing query. When that was present, it got properly normalized and all the operations on the demonstration types tab properly updated on submit. 

### In addition
resolved a snyk finding by removing the package that was causing it. we werent using it. this change parallels the change in https://github.com/Enterprise-CMCS/demos/pull/1698  

## Automated review

- [x] Run AI Code Review

### DIFF
--- a/client/src/components/dialog/deliverable/RemoveDeliverableDialog.tsx
+++ b/client/src/components/dialog/deliverable/RemoveDeliverableDialog.tsx
@@ -1,5 +1,5 @@
 import React, { useState } from "react";
-import { DocumentNode, gql, useMutation } from "@apollo/client";
+import { gql, useMutation } from "@apollo/client";
 import { ErrorButton } from "components/button";
 import { BaseDialog } from "components/dialog/BaseDialog";
 import { useToast } from "components/toast";
@@ -27,13 +27,12 @@ export const getRemoveDeliverableConfirmMessage = (deliverableCount: number): st
 export const RemoveDeliverableDialog: React.FC<{
   deliverableIds: string[];
   onClose: () => void;
-  refetchQueries: DocumentNode[];
   onDeleted?: () => void;
-}> = ({ deliverableIds, onClose, onDeleted, refetchQueries }) => {
+}> = ({ deliverableIds, onClose, onDeleted }) => {
   const { showSuccess, showError } = useToast();
   const [isDeleting, setIsDeleting] = useState(false);
   const [deleteDeliverable] = useMutation(DELETE_DELIVERABLE_MUTATION, {
-    refetchQueries,
+    refetchQueries: "active",
     awaitRefetchQueries: true,
   });
 
--- a/client/src/components/dialog/deliverable/RemoveDeliverableDialog.tsx
+++ b/client/src/components/dialog/deliverable/RemoveDeliverableDialog.tsx
@@ -1,5 +1,5 @@
 import React, { useState } from "react";
-import { gql, useMutation } from "@apollo/client";
+import { DocumentNode, gql, useMutation } from "@apollo/client";
 import { ErrorButton } from "components/button";
 import { BaseDialog } from "components/dialog/BaseDialog";
 import { useToast } from "components/toast";
@@ -27,12 +27,13 @@ export const getRemoveDeliverableConfirmMessage = (deliverableCount: number): st
 export const RemoveDeliverableDialog: React.FC<{
   deliverableIds: string[];
   onClose: () => void;
+  refetchQueries: DocumentNode[];
   onDeleted?: () => void;
-}> = ({ deliverableIds, onClose, onDeleted }) => {
+}> = ({ deliverableIds, onClose, onDeleted, refetchQueries }) => {
   const { showSuccess, showError } = useToast();
   const [isDeleting, setIsDeleting] = useState(false);
   const [deleteDeliverable] = useMutation(DELETE_DELIVERABLE_MUTATION, {
-    refetchQueries: ["GetDeliverablesPage"],
+    refetchQueries,
     awaitRefetchQueries: true,
   });
 
--- a/client/src/components/table/tables/DeliverableActionButtons.tsx
+++ b/client/src/components/table/tables/DeliverableActionButtons.tsx
@@ -9,7 +9,6 @@ import { EditIcon } from "components/icons/Navigation/EditIcon";
 
 import { selectionTooltip } from "./actionTooltips";
 import type { FormattedDeliverableTableRow } from "./DeliverableTable";
-import { DocumentNode } from "@apollo/client";
 
 export const DELIVERABLE_CANT_DELETE_HAS_FILES =
   "Only Upcoming/Past Due \ndeliverables w/o comments \nor files can be deleted.";
@@ -44,8 +43,7 @@ const getDeleteTooltip = (selectedCount: number, selectedHasFilesOrComments: boo
  */
 export const DeliverableActionButtons: React.FC<{
   table: TanstackTable<FormattedDeliverableTableRow>;
-  refetchQueries: DocumentNode[];
-}> = ({ table, refetchQueries }) => {
+}> = ({ table }) => {
   const { showEditDeliverableDialog, showRemoveDeliverableDialog } = useDialog();
   const selectedRows = table.getSelectedRowModel().rows;
   const selectedCount = selectedRows.length;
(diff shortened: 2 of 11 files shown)


client/src/components/dialog/deliverable/RemoveDeliverableDialog.tsx
```diff
@@ -27,12 +27,13 @@ export const getRemoveDeliverableConfirmMessage = (deliverableCount: number): st
...
   onClose: () => void;
+  refetchQueries: DocumentNode[];
   onDeleted?: () => void;
-}> = ({ deliverableIds, onClose, onDeleted }) => {
+}> = ({ deliverableIds, onClose, onDeleted, refetchQueries }) => {
   const { showSuccess, showError } = useToast();
   const [isDeleting, setIsDeleting] = useState(false);
   const [deleteDeliverable] = useMutation(DELETE_DELIVERABLE_MUTATION, {
-    refetchQueries: ["GetDeliverablesPage"],
+    refetchQueries,
```
connor-parke-icf: because this dialog is used on the main deliverable listing page as well as the demonstration deliverables tab, the refetch had to be parameterized since the query on the main deliverable listing page would take a long time to resolve. so calls to the removeDeliverableDialog supply it with the query to refetch

client/src/pages/DemonstrationDetail/deliverables/DeliverablesTab.tsx
```diff
@@ -14,8 +14,9 @@ import { useNavigate } from "react-router-dom";
 export const ADD_DELIVERABLE_SLOT_BUTTON_NAME = "button-add-deliverable-slot";
 
 export const DEMONSTRATION_DELIVERABLE_TAB = gql`
-  query GetDeliverablesPage($id: ID!) {
+  query GetDemonstrationDeliverables($id: ID!) {
     demonstration(id: $id) {
+      id
```
connor-parke-icf, line 19: simply including the id here resolved the issues on the demo types tab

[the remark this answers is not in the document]
connor-parke-icf: this is fine as the variables used in this refetch will be the same as its initial fetch, so theres no staleness to worry about

client/src/components/table/tables/DeliverableActionButtons.tsx
```diff
@@ -43,7 +44,8 @@ const getDeleteTooltip = (selectedCount: number, selectedHasFilesOrComments: boo
  */
 export const DeliverableActionButtons: React.FC<{
   table: TanstackTable<FormattedDeliverableTableRow>;
-}> = ({ table }) => {
+  refetchQueries: DocumentNode[];
```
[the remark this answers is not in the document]
connor-parke-icf: this is fine for the above issue, we dont need to pass in variables and this is intended only to be used in the places that were affected by this PR, not generically.

client/src/pages/DemonstrationDetail/deliverables/DeliverablesTab.tsx
```diff
@@ -14,8 +14,9 @@ import { useNavigate } from "react-router-dom";
 export const ADD_DELIVERABLE_SLOT_BUTTON_NAME = "button-add-deliverable-slot";
 
 export const DEMONSTRATION_DELIVERABLE_TAB = gql`
-  query GetDeliverablesPage($id: ID!) {
+  query GetDemonstrationDeliverables($id: ID!) {
```
tresdonicf, line 17: Here it's not a bad idea to `export const DEMONSTRATION_DELIVERABLE_QUERY_NAME` to be used elsewhere.

client/src/components/dialog/deliverable/RemoveDeliverableDialog.tsx
```diff
@@ -27,12 +27,13 @@ export const getRemoveDeliverableConfirmMessage = (deliverableCount: number): st
...
   deliverableIds: string[];
   onClose: () => void;
+  refetchQueries: DocumentNode[];
   onDeleted?: () => void;
-}> = ({ deliverableIds, onClose, onDeleted }) => {
+}> = ({ deliverableIds, onClose, onDeleted, refetchQueries }) => {
   const { showSuccess, showError } = useToast();
   const [isDeleting, setIsDeleting] = useState(false);
   const [deleteDeliverable] = useMutation(DELETE_DELIVERABLE_MUTATION, {
-    refetchQueries: ["GetDeliverablesPage"],
```
tresdonicf, line 35: This has been an annoying reality of the codebase and I see AI agents fill in these gaps a lot without note that they might need to be made more robust. In some places this has been replaced with a `..._QUERY_NAME` parameter which is quite a bit better if we're trying to target it.

client/src/components/dialog/deliverable/RemoveDeliverableDialog.tsx
```diff
@@ -27,12 +27,13 @@ export const getRemoveDeliverableConfirmMessage = (deliverableCount: number): st
 export const RemoveDeliverableDialog: React.FC<{
   deliverableIds: string[];
   onClose: () => void;
+  refetchQueries: DocumentNode[];
   onDeleted?: () => void;
-}> = ({ deliverableIds, onClose, onDeleted }) => {
+}> = ({ deliverableIds, onClose, onDeleted, refetchQueries }) => {
```
tresdonicf: Overall this is okay to pass in to the prop but a little less than ideal. It's more ideal to just update the list used in `useMutation` below with all the possible things that can be refetched as apollo will just discard the ones that are not in-use at the time. This way all need to do is update this one spot rather than pollute the calling components to pass this in as well which expands the infra needed for those.

alternatively even we've talked about it a lot but `refetchQueries: "active"` is a one-line fix here I believe

tresdonicf (commented):
Left some comments, I'd like to see this change a little more self-contained rather than spilling out into other components, either through using the "active" keyword to refetch or importing the symbols of what's needed to be refreshed at the site of the `useMutation`

client/src/components/dialog/deliverable/RemoveDeliverableDialog.tsx
tresdonicf: For inspiration take this section of code from the concept phase where we refetch it for all of "amendments / extensions / demonstrations" and apollo just throws away the ones that aren't being used but it means we don't have to parameterize call-sites of the individual functions, keeping the code cleaner:

```
  const skipConceptPhase = async (applicationId: string) => {
    return await mutate({
      variables: { applicationId },
      refetchQueries: [
        GET_WORKFLOW_DEMONSTRATION_QUERY,
        GET_AMENDMENT_WORKFLOW_QUERY,
        GET_EXTENSION_WORKFLOW_QUERY,
      ],
    });
  };
  ```

connor-parke-icf: I would much rather refetch active queries than specify queries that arent used, since the latter produces console warnings/errors.

While passing around the refetch queries as a prop is not ideal, we are already running up against performance concerns with over-fetching already. On the deliverable listing page, the initial query takes around 5 seconds on prod currently, and any refetch of that query repeats the long load. Given that PMDA goes dark today, im willing to refetch "active" as a hotfix, but it is a pattern i still want to lean away from.

client/src/components/dialog/deliverable/RemoveDeliverableDialog.tsx
connor-parke-icf: ideally, i would rather reference queries via their DocumentNode definition or more preferably, a TypedDocumentNode definition.  Least of all of course, the plain string name

tresdonicf (approved):

PR #1703, opened 2026-08-13
Demos 1900: EMAILER

dustbuster:
NOT READY FOR MERGE!! 
This was based off post-laucnh, but i back merged main into it and made main my target branch. 

This will be a bit loaded until post-launch gets merged. 

## Summary

Added EMAILS. 

## Changes

Basically added a template engine and improved the emailer lambda. 
And added mailpit for local. 
-

https://github.com/user-attachments/assets/f56205dd-b265-4d5e-822d-34bad1811672

Screenshot 2026-08-13 at 12 37 10

Medium risk. 

## Automated review

- [ ] Run AI Code Review


lambdas/emailer/emailTemplates/parts/MultipleDeliverablesEmail.tsx
```diff
@@ -0,0 +1,35 @@
...
+    <EmailLayout>
+      <Text style={textStyle}>Hello,</Text>
+      <Text style={textStyle}>
+        You have been assigned new {deliverableType} deliverables for your Demonstration. View these deliverables and
+        any required next steps in the DEMOS system:
+      </Text>
+      {deliverables.map(({ dueDate, link }) => (
+        <Text key={link} style={textStyle}>
+          <Link href={link}>{link}</Link> due on {dueDate}
+        </Text>
```
dustbuster, line 24: This is a foreach so if we add more than 4 at a time deliverables we'll be GTG.

client/src/components/dialog/deliverable/AddDeliverableSlotDialog.tsx
```diff
@@ -142,7 +142,7 @@ export const AddDeliverableSlotDialog = ({
   demonstration: AddDeliverableSlotDemonstration;
 }) => {
   const { showSuccess } = useToast();
-  const { createDeliverables, loading } = useCreateDeliverable();
+  const { createDeliverables, loading } = useCreateDeliverables();
```
dustbuster, line 146: All these changes were made to support being able to send 1 email for multiple deliverables. 

Previously, the dialog called `createDeliverable` once per item with Promise.all, which generated separate notifications. The change made one createDeliverables mutation so the server could create the batch and send one grouped email.

Otherwise if we made 4 deliverables (or more) at once. then it would just send off 4 emails. This bulks them together.

client/src/components/dialog/referenceAgreement/ReferenceAgreementDialog.tsx
```diff
@@ -67,13 +72,21 @@ export const ReferenceAgreementDialog = ({
...
-          <span className="text-sm text-text-font">I accept the terms</span>
+          <span className="text-sm text-text-font">&nbsp;I accept the terms</span>
+        </label>
+        <label className={STYLES.checkbox}>
+          <Checkbox
+            name="checkbox-email-agreement"
+            checked={emailAgreement}
+            onChange={() => setEmailAgreement((prev) => !prev)}
+          />
+          <span className="text-sm text-text-font">{EMAIL_AGREEMENT_LABEL}</span>
```
dustbuster, line 89: Reasons for changes: 

- Adds the second “Receive an email…” checkbox.
- Tracks whether the user opted in.
- Passes emailAgreement: true/false into downloadReference.
- Renames the checkbox style because it is shared by both checkboxes.

dustbuster: OG ticket: [DEMOS-1542](https://jiraent.cms.gov/browse/DEMOS-1542)

client/src/components/dialog/AttestationDialog.tsx
```diff
@@ -15,7 +15,7 @@ const ATTESTATION_STATEMENT =
   "By uploading this Budget Neutrality (BN) notebook, I attest the information included " +
   "with this submission is true and accurate to the best of my knowledge.";
 
-const ACKNOWLEDGEMENT_LABEL = "I acknowledge and attest to the statement above.";
+const ACKNOWLEDGEMENT_LABEL = "\u00A0 I acknowledge and attest to the statement above.";
```
dustbuster, line 18: Added this little gap. 
between checkbox and words. `[ ] I accept`
Image

server/src/model/deliverable/deliverableSchema.ts
```diff
@@ -90,6 +90,8 @@ export const deliverableSchema = gql`
   type Mutation {
     createDeliverable(input: CreateDeliverableInput!): Deliverable!
       @auth(requires: ["Perform CMS Action"])
+    createDeliverables(inputs: [CreateDeliverableInput!]!): [Deliverable!]!
+      @auth(requires: ["Perform CMS Action"])
```
connor-parke-icf, line 94: not sure we should be opening up a new mutation for deliverables in this PR.  Not wholly against opening up a separate mutation for when the create deliverable dialog creates multiple deliverables, but i think we should have that in a concentrated effort.  We should probably transactionalize the mutation if we're adding it

server/src/model/deliverable/approveDeliverableExtension.ts
```diff
@@ -84,6 +86,22 @@ export async function approveDeliverableExtension(
...
     );
-    return approvedDeliverable;
+    return {
+      approvedDeliverable,
+      sourceActionId: action.id,
+      previousDueDate: unapprovedDeliverable.dueDate,
+    };
   });
+
+  if (options.sendEmailNotifications !== false) {
```
connor-parke-icf, line 96: if options.sendEmailNotifications is undefined, it will send an email. That may be the intent, but the behavior isnt super clear

server/src/model/deliverable/approveDeliverableExtension.ts
```diff
@@ -15,19 +15,21 @@ import {
...
 } from "../deliverableExtension/queries";
 import { checkOptionalNotNullFields } from "../../errors/checkOptionalNotNullFields";
+import { dispatchExtensionDecisionMadeEmail } from "../email";
 
 export async function approveDeliverableExtension(
   deliverableId: string,
   input: ApproveDeliverableExtensionInput,
-  context: GraphQLContext
+  context: GraphQLContext,
+  options: { sendEmailNotifications?: boolean } = {}
```
connor-parke-icf, line 24: i dont see a corresponding change where these additional params are invoked.  Im a little hesitant of putting the decision to email or not at the resolver level as my instinct is to move all business logic out of resolvers as much as possible, leaving them to be a thin routing layer with maybe some field mapping

server/src/model/deliverable/completeDeliverable.ts
```diff
@@ -9,14 +9,20 @@ import {
...
+type CompleteDeliverableOptions = {
+  sendEmailNotifications?: boolean;
+};
 
 export async function completeDeliverable(
   deliverableId: string,
   finalStatus: FinalDeliverableStatus,
-  context: GraphQLContext
+  context: GraphQLContext,
+  options: CompleteDeliverableOptions = {}
```
connor-parke-icf, line 22: the only non-test call i see invoking this function with the options is sound in the seeder.  we may want to separate out the framework parts of the email sender from the instances where emails are being sent into separate PRs

server/src/model/deliverable/completeDeliverable.ts
```diff
@@ -46,6 +52,20 @@ export async function completeDeliverable(
...
     );
 
-    return completedDeliverable;
+    return {
+      completedDeliverable,
+      sourceActionId: action.id,
+    };
   });
+
+  if (options.sendEmailNotifications !== false) {
```
connor-parke-icf, line 61: same here, if options.sendEmailNotifications is undefined or null, it will send by default

server/src/model/deliverable/createDeliverable.ts
```diff
@@ -46,7 +49,19 @@ export async function createDeliverable(
...
 
-    return newDeliverable;
+    return {
+      newDeliverable,
+      sourceActionId: action.id,
+    };
   });
-  return createdDeliverable;
+
+  if (options.sendEmailNotifications !== false) {
```
connor-parke-icf, line 58: We may want to include the email sending within any transactions so that if the email fails to queue the database changes are rolled back.

server/src/model/deliverable/createDeliverables.ts
```diff
@@ -0,0 +1,42 @@
...
+  if (
+    inputs.some(
+      (input) =>
+        input.demonstrationId !== firstInput.demonstrationId ||
+        input.deliverableType !== firstInput.deliverableType ||
+        input.cmsOwnerUserId !== firstInput.cmsOwnerUserId
+    )
+  ) {
+    throw new Error("Multiple deliverables must have the same demonstration, deliverable type, and CMS owner.");
+  }
```
connor-parke-icf, line 30: this check makes sense to me. Theres still a larger conversation about how we want to handle createMany operations in the API in a general sense. It seems the purpose here is to package the single submit button click into one action that sends one email, which makes sense to me.  Not sure if that essentially replaces the createDeliverable mutation in general

server/src/model/deliverable/createDeliverables.ts
```diff
@@ -0,0 +1,42 @@
...
+export async function createDeliverables(
+  inputs: CreateDeliverableInput[],
+  context: GraphQLContext
+): Promise<Deliverable[]> {
+  if (inputs.length === 0) {
+    throw new Error("Cannot create deliverables without at least one input.");
+  }
+
+  if (inputs.length === 1) {
+    return [await createDeliverable(inputs[0], context)];
```
connor-parke-icf, line 17: if it only creates one deliverable, it seems to skip any email sending. is that intentional?

server/src/model/deliverable/denyDeliverableExtension.ts
```diff
@@ -56,6 +58,18 @@ export async function denyDeliverableExtension(
       },
       tx
     );
-    return deliverable;
+    return { deliverable, sourceActionId: action.id };
   });
+
+  if (options.sendEmailNotifications !== false) {
+    await dispatchExtensionDecisionMadeEmail({
```
connor-parke-icf, line 65: again, maybe this should be included in the same transaction as the rest of the changes. Also the strict equal to false may have unintended side effects since the param is optional

server/src/model/deliverable/updateDeliverable.ts
```diff
@@ -46,8 +52,27 @@ export async function updateDeliverable(
...
   });
+
+  if (dueDateUpdate && options.sendEmailNotifications !== false) {
+    await dispatchDeliverableDueDateUpdatedEmail({
+      deliverableId,
+      previousDueDate: dueDateUpdate.previousDueDate,
+      sourceActionId: dueDateUpdate.sourceActionId,
+      triggeredByUserId: context.user.id,
+    });
+  }
```
connor-parke-icf, line 75: if the only email we need to dispatch from this mutation comes from the manuallyUpdateDeliverableDueDate result, we might should put the dispatch logic in that function instead

server/src/model/email/createTestEmail.ts
```diff
@@ -0,0 +1,30 @@
+import { EmailAddressResolver } from "graphql-scalars";
```
connor-parke-icf, line 1: interesting, i had completely forgotten about these scalars. Might be a good idea to change the type of our emails in the schema to be the EmailAddress from here. Total aside though, separate from this PR

server/src/model/email/createTestEmail.ts
```diff
@@ -0,0 +1,30 @@
...
+  const domain = email.slice(email.lastIndexOf("@") + 1).toLowerCase();
+  if (!TEST_EMAIL_DOMAINS.includes(domain)) {
+    throw new Error();
+  }
+}
+
+export async function createTestEmail(recipientEmail: string): Promise<string> {
+  const normalizedRecipientEmail = recipientEmail.trim();
+  try {
+    EmailAddressResolver.parseValue(normalizedRecipientEmail);
```
connor-parke-icf, line 19: does parseValue return a parsed email or just throw an error if its not parseable? Id think we should pass the result of this into verifyAdminEmail, so its clear that it passed this initial email syntax validation

server/src/model/email/deliverableEmail.ts
```diff
@@ -0,0 +1,752 @@
...
+  personId: string;
+  name: string;
+  address: string;
+};
+
+export async function dispatchDeliverableCreatedEmail(
+  input: DeliverableEmailDispatchInput
+): Promise<void> {
+  return dispatchDeliverableEmail(input, "Deliverable Created", enqueueDeliverableCreatedEmail);
+}
```
connor-parke-icf, line 56: might be a way to combine these types using Generics, but not sure if would be a boon to clarity or not. Though, since the majority of these types are not exported, we might should just inline them in the function definitions to keep the requred props and the function that requires them close to eachother

server/src/model/email/deliverableEmail.ts
```diff
@@ -0,0 +1,752 @@
...
+export async function dispatchDeliverableSubmittedEmail(
+  input: DeliverableEmailDispatchInput
+): Promise<void> {
+  return dispatchDeliverableEmail(
+    input,
+    "Deliverable Submitted",
+    enqueueDeliverableSubmittedEmail
+  );
+}
+
```
connor-parke-icf, line 92: having a number of passthrough functions like this makes me think that we may have an extra layer. We might can combine the "dispatch" functions with the "enqueue" functions since it appears that most of the time the dispatch functions are just calling the enqueue functions. Might be an aspect to this im not thinking about though

server/src/model/email/emailNotification.ts
```diff
@@ -0,0 +1,83 @@
...
+  personId?: string;
+  emailAddress: string;
+};
+
+export async function enqueueTrackedRealtimeEmail(
+  message: RealtimeEmailEnvelope,
+  sourceActionId: string | undefined,
+  recipients: EmailNotificationRecipient[]
+): Promise<string> {
+  const notification = await prisma().emailNotification.create({
```
connor-parke-icf, line 19: we probably should have this included in the same transaction of the thing thats calling it is in. to keep the email queuing atomic with the rest of the operations

server/src/model/email/emailNotification.ts
```diff
@@ -0,0 +1,83 @@
...
+  });
+  await prisma().emailNotification.updateMany({
+    where: {
+      id: notification.id,
+      statusId: "Pending",
+    },
+    data: {
+      statusId: "Queued",
+    },
+  });
```
connor-parke-icf, line 80: this feels a little off to me. We are creating an email notification, adding it to the sqs queue, updating its record with the messageId... but then setting all the pending email notifications with that notificationId to being queued? This smells like a workaround with prisma's update functions requiring a selection by unique key. Usually this indicates that something is up with how we've defined the schema models

server/src/model/email/referenceEmail.ts
```diff
@@ -0,0 +1,94 @@
...
+  try {
+    const user = await prisma().user.findUniqueOrThrow({
+      where: { id: input.triggeredByUserId },
+      include: { person: true },
+    });
+    const emailAddress = user.person.email.trim();
+    if (!emailAddress) {
+      throw new Error(
+        `Cannot dispatch Terms And Conditions Requested email for user ${input.triggeredByUserId}: ` +
+          "the registered email address is blank.",
```
connor-parke-icf, line 31: could prob combine these into the same template string instead of template string + string literal.

server/src/services/emailQueue.ts
```diff
@@ -0,0 +1,152 @@
...
+    throw new Error("Failed to enqueue email message.");
+  }
+
+  log.info(
+    {
+      messageId: response.MessageId,
+      ...messageContext,
+    },
+    "Email queued"
+  );
```
connor-parke-icf, line 149: not sure how much it matters, but we should be careful about logging any PII in these logs

server/src/model/email/referenceEmail.ts
```diff
@@ -0,0 +1,94 @@
...
+        },
+        referenceMaterial: {
+          id: input.referenceId,
+          name: input.referenceName,
+        },
+        termsAndConditions: {
+          id: input.referenceAgreementId,
+          name: input.referenceAgreementName,
+          fileName: attachmentFileName,
+          s3Path: input.referenceAgreementS3Path,
```
connor-parke-icf, line 65: Hmm.. im guessing this is handled somehow in the lambda, but i hadnt considered the implications of a link to a document in an email. Im guessing it builds a link and attaches it to the email in some way, but i have no idea how that might interact with the lifetime of these links. some broader considerations there

server/src/model/emailNotificationRecipient/emailNotificationRecipient.prisma
```diff
@@ -0,0 +1,17 @@
...
+  normalizedEmail     String   @map("normalized_email")
+  createdAt           DateTime @default(now()) @map("created_at") @db.Timestamptz
+
+  emailNotification EmailNotification @relation(fields: [emailNotificationId], references: [id], onDelete: Restrict)
+  person            Person?           @relation(fields: [personId], references: [id], onDelete: Restrict)
+
+  @@id([id])
+  @@unique([emailNotificationId, normalizedEmail])
+  @@map("email_notification_recipient")
+}
```
connor-parke-icf, line 17: a little concerned with this table. It seems to be acting like an action log, such that it records the email recipients such that if a user's email changes, the old email is still tracked. This makes sense to me, but then we also have the optional column of personId, meaning an email can exist that wasnt to a registered user in DEMOS. My guess is that this accomodates a generic email sending, where the person doesnt exist in demos (in the case of the test email send), but we have the concept of the non-user-contact for that reason. We maybe should enforce that the email goes to someone we are aware of in the system, and make the admin's test email sending only be able to be sent to a valid Person instead of a generic email address

server/src/model/migrations/20260814160000_add_reference_email_entity/migration.sql
```diff
@@ -0,0 +1,5 @@
+INSERT INTO demos_app.email_notification_entity_type (id)
+VALUES ('reference');
+
+INSERT INTO demos_app.email_notification_type_entity_type (email_type_id, entity_type_id)
+VALUES ('Terms And Conditions Requested', 'reference');
```
connor-parke-icf, line 5: we should consider moving the rest of the inserts form the previous migration into this one, to separate out the Prisma/schema changes from the value initialization. We should be able to do that here, since these are new tables and dont have existing data we need to dance around

server/src/model/publicComment/createPublicComment.ts
```diff
@@ -4,21 +4,47 @@ import { GraphQLContext } from "../../auth";
 import { prisma } from "../../prismaClient";
 import { validateUserPermittedToMakePublicComment } from ".";
 import { insertPublicComment } from "./queries";
+import { dispatchPublicCommentAddedEmail } from "../email";
+
+const COMMENT_NOTIFICATION_DELIVERABLE_STATUSES = [
+  "Accepted",
+  "Approved",
+  "Received and Filed",
```
connor-parke-icf, line 12: is the intent here to only send emails when comments are added to deliverables in a finalized status? We _might_ have a const already for those statuses. but also, this might be something we should also handle with a DB constraint, cause its would still be possible to create a notification for a comment in those statuses in the db even if the API tries to block it.

server/src/model/publicComment/createPublicComment.ts
```diff
@@ -4,21 +4,47 @@ import { GraphQLContext } from "../../auth";
...
+    COMMENT_NOTIFICATION_DELIVERABLE_STATUSES.includes(
+      deliverableStatus as (typeof COMMENT_NOTIFICATION_DELIVERABLE_STATUSES)[number]
+    )
+  ) {
+    await dispatchPublicCommentAddedEmail({
+      deliverableId,
+      publicCommentId: publicComment.id,
+      triggeredByUserId: context.user.id,
+    });
+  }
```
connor-parke-icf, line 47: unlike the deliverable email sending, this one doesnt have a param to determine whether or not an email should be sent or not passed in from above. In general, i think i prefer it more towards the way you have it here, where we can determine whether the email should or shouldnt be set from the event itself instead of what may be passed in from the resolver. But we should decide on a pattern and be consistent with it

server/src/model/reference/getReferenceDownloadUrl.ts
```diff
@@ -1,33 +1,73 @@
...
 import { prisma } from "../../prismaClient";
+import { dispatchTermsAndConditionsRequestedEmail } from "../email";
 import { insertReferenceAgreementAcceptance } from "../referenceAgreementAcceptance/queries";
+import { POINT_AND_CLICK_AGREEMENT } from "./pointAndClickAgreement";
 import { validateReferenceDownloadRequest } from "./validateReferenceDownloadRequest";
 
 export async function getReferenceDownloadUrl(
   parent: unknown,
-  args: { id: string; acceptedAgreementId?: string },
+  args: { id: string; acceptedAgreementId?: string; emailAgreement?: boolean },
```
connor-parke-icf, line 11: im assuming there are corresponding frontend changes for this which insert the additional checkbox to the dialog. We should verify that the flow from PMDA is the flow we want in DEMOS

server/src/model/reference/pointAndClickAgreement.ts
```diff
@@ -0,0 +1,6 @@
+export const POINT_AND_CLICK_AGREEMENT = {
+  id: "national-measure-stewards-terms-and-conditions",
+  name: "National Measure Stewards Terms and Conditions",
+  fileName: "point-click-agreement.pdf",
+  s3Path: "static://point-click-agreement.pdf",
+} as const;
```
connor-parke-icf, line 6: kinda confused on where the static point and click agreement is coming from. each agreement should already be defined in the database and is referenced by the reference material we're fetching. i feel like we should not include this agreement and hardcoded data

server/src/model/reference/getReferenceDownloadUrl.ts
```diff
@@ -1,33 +1,73 @@
...
-        {
-          referenceId: validatedReferenceConfiguration.reference.id,
-          referenceAgreementId: validatedReferenceConfiguration.referenceAgreement!.id,
-          userId: context.user.id,
-        },
-        tx
+  }
+
+  const usesStaticAgreement =
+    args.acceptedAgreementId === POINT_AND_CLICK_AGREEMENT.id;
```
connor-parke-icf, line 21: what does a static agreement denote? i dont recognize that concept

server/src/model/reference/referenceSchema.ts
```diff
@@ -22,7 +22,11 @@ export const referenceSchema = gql`
 
   type Query {
     references(withTag: TagName): [Reference!]!
-    referenceDownloadUrl(id: ID!, acceptedAgreementId: ID): String!
+    referenceDownloadUrl(
+      id: ID!
+      acceptedAgreementId: ID
+      emailAgreement: Boolean = false
```
connor-parke-icf, line 28: somewhat nervous about optional params in the gql schema. Im not sure the implications of them, since as far as i can remember we havnt used them anywhere else. Not saying it incorrect or wrong, but i _feel_ that we should be explicit in requiring the parameter to send an email or not.

server/src/constants.ts
```diff
@@ -143,6 +143,43 @@ export const PERSON_TYPES = [
...
+  "Deliverable Received and Filed",
+  "Extension Requested",
+  "Extension Decision Made",
+  "Resubmission Requested",
+  "Public Comment Added",
+  "Terms And Conditions Requested",
+  "Application Status Updated",
+] as const;
+
+export const EMAIL_NOTIFICATION_ENTITY_TYPES = ["deliverable", "application", "reference"] as const;
```
connor-parke-icf, line 162: Should probably capitalize these to match the convention of our other natural keys

connor-parke-icf (changes requested):
So went through only the server area and found a few things that i think we should revise. I havnt looked at the Client changes nor deployment/lambda changes yet nor the hard specifics of the database changes

Main difficulty here is i think there are just too many things going on this in the PR to effectively review or provide meaningful commentary on patterns/strategy for. I think we should separate out the concerns of building just the email infrastructure from the usages of that infrastructure. If it is possible to trim down to just the database changes, basic test email api, and email sending lambda, leaving out the bulk of the places where we call the email sending functions, it will be a lot easier to conceptualize. 

The thing that is missing are changes to the database diagram .mmd file and image; without those its hard to reason the db structure from a high level. 

The other thing, im not entirely sure how we intend to handle the downloading of the reference materials or other s3 files via an email link; i feel like theres a larger set of considerations there we need to address holistically. 

Another thing is from a general pattern perspective is where the decision whether or not to send an email comes from. In some cases its determined by the caller of the main business logic functions, usually found in resolvers, and other cases its determined in the main functions themselves. I think we should try to keep the resolvers as thin as possible and free from much hard business logic. At the moment on main, there are definitly spots in the Server where the resolvers are bloated, but in general thats the direction im hoping to tend towards. 

Im nervous about the concept of the static reference material agreement, i feel like we shouldnt be hardcoding an s3path and data into the codebase and we can already rely on the existing reference material-agreement relationship for this

I didnt look too too deeply at the database changes yet, but in some of the prisma calls it looks like we're using an updateMany call as a workaround for a preferred update call with a unique key. Prisma can be wack, but most of the times ive run into this has shown that the schems model is missing things or doesnt quite fit the intent.

client/src/components/dialog/deliverable/AddDeliverableSlotDialog.tsx
dustbuster: I'm not saying this is correct or incorrect. But alternatives are lacking.

dustbuster:
I am breaking this into chunks. The review so far has given me a cleaner code path.

